PR #2, opened 2020-12-16
implement the HMC kernel

rlouf:
## Scope

Implement a HMC kernel that moves one chain for one step, with the following elements:

- [x] Velocity verlet integrator + tests
- [x] Proposal with a fixed number of steps
- [x] Gaussian euclidean metric
- [x] HMC base kernels
- [x] High-level API
- [x] Setup simple integration test

The goal is to get to a concrete implementation quickly to see the design in action.

### DIFF
--- a/blackjax/hmc.py
+++ b/blackjax/hmc.py
@@ -10,7 +10,7 @@ import blackjax.inference.integrators as integrators
 import blackjax.inference.metrics as metrics
 import blackjax.inference.proposals as proposals
 
-Array = Union[np.array, jnp.DeviceArray]
+Array = Union[np.ndarray, jnp.DeviceArray]
 PyTree = Union[Dict, List, Tuple]
 
 
--- a/blackjax/hmc.py
+++ b/blackjax/hmc.py
@@ -1,0 +1,101 @@
+"""Public API for the HMC Kernel"""
+from typing import Callable, Dict, List, NamedTuple, Tuple, Union
+
+import jax
+import jax.numpy as jnp
+import numpy as np
+
+import blackjax.inference.base as base
+import blackjax.inference.integrators as integrators
+import blackjax.inference.metrics as metrics
+import blackjax.inference.proposals as proposals
+
+Array = Union[np.array, jnp.DeviceArray]
+PyTree = Union[Dict, List, Tuple]
+
+
+class HMCParameters(NamedTuple):
+    step_size: float = 1e-3
+    num_integration_steps: int = 30
+    inv_mass_matrix: Array = None
+    divergence_threshold: int = 1000
+
+
+def new_states(position: PyTree, potential_fn: Callable) -> base.HMCState:
+    """Create chain states from positions.
+
+    The HMC kernel works with states that contain the current chain position
+    and its associated potential energy and derivative of the potential energy.
+    This function computes initial states from an initial position.  While this
+    function is intended to work for one chain, it is possible to use
+    `jax.vmap` to compute the initial state of several chains.
+
+    Example
+    -------
+    Let us assume a model with two random variables. We wish to sample from
+    4 chains with the following initial positions:
+
+        >>> import numpy as np
+        >>> init_positions = (np.random.rand(4, 1000), np.random.rand(4, 300))
+
+    We have a `logpdf` function that returns the log-probability associated with
+    the chain at a given position:
+
+        >>> potential_fn((np.random.rand(1000), np.random.rand(3000)))
+        -3.4
+
+    We can compute the initial state for each of the 4 chain as follows:
+
+        >>> import jax
+        >>> jax.vmap(new_states, in_axes=(0, None))(init_positions, potential_fn)
+
+    Parameters
+    ----------
+    position
+        The current values of the random variables whose posterior we want to
+        sample from. Can be anything from a list, a (named) tuple or a dict of
+        arrays. The arrays can either be Numpy arrays or JAX DeviceArrays.
+    potential_fn
+        A function that returns the value of the potential energy when called
+        with a position.
+
+    Returns
+    -------
+    A HMC state that contains the position, the associated potential energy and gradient of the
+    potential energy.
+    """
+    potential_energy, potential_energy_grad = jax.value_and_grad(potential_fn)(position)
+    return base.HMCState(position, potential_energy, potential_energy_grad)
+
+
+def kernel(potential_fn: Callable, parameters: HMCParameters) -> Callable:
+    """Build a HMC kernel.
+
+    Parameters
+    ----------
+    potential_fn
+        A function that returns the potential energy of a chain at a given position.
+    parameters
+        A NamedTuple that contains the parameters of the kernel to be built.
+    """
+    step_size, num_integration_steps, inv_mass_matrix, divergence_threshold = parameters
+
+    if not inv_mass_matrix:
+        raise ValueError(
+            "Expected a value for `inv_mass_matrix`,"
+            " got None. Please specify a value when initializing"
+            " the parameters or run the window adaptation."
+        )
+
+    momentum_generator, kinetic_energy_fn = metrics.gaussian_euclidean(inv_mass_matrix)
+    integrator = integrators.velocity_verlet(potential_fn, kinetic_energy_fn)
+    proposal = proposals.hmc(integrator, step_size, num_integration_steps)
+    kernel = base.hmc(
+        potential_fn,
+        proposal,
+        momentum_generator,
+        kinetic_energy_fn,
+        divergence_threshold,
+    )
+
+    return kernel
--- a/blackjax/hmc.py
+++ b/blackjax/hmc.py
@@ -80,7 +80,7 @@ def kernel(potential_fn: Callable, parameters: HMCParameters) -> Callable:
     """
     step_size, num_integration_steps, inv_mass_matrix, divergence_threshold = parameters
 
-    if not inv_mass_matrix:
+    if inv_mass_matrix is None:
         raise ValueError(
             "Expected a value for `inv_mass_matrix`,"
             " got None. Please specify a value when initializing"
--- a/blackjax/hmc.py
+++ b/blackjax/hmc.py
@@ -91,7 +91,6 @@ def kernel(potential_fn: Callable, parameters: HMCParameters) -> Callable:
     integrator = integrators.velocity_verlet(potential_fn, kinetic_energy_fn)
     proposal = proposals.hmc(integrator, step_size, num_integration_steps)
     kernel = base.hmc(
-        potential_fn,
         proposal,
         momentum_generator,
         kinetic_energy_fn,
--- a/blackjax/inference/base.py
+++ b/blackjax/inference/base.py
@@ -59,7 +59,6 @@ class HMCInfo(NamedTuple):
 
 
 def hmc(
-    potential_fn: Callable,
     proposal_generator: Callable,
     momentum_generator: Callable,
     kinetic_energy: Callable,
@@ -148,17 +147,20 @@ def hmc(
         energy = potential_energy + kinetic_energy(momentum, position)
 
         proposal, proposal_info = proposal_generator(
-            key_integrator, HMCProposalState(position, momentum, potential_energy_grad)
+            key_integrator,
+            HMCProposalState(
+                position, momentum, potential_energy, potential_energy_grad
+            ),
         )
-        new_position, new_momentum, new_potential_energy_grad = proposal
 
-        flipped_momentum = jax.tree_util.tree_multimap(lambda m: -1.0 * m, new_momentum)
-        new_potential_energy = potential_fn(new_position)
-        new_energy = new_potential_energy + kinetic_energy(
-            flipped_momentum, new_position
+        flipped_momentum = jax.tree_util.tree_multimap(
+            lambda m: -1.0 * m, proposal.momentum
+        )
+        new_energy = proposal.potential_energy + kinetic_energy(
+            flipped_momentum, proposal.position
         )
         new_state = HMCState(
-            new_position, new_potential_energy, new_potential_energy_grad
+            proposal.position, proposal.potential_energy, proposal.potential_energy_grad
         )
 
         delta_energy = energy - new_energy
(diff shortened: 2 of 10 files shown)


blackjax/hmc.py
```diff
@@ -0,0 +1,101 @@
...
+import jax
+import jax.numpy as jnp
+import numpy as np
+
+import blackjax.inference.base as base
+import blackjax.inference.integrators as integrators
+import blackjax.inference.metrics as metrics
+import blackjax.inference.proposals as proposals
+
+Array = Union[np.array, jnp.DeviceArray]
```
twolodzko, line 13: Technically `np.ndarray` is the type.

blackjax/hmc.py
```diff
@@ -0,0 +1,101 @@
...
+PyTree = Union[Dict, List, Tuple]
```
twolodzko, line 14: Those types are repeated, probably nicer to have them in single place for consistency.

blackjax/hmc.py
rlouf: Technically you're right 🙂 Thanks!

blackjax/hmc.py
rlouf: I agree. I usually duplicate things until where that single place should be becomes obvious.

rlouf:
The HMC algorithm with diagonal mass matrix works. I still need to fix the dense matrix case.

rlouf (commented):
I fixed the dense mass matrix case, and replaced `tree_flatten` with `ravel_pytree` (which also flattens arrays at the leaves). Ready for review.

blackjax/inference/base.py
```diff
@@ -0,0 +1,200 @@
...
+    is_accepted: bool
+    is_divergent: bool
+    energy: float
+    proposal: HMCProposalState
+    proposal_info: HMCProposalInfo
+
+
+def hmc(
+    potential_fn: Callable,
+    proposal_generator: Callable,
```
MarcoGorelli, line 63: I don't know how far you want to go with type hints, but perhaps they could be added here as well? e.g.
```
    proposal_generator: Callable[[Array, HMCProposalState], Tuple[HMCProposalState, HMCProposalInfo]],
```
(if I've understood correctly)

(I can work on adding these if you're open to them)

rlouf: I think it could eventually be useful to users to have their IDE raise a warning if the function does not have the correct signature. Let's wait  until NUTS to see if the signature stays the same :)

MarcoGorelli: OK, sure!

rlouf:
_Note for future self_

Implementing NUTS I realize there is a much better way to cut HMC. Indeed, HMC first generates a trajectory then draws either the last or first state at random. This can be generalized, as we may want to draw an intermediate state of the trajectory instead (see Conceptual Introduction to HMC). Or draw several states from this trajectory (see Recycled Hamiltonian Monte Carlo). The proposal is thus a combination of constructing the trajectory and sampling states from that trajectory. We should thus remove the metropolis acceptance step from `base_hmc` and have `hmc_proposal` be a combination of `forward_trajectory` and `choosing_start_or_end` sampling scheme. NUTS fits more easily in this formulation, and it allows to build all sorts of variants.

twiecki:
Is this ready to merge?

rlouf:
Ready to merge on my end, I was just waiting for potential reviews.

blackjax/inference/proposals.py
```diff
@@ -0,0 +1,69 @@
...
+        """Integrate the trajectory  `num_integration_steps` times starting from `initial_state`."""
+
+        def one_step(state, _):
+            state = integrator(state, step_size)
+            return state, state
+
+        new_state, trajectory = jax.lax.scan(
+            one_step, initial_state, jnp.arange(num_integration_steps)
+        )
+        info = HMCProposalInfo(step_size, num_integration_steps, trajectory)
```
fehiepsi, line 65: I think this is useful but memory cost will be O(num_steps), while in NUTS, it is O(log(num_steps)) and in usual HMC, it is O(1).

fehiepsi (commented):
Thanks @rlouf! Looks great overall. I just have some comments below.

blackjax/inference/base.py
```diff
@@ -0,0 +1,200 @@
...
+        momentum = momentum_generator(key_momentum, position)
+        energy = potential_energy + kinetic_energy(momentum, position)
+
+        proposal, proposal_info = proposal_generator(
+            key_integrator, HMCProposalState(position, momentum, potential_energy_grad)
+        )
+        new_position, new_momentum, new_potential_energy_grad = proposal
+
+        flipped_momentum = jax.tree_util.tree_multimap(lambda m: -1.0 * m, new_momentum)
+        new_potential_energy = potential_fn(new_position)
```
neerajprad, line 156: We can avoid this PE evaluation by caching `potential_energy` or `log_prob` in `HMCProposalState`.

blackjax/inference/base.py
```diff
@@ -0,0 +1,200 @@
...
+        p_accept = jnp.clip(jnp.exp(delta_energy), a_max=1)
+        do_accept = jax.random.bernoulli(key_accept, p_accept)
+        accept_state = (
+            new_state,
+            HMCInfo(
+                p_accept,
+                True,
+                is_divergent,
+                new_energy,
+                proposal,
```
neerajprad, line 177: nit: If we store `potential_energy` in `HMCProposalState`, we can probably have a single `HMCState` for both representing `(q, p, potential_energy, potential_energy_grad)` and not have to store `proposal` separately in `HMCInfo`. Please disregard my comment if there are other reasons for this separation.

neerajprad:
The interface looks nice to me! Feel free to disregard/stash some of my comments as future work, and merge this if needed.

blackjax/inference/integrators.py
```diff
@@ -0,0 +1,85 @@
...
+            (2018): 113-206.
+    .. [2]: McLachlan, Robert I. "On the numerical integration of ordinary
+            differential equations by symmetric composition methods." SIAM Journal on
+            Scientific Computing 16.1 (1995): 151-168.
+    .. [3]: Schlick, Tamar. Molecular modeling and simulation: an
+            interdisciplinary guide: Vol. 21. Springer
+            Science & Business Media, 2010.
+
+    """
+    a1 = 0
```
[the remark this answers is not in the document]
rlouf: No it's not. It's a remnant of MCX where several palindromic integrators (cf https://arxiv.org/abs/1711.05337) are implemented, including one where `a1` is not null. I'd leave it as "documentation".

blackjax/inference/integrators.py
```diff
@@ -0,0 +1,85 @@
...
+        )
+
+        kinetic_grad = kinetic_energy_grad_fn(momentum)
+        position = jax.tree_util.tree_multimap(
+            lambda position, kinetic_grad: position + a2 * step_size * kinetic_grad,
+            position,
+            kinetic_grad,
+        )
+
+        log_prob_grad = potential_grad_fn(position)
```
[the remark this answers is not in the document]
rlouf: You're right. In the NUTS PR I made the integrator return both the value of the potential. I'll introduce the change here too. Also `s/log_prob/potential_energy`.

blackjax/inference/integrators.py
```diff
@@ -0,0 +1,85 @@
...
+            interdisciplinary guide: Vol. 21. Springer
+            Science & Business Media, 2010.
+
+    """
+    a1 = 0
+    b1 = 0.5
+    a2 = 1 - 2 * a1
+
+    potential_grad_fn = jax.grad(potential_fn)
+    kinetic_energy_grad_fn = jax.grad(kinetic_energy_fn)
```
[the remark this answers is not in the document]
rlouf: I'd expect it to be negligible compared to the gradient of the potential, but that's a good point. Maybe we should do some perf test on a separate branch once this one is merged?

blackjax/inference/integrators.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def one_step(state: IntegratorState, step_size: float) -> IntegratorState:
+        position, momentum, log_prob_grad = state
+
+        momentum = jax.tree_util.tree_multimap(
+            lambda momentum, log_prob_grad: momentum - b1 * step_size * log_prob_grad,
```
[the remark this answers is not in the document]
rlouf: Yep there's a confusion in the names, it is the name. I will replace `log_prob` by `potential_enegy` as I said earlier.

blackjax/inference/proposals.py
rlouf: I do not have a case for this variable, apart from maybe diagnosis. With hindsight I believe we should discard all intermediate states by default; if someone ever needs the entire trajectories this can be implemented as a separate proposal.

Related to this I would like to implement a "proposal" that recycles intermediate states (https://arxiv.org/abs/1511.06925), but it can implemented so as to only keep the useful steps as well, discarding the ones that won't be in the trace.

blackjax/inference/metrics.py
```diff
@@ -0,0 +1,103 @@
...
+
+    The gaussian euclidean metric is a euclidean metric further characterized
+    by setting the conditional probability density :math:`\pi(momentum|position)`
+    to follow a standard gaussian distribution. A Newtonian hamiltonian
+    dynamics is assumed.
+
+    Arguments
+    ---------
+    inverse_mass_matrix
+        One of two-dimensional array corresponding respectively to a diagonal
```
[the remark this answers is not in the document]
rlouf: That's the constraint indeed, and I agree users should not have to read JAX's code to know that it does a `sort` on the dictionary keys when flattening :) I will add a short paragraph about that.

blackjax/hmc.py
```diff
@@ -0,0 +1,101 @@
...
+    parameters
+        A NamedTuple that contains the parameters of the kernel to be built.
+    """
+    step_size, num_integration_steps, inv_mass_matrix, divergence_threshold = parameters
+
+    if inv_mass_matrix is None:
+        raise ValueError(
+            "Expected a value for `inv_mass_matrix`,"
+            " got None. Please specify a value when initializing"
+            " the parameters or run the window adaptation."
```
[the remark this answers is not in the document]
rlouf: Yes, do you see any reason not to? The current plan is to publish a first release with the HMC and NUTS kernel and then move on to window adaptation.

blackjax/inference/base.py
rlouf: Yes, you and @fehiepsi are right, I will edit the integrator.

blackjax/inference/base.py
rlouf: The `proposal` in `HMCInfo` is the last state of the trajectory. `new_state` is what I now think is what should be called the proposal. I return the last state of the trajectory in `HMCInfo` because it can be useful to diagnose when there are divergences or the acceptance rate is very low.

blackjax/inference/integrators.py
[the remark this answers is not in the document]
rlouf: Pushed.

blackjax/inference/integrators.py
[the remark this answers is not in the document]
rlouf: Pushed.

blackjax/inference/proposals.py
rlouf: I pushed the modification, but leaving the discussion open in case you want to add something.

blackjax/inference/integrators.py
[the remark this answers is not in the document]
rlouf: Pushed.

blackjax/inference/integrators.py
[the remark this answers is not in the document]
rlouf: Pushed.

blackjax/inference/metrics.py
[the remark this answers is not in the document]
rlouf: Pushed.

blackjax/inference/proposals.py
rlouf: :+1: to keep it, :-1: to get rid of it.

fehiepsi: It depends on your usage case. I was worried that if step_size=1e-6 and trajectory length is 1 (which is popular for tricky posterior I believe) then this will hold 1e6 samples in memory.

blackjax/inference/integrators.py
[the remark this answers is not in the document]
fehiepsi: I think so too.

blackjax/inference/proposals.py
rlouf: I never thought that someone could integrate the trajectory 1e6 times in 1 HMC step! If that's common we indeed better make sure the memory doesn't explode doing so. I'll implement the change, we can always revert the commit if someone knows for sure that XLA makes some wicked optimization when the states are later discarded.

rlouf:
I finished implementing the proposed changes. @fehiepsi suggested a performance micro-optimization that we should keep in mind and experiment with later.

@twiecki @neerajprad @fehiepsi Unless you have other comments I will merge this PR tomorrow pm (Central European Time) and rebase the NUTS branch on this one.

blackjax/inference/proposals.py
fehiepsi: Haha, I believe no one wants to do so, but adaptive scheme might be not smart (if accept probs in initial steps are small, step size will be dropped pretty fast). ;)

blackjax/hmc.py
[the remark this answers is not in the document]
neerajprad: Sounds good, I was only asking since I saw window adaptation being referred here but wasn't sure about the plan. It will be good to provide some adaptation scheme for sure. If that requires any changes to the interface, we can discuss that later.

blackjax/inference/base.py
neerajprad: I see, that makes sense. If a proposal is rejected, we have both the earlier state and the proposed state to debug why it was rejected.

neerajprad (approved):

blackjax/inference/proposals.py
rlouf: True for jitting, but then we don't have to return the trajectory since we encapsulate the information in `ProposalInfo`. We will add adaptive schemes, so I think it's better to drop it for now :)

rlouf:
Thanks everyone for reviewing the code! Merging :champagne:

PR #3, opened 2020-12-31
iterative NUTS proposal

rlouf:
Like the title said. Branched from `hmc`. I am essentially re-arranging code in Numpyro to make it slightly more general to re-use existing code from HMC and so we can possibly re-use parts of it (the progressive sampling mechanisms, the tree doubling, etc) to build other HMC variants.


blackjax/inference/nuts.py
```diff
@@ -0,0 +1,115 @@
...
+    return trajectory
+
+
+def progressive_uniform_sampling(trajectory, new_state):
+    p_accept = jax.scipy.special.expit(trajectory.weight, new_state.energy)
+    return p_accept
+
+
+def progressive_biased_sampling(trajectory, new_trajectory):
+    """numpyro includes the turning and diverging in this check but I don't
```
fehiepsi, line 111: whatever where you put turning/diverging check, just making sure that you won't propose a point in a turning/diverging new_trajectory. This is necessary for detailed balance.

rlouf (commented):
In progress.

blackjax/inference/nuts.py
```diff
@@ -0,0 +1,115 @@
...
+import jax.numpy as jnp
+
+integrator = None
+max_tree_depth = 0
+rng_key = 0
+step_size = 0
+
+"""
+TODO:
+- U-turn checks;
```
rlouf, line 12: cf https://discourse.mc-stan.org/t/nuts-misses-u-turns-runs-in-circles-until-max-treedepth/9727/55

blackjax/hmc.py
```diff
@@ -0,0 +1,101 @@
...
+    step_size, num_integration_steps, inv_mass_matrix, divergence_threshold = parameters
+
+    if inv_mass_matrix is None:
+        raise ValueError(
+            "Expected a value for `inv_mass_matrix`,"
+            " got None. Please specify a value when initializing"
+            " the parameters or run the window adaptation."
+        )
+
+    momentum_generator, kinetic_energy_fn = metrics.gaussian_euclidean(inv_mass_matrix)
```
MarcoGorelli, line 90: I realise this is marked as a draft, but still, making a quick comment I noticed:
```diff
-    momentum_generator, kinetic_energy_fn = metrics.gaussian_euclidean(inv_mass_matrix)
+    momentum_generator, kinetic_energy_fn, _ = metrics.gaussian_euclidean(inv_mass_matrix)
```
or
```diff
-    momentum_generator, kinetic_energy_fn = metrics.gaussian_euclidean(inv_mass_matrix)
+    momentum_generator, kinetic_energy_fn, is_turning = metrics.gaussian_euclidean(inv_mass_matrix)
```
?

rlouf: yes, I didn't work very cleanly here :grin:

blackjax/inference/proposals.py
```diff
@@ -9,7 +9,7 @@
 step size [1]_. It is also common to draw at each step the number of
 integration steps from a distribution [1,2]_ ; empirical HMC [2]_ for instance
 learns this distribution during the adaptation phase. Other algorithms, like
-NUTS [3, 4, 5]_, determine the number of integration steps dynamically at runtime.
+NUT [3, 4, 5]_, determine the number of integration steps dynamically at runtime.
```
twiecki, line 12:
```suggestion
NUTS [3, 4, 5]_, determine the number of integration steps dynamically at runtime.
```

rlouf:
I am currently hitting a roadblock that might be on JAX's side. I have `cond` statement that looks like:

```python
updated_proposal = Proposal(
    new_proposal.state,
    jnp.logaddexp(proposal.log_weight, new_proposal.log_weight),
    new_proposal.is_diverging,
)

jax.lax.cond(
    do_accept,
    lambda _: updated_proposal,
    lamda _: proposal,
    operand=True,
)

And I get the following error:

```
TypeError: true_fun and false_fun output must have identical types, got
       Proposal(state=IntegratorState(position={'coefs': ShapedArray(float32[]), 'scale': ShapedArray(float32[])}, momentum={'coefs': ShapedArray(float32[]), 'scale': ShapedArray(float32[])}, potential_energy=ShapedArray(float32[]), potential_energy_grad={'coefs': ShapedArray(float32[]), 'scale': ShapedArray(float32[])}), log_weight=ShapedArray(float32[]), is_diverging=ShapedArray(bool[]))
       and
       Proposal(state=IntegratorState(position={'coefs': ShapedArray(float32[], weak_type=True), 'scale': ShapedArray(float32[], weak_type=True)}, momentum={'coefs': ShapedArray(float32[]), 'scale': ShapedArray(float32[])}, potential_energy=ShapedArray(float32[]), potential_energy_grad={'coefs': ShapedArray(float32[]), 'scale': ShapedArray(float32[])}), log_weight=ShapedArray(int32[], weak_type=True), is_diverging=ShapedArray(bool[], weak_type=True)).
```

As you can see, the `true_fun` and `false_fun` have the same shape and types, and only differ by their weak type. I can't find a way to update the weak_type without concretizing the ShapedArray.

@fehiepsi @neerajprad Any idea?

**Comment by fehiepsi** [comment]
@rlouf I think `log_weight=ShapedArray(int32[], weak_type=True)` is the reason. It should be float32 right?

**Comment by rlouf** [comment]
Omg I've been focusing on weak type for so long that I missed it. I need some sleep 😅

**Comment by rlouf** [comment]
@fehiepsi the tests finally pass! The structure is essentially there, so you can start reviewing the code. A few things that I still have to change:

- [ ] Plumbing to get the sampling info out
- [ ] In every `cond_fn` I added a condition to not check for U-turn when it is the first step. I will instead compute the condition in the `body_fn` and pass it around;
- [ ] There still is some confusion in the code between `Proposal` and `IntegratorState`, especially in the `mh_step.py` file. I need to fix this before we move forward.

You will see that the code is very different from Numpyro's, except the u-turn criterion (I'll add the licence at the top of the file before merging). I dropped the tree lingo to think in terms of trajectory sampling instead. Which generalizes a lot easier to other cases.

**Comment by fehiepsi** [comment]
Hi @rlouf, this is great work! The overall logic looks reasonable and clean to me. Could you let me know which part you need me to pay attention to? I think you might want to add more tests for this because one small change might heavily affect the result (especially for a complicated algorithm like NUTS).

> the code is very different from Numpyro's, except the u-turn criterion

Yeah, I believe this PR has a more flexible structure. I'm not sure what's the meaning of "very different" though... :) Regarding iterative nuts, you can add a reference (for curious readers) to [Appendix A section](https://arxiv.org/pdf/1912.11554.pdf) if you think it is a key point of the algorithm.

By the way, while working on this, do you have any feedback on numpyro codebase? I'm happy to incorporate your suggestions if there is not much change. ;)

**Comment by rlouf** [comment]
I was just wondering if you had general comments on the code structure, stuff you did not understand? Especially the dynamic integration / dynamic extension parts which defer from the traditional "tree building". You are right for the tests, it can indeed break quite easily!

>  I'm not sure what's the meaning of "very different" though... :) 

It's a relative. I'm closer to Betancourt's paper than the original implementation, which indeed makes it more flexible. But a NUTS algorithm is a NUTS algorithm, after all :)

I'll let you know for the feedback on Numpyro. Haven't looked at the code base in a long time...

**Comment by fehiepsi** [comment]
> general comments on the code structure, stuff you did not understand?

Probably not. I used Betancourt's paper as the main reference so I can follow the logic in your implementation. When you look back to the numpyro codebase, you will find that it is quite similar (modulo naming, refactoring,...). :) Apart from more tests to trust the implementation details, if you care about the compiling time for large models, you can mimic the test [here](https://github.com/pyro-ppl/numpyro/blob/master/test/test_compile.py#L15) to test for the number of times spent for compiling the potential function (it is not important if your main target is to have a flexible codebase). The optimal solution is to only compile 1 time while sampling. I guess your implementation does that but it would be nice to add a test to confirm and to prevent future codes break that optimization.

**Comment by rlouf** [comment]
> When you look back to the numpyro codebase, you will find that it is quite similar (modulo naming, refactoring,...). :) 

Yes I just had a look, I guess you can say they're similar, yes. I reimplemented from Betancourt's paper but I copied the code for the iterative u-turn.

> Apart from more tests to trust the implementation details, 

That needs to be done, yes. I'll have a shameless look at numpyro :)

> if you care about the compiling time for large models, you can mimic the test [here](https://github.com/pyro-ppl/numpyro/blob/master/test/test_compile.py#L15) to test for the number of times spent for compiling the potential function (it is not important if your main target is to have a flexible codebase).

Target is both :) At the very least we should know if it is a problem moving forward.

> The optimal solution is to only compile 1 time while sampling. I guess your implementation does that but it would be nice to add a test to confirm and to prevent future codes break that optimization.

Yes, the kernel is compiled with the potential so it shouldn't be a problem. It is not the case for the warmup, but after a lot of refactoring I managed to make it work quite fast.

**Comment by fehiepsi** [comment]
> the kernel is compiled with the potential so it shouldn't be a problem

Probably, but adding a test would be better; sometimes, the issue is more jax specific (e.g. weak type,...), rather than algorithm-specific. This will be important because you also target speed (compiling will be an issue when the model grows up). Just my 2 cents.

File: blackjax/inference/metrics.py [hunk 5]
```diff
@@ -101,7 +131,37 @@ def kinetic_energy(momentum: PyTree, *_) -> float:
...
+            ----------
+            momentum_left
+                Momentum of the leftmost point of the trajectory.
+            momentum_right
+                Momentum of the rightmost point of the trajectory.
+            momentum_sum
+                Sum of the momenta along the trajectory.
+
+            .. [1]: Betancourt, Michael J. "Generalizing the no-U-turn sampler to Riemannian manifolds." arXiv preprint arXiv:1304.1920 (2013).
+            .. [2]: "NUTS misses U-turn, runs in cicles until max depth", Stan Discourse Forum
```
canyon289, line 149:
```suggestion
            .. [2]: "NUTS misses U-turn, runs in circles until max depth", Stan Discourse Forum
```

blackjax/inference/mh_step.py
```diff
@@ -0,0 +1,150 @@
+"""Metropolis-Hasting Step"""
```
junpenglao, line 1: It is not really a MH step - probably more precise to call it "(Multinomial) sample from path" or "(Multinomial) sample from tree"

blackjax/inference/metrics.py
```diff
@@ -101,7 +131,37 @@ def kinetic_energy(momentum: PyTree, *_) -> float:
             velocity = jnp.matmul(inverse_mass_matrix, momentum)
             return 0.5 * jnp.dot(velocity, momentum)
 
-        return momentum_generator, kinetic_energy
+        def is_turning(
```
junpenglao, line 134: Seems we can simplify the code path a bit here to do reduce duplicate code with:
```python
    if ndim == 1:  # diagonal mass matrix
        mass_matrix_sqrt = jnp.sqrt(jnp.reciprocal(inverse_mass_matrix))
        multiply_op = jnp.multiply
    elif ndim == 2:
        tril_inv = jscipy.linalg.cholesky(inverse_mass_matrix)
        identity = jnp.identity(shape[0])
        mass_matrix_sqrt = jscipy.linalg.solve_triangular(
            tril_inv, identity, lower=True
        )
        multiply_op = jnp.matmul
    def momentum_generator(rng_key: jax.random.PRNGKey, position: PyTree) -> PyTree:
        _, unravel_fn = ravel_pytree(position)
        std = jax.random.normal(rng_key, shape)
        momentum = multiply_op(std, mass_matrix_sqrt)
        return unravel_fn(momentum)
    def kinetic_energy(momentum: PyTree, *_) -> float:
        momentum, _ = ravel_pytree(momentum)
        momentum = jnp.array(momentum)
        velocity = multiply_op(inverse_mass_matrix, momentum)
        return 0.5 * jnp.dot(velocity, momentum)
```

rlouf: Yes, @canyon289 implemented a similar simplification in another PR. I'll rebase before merging.

blackjax/inference/mh_step.py
rlouf: I'll do a pass on naming/docstring once it's well tested and the API is fixed to make it as pedagogical as possible.

rlouf:
I just added the test suggested by @fehiepsi. We still need to test the proposal sampling functions, test the trajectory sampling by making sure those return early when expected, and propagate the useful info so they can be returned by the kernel.

rlouf:
The code coverage not being at 100% is normal, there's a portion of code that isn't being used yet but will be in the next iterations on HMC. Ready to merge on my end.

fehiepsi (approved):
The functionality and user API look clean to me! I love it. I only have a small comment on the relation of `kernel` and its `params`:
```python
     params = nuts.NUTSParameters(
         step_size=1e-3, max_tree_depth=10, inv_mass_matrix=inv_mass_matrix
     )
     kernel = nuts.kernel(potential, params)
```
to simplify those lines but I guess this is just a design choice. It would be nice to add some example code on how to define `parameters` in the docs of `kernel` (I understand that having the type `NUTSParameters` is already informative, but because this is a design choice, an example of how the API works would be helpful for users). Just curious, have you thought about chain sampling using `vmap`/`pmap`,... - will the current API simplify that process? (this can be deferred to another pr I guess)

rlouf:
Hey - I am not sure what changes you are suggesting?

> It would be nice to add some example code on how to define parameters in the docs of kernel

Yes, that's a good idea. Did you check the example notebook?

https://github.com/blackjax-devs/blackjax/blob/7f53ac645e9b71de8a9bd3ff5190f641ceb950d4/notebooks/Introduction.ipynb

> Just curious, have you thought about chain sampling using vmap/pmap,... - will the current API simplify that process? 

The following should work -- I will add an example in the notebook.

```python
@jax.jit
def update_loop(state, key):
    keys = jax.random.split(key, num_chains)
    state, _ = jax.vmap(kernel, in_axes=(0, 0))(keys, state)
    return state, state
```

Another fun thing: I thought my implementation was really slow and just realized the default `max_tree_depth` is 100 :)

rlouf:
@junpenglao @canyon289 @twiecki Unless you see something I will merge the branch sometime tomorrow and then get the repo ready for the first release.

canyon289:
Merge it, any comments I have I can make on follow up per prs or tickets or whatever

fehiepsi:
> what changes you are suggesting

I thought that we can simplify it to `kernel = nuts.kernel(potential, step_size=1e-3, ...)` but either way is good to me. The code for chain sampling looks clean to me. Thanks!

rlouf:
@fehiepsi Using the NamedTuple will make more sense when we add the warmup and have to pass the parameters around a lot.

rlouf:
Well, actually I'll follow your suggestion :)

PR #5, opened 2021-01-16
add disambiguation of what "potential energy function" is

ericmjl:
As per title.

Tagging @rlouf.

Based on Twitter conversation: https://twitter.com/ericmjl/status/1350053510876962819


rlouf:
Thanks for this @ericmjl! I'm sure that in a parallel universe where you didn't make that change dozens of people are lost because they don't know what "potential" stands for :)

ericmjl:
My pleasure, @rlouf! I was lost in that parallel universe for the longest time with PyMC3's `pm.Potential` 😛. Let's bring these lost sheep back to reality!

PR #6, opened 2021-01-23
Code coverage and a bunch of other packaging updates

canyon289:
Work in progress while I debug live for Code Coverage uploads


canyon289:
Ready for review

requirements-dev.txt
```diff
@@ -18,6 +18,8 @@ pycodestyle==2.6.0
 pyflakes==2.2.0
 pyparsing==2.4.7
 pytest==6.1.2
+pytest-cov
```
rlouf, line 21: don't you want to pin the versions for reproducibility?

setup.py
```diff
@@ -32,8 +35,6 @@ def get_version(rel_path):
     version=get_version("blackjax/__init__.py"),
     description="Flexible and fast inference in Python",
     packages=setuptools.find_packages(),
-    install_requires=[
-        "jax==0.2.7",
-        "jaxlib==0.1.57",
-    ],
+    install_requires=INSTALL_REQS
```
rlouf, line 38: What is the advantage of having the requirements outside setup.py? I'm asking because this is the first time I see this.

tests/test_hmc.py
```diff
@@ -62,4 +62,4 @@ def test_hmc(inv_mass_matrix):
     scale_samples = states.position["scale"][5000:]
 
     assert np.mean(scale_samples) == pytest.approx(1, 1e-1)
-    assert np.mean(coefs_samples) == pytest.approx(3, 1e-1)
+    assert np.mean(coefs_samples) == pytest.approx(3, 2e-1)
```
rlouf, line 65: It's weird that tests don't pass at this level of precision (on your machine?). Irrelevant for this PR, but maybe something we should address when adding more tests for the kernel.

rlouf (approved):
Looks great to me, I like seeing all this green in a PR :) I just have a couple questions; feel free to squash & merge whenever you want.

tests/test_hmc.py
canyon289: Its not on my machine, its in CI see above for the failed test

setup.py
canyon289: so you can run `pip install -r requirements.txt`, or so automated bots like the github bot can scan the requirements.txt

requirements-dev.txt
canyon289: Ah yes I can do that. Will update

PR #7, opened 2021-02-09
Refactored gaussian_euclidean and added some tests

canyon289:
Getting some mypy failures that I think are unrelated to the code change but can investigate later

### DIFF
--- a/blackjax/inference/metrics.py
+++ b/blackjax/inference/metrics.py
@@ -84,8 +84,8 @@ def gaussian_euclidean(
 
     def momentum_generator(rng_key: jax.random.PRNGKey, position: PyTree) -> PyTree:
         _, unravel_fn = ravel_pytree(position)
-        std = jax.random.normal(rng_key, shape)
-        momentum = dot(std, mass_matrix_sqrt)
+        standard_normal_sample = jax.random.normal(rng_key, shape)
+        momentum = dot(standard_normal_sample, mass_matrix_sqrt)
         momentum_unravel = unravel_fn(momentum)
         return momentum_unravel
 
--- a/blackjax/inference/metrics.py
+++ b/blackjax/inference/metrics.py
@@ -68,7 +68,6 @@ def gaussian_euclidean(
         dot, matmul = jnp.multiply, jnp.multiply
 
     elif ndim == 2:
-
         tril_inv = jscipy.linalg.cholesky(inverse_mass_matrix)
         identity = jnp.identity(shape[0])
         mass_matrix_sqrt = jscipy.linalg.solve_triangular(
--- a/blackjax/inference/metrics.py
+++ b/blackjax/inference/metrics.py
@@ -93,6 +93,7 @@ def gaussian_euclidean(
         momentum, _ = ravel_pytree(momentum)
         momentum = jnp.array(momentum)
         velocity = matmul(inverse_mass_matrix, momentum)
-        return 0.5 * jnp.dot(velocity, momentum)
+        kinetic_energy_val = 0.5 * jnp.dot(velocity, momentum)
+        return kinetic_energy_val
 
     return momentum_generator, kinetic_energy
--- a/blackjax/inference/metrics.py
+++ b/blackjax/inference/metrics.py
@@ -63,23 +63,15 @@ def gaussian_euclidean(
     ndim = jnp.ndim(inverse_mass_matrix)
     shape = jnp.shape(inverse_mass_matrix)[:1]
 
-    if ndim == 1:  # diagonal mass matrix
+    if ndim > 2:
+        raise ValueError(
+            "The mass matrix has the wrong number of dimensions:"
+            f" expected 1 or 2, got {jnp.ndim(inverse_mass_matrix)}."
+        )
 
+    elif ndim == 1:  # diagonal mass matrix
         mass_matrix_sqrt = jnp.sqrt(jnp.reciprocal(inverse_mass_matrix))
-
-        def momentum_generator(rng_key: jax.random.PRNGKey, position: PyTree) -> PyTree:
-            _, unravel_fn = ravel_pytree(position)
-            std = jax.random.normal(rng_key, shape)
-            momentum = jnp.multiply(std, mass_matrix_sqrt)
-            return unravel_fn(momentum)
-
-        def kinetic_energy(momentum: PyTree, *_) -> float:
-            momentum, _ = ravel_pytree(momentum)
-            momentum = jnp.array(momentum)
-            velocity = jnp.multiply(inverse_mass_matrix, momentum)
-            return 0.5 * jnp.dot(velocity, momentum)
-
-        return momentum_generator, kinetic_energy
+        dot, matmul = jnp.multiply, jnp.multiply
 
     elif ndim == 2:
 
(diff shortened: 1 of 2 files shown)


canyon289:
Huh interesting I get errors locally but seems like were good in CI

image

tests/test_metrics.py
```diff
@@ -0,0 +1,57 @@
...
+import pytest
+from jax import random
+
+from blackjax.inference import metrics
+
+KEY = random.PRNGKey(0)
+
+
+@pytest.fixture()
+def patch_jax(monkeypatch):
```
MarcoGorelli, line 13: this is (`monkeypatch`) used here?

canyon289: Yes its monkeypatch? I feel like I'm missing a question hee

MarcoGorelli: sorry, ignore me, you have `monkeypatch.setattr(metrics.jnp, "multiply", multiply)`, not sure what I was thinking :flushed:

canyon289: No worries! I appreciate the review :)

blackjax/inference/metrics.py
```diff
@@ -63,23 +63,15 @@ def gaussian_euclidean(
     ndim = jnp.ndim(inverse_mass_matrix)
     shape = jnp.shape(inverse_mass_matrix)[:1]
 
-    if ndim == 1:  # diagonal mass matrix
+    if ndim > 2:
```
rlouf, line 66: Nice, I like this! You should raise an error if `ndim=0` as well. The following flow would work:

```python
if ndim == 1:
    dot, matmul = jnp.multiply, jnp.multiply
elif ndim==2:
    dot, matmul = jnp.dot, jnp.matmul
else:
    raise ValueError
    
# Everything else
```

tests/test_metrics.py
```diff
@@ -0,0 +1,57 @@
...
+    momentum, velocity = metrics.gaussian_euclidean(x)
+
+    try:
+        momentum(KEY, [1])
+    except RecursionError:
+        pass
+    assert multiply.call_count == 1
+
+
+def test_gaussian_euclidean_ndim_2(patch_jax):
```
rlouf, line 36: I learned a lot reading this!

I am wondering if instead of monkey patching the methods we could implement some functional tests? It would be simpler and test the whole chain. The issue with very granular tests is that you end up fighting against them every time you make a small change that doesn't affect the rest of the library or the results.

I tend to only test interfaces; tests are here to guarantee a contract that you make with users and other developers that are interfacing with this code. This goes both ways: the output of your function should be "correct", that includes making sure at initialization that you are not going to make computations with garbage. How about making sure here that:

1. An exception is raised with the wrong number of dimensions;
2. For 1- and 2-dimensional arrays we get values for the momentum and kinetic energy that are correct;

rlouf (changes requested):
Sorry for coming back to this so late! I like that you simplified, thus improved the `metrics` code. I have some remarks about the tests; what you did with monkey patching is impressive but I think you may be missing the target.

tests/test_metrics.py
canyon289: First off happy to change it and will do so sometime soon. Anything I say after this is just for conversation.

Dont interface tests suffer from the condition youre mentioning about fighting the testing suite after a small change? If someone makes a chance to a small piece one line of a function now you have to dig through a bunch of integration tests to figure out where it broke, versus with smaller unit tests only the the piece that broke will have an issue?

This is not an argument against integration tests, theyre great for all the reasons you mention. In my experience having both has been helpful so even if two things fail, at least one of them tells me specifically what is failing that helps fix everything else.

tests/test_metrics.py
```diff
@@ -0,0 +1,72 @@
...
+def test_gaussian_euclidean_ndim_3():
+    """Test Gaussian Euclidean Function returns correct function when ndim > 1"""
+    x = jnp.ones(shape=(1, 2, 3))
+
+    with pytest.raises(ValueError) as e:
+        metrics.gaussian_euclidean(x)
+    assert "The mass matrix has the wrong number of dimensions" in str(e)
+
+
+def test_gaussian_euclidean_dim_1():
```
canyon289, line 60: @rlouf This one better? If so I'll test kinetic energy, add an ndim 2 test and make all the other changes too

blackjax/inference/metrics.py
```diff
@@ -88,23 +80,19 @@ def kinetic_energy(momentum: PyTree, *_) -> float:
...
-    else:
-        raise ValueError(
-            "The mass matrix has the wrong number of dimensions:"
-            f" expected 1 or 2, got {jnp.ndim(inverse_mass_matrix)}."
-        )
+        dot, matmul = jnp.dot, jnp.matmul
+
+    def momentum_generator(rng_key: jax.random.PRNGKey, position: PyTree) -> PyTree:
+        _, unravel_fn = ravel_pytree(position)
+        std = jax.random.normal(rng_key, shape)
```
canyon289, line 87: What does std stand for? Standard deviation?

blackjax/inference/metrics.py
canyon289: Will do

blackjax/inference/metrics.py
rlouf: Yeah. It's a terrible name when you think too much about it.

canyon289: Maybe it should be named something like random sample, or momentum_sample?

rlouf: Oh no it stands for standard normal I believe.

tests/test_metrics.py
rlouf: If your interfaces are well-designed they don't; with a mature library interfaces are less likely to change than what they enclose. I would argue you *want* the tests to break when there is a change of interface rather, but not when you add a numerical optimization somewhere for instance. Unit tests would break on the latter.

I am not against unit tests in general; I find that they are very useful when you are first building the library as it makes you confident that you are moving in the right direction. But eventually they become a burden. Plus when you make a modification later, you know exactly what broke the tests on the interface: the modification you just made.

rlouf: You want to optimize for developer speed and test suite running speed as much as you can. When  you have too many unit tests they tend to get in the way of developers and slow down the CI.

tests/test_metrics.py
rlouf: Yes! You can add a few edge cases that you think are important. For instance, when testing distributions in MCX I made sure that for a random variable with a stricly positive support `logpdf(0)` returned `-inf` as this is an important edge case.

blackjax/inference/metrics.py
canyon289: Would you be opposed if it was renamed standard_normal to make it unambiguous?

rlouf:
Explicit is always better :)

tests/test_metrics.py
canyon289: What would the edge cases here be? Negative mass matrix? Other than that the other calculations look pretty straightforward

canyon289:
Can I merge this now?

rlouf:
Will have a look this morning, I'm sorry for no having done this earlier.

blackjax/inference/metrics.py
```diff
@@ -64,22 +64,8 @@ def gaussian_euclidean(
...
-        def kinetic_energy(momentum: PyTree, *_) -> float:
-            momentum, _ = ravel_pytree(momentum)
-            momentum = jnp.array(momentum)
-            velocity = jnp.multiply(inverse_mass_matrix, momentum)
-            return 0.5 * jnp.dot(velocity, momentum)
-
-        return momentum_generator, kinetic_energy
+        dot, matmul = jnp.multiply, jnp.multiply
 
     elif ndim == 2:
```
rlouf, line 71: remove empty line?

rlouf:
It looks great! Would you mind removing the empty line after `if ndim==2`? You can merge once that's done.

blackjax/inference/metrics.py
canyon289: Done, will merge after test pass

PR #16, opened 2021-04-08
Stan warmup

rlouf:
In this Pull Request we implement the [Stan warmup](https://mc-stan.org/docs/2_26/reference-manual/hmc-algorithm-parameters.html) for a single chain. We also change the signature of the HMC and NUTS kernel factories. Example:

```python
kernel_factory = lambda step_size, inverse_mass_matrix: nuts.kernel(potential_fn, step_size, inverse_mass_matrix)

# `info` is a tuple that contains the chain, all warmup states and the chain info
state, (step_size, inverse_mass_matrix), info = stan_warmup(
    rng_key,
    kernel_factory,
    initial_state,
    num_steps=1000
)

# We can use different kernels in the HMC family for warmup and sampling
kernel = hmc.kernel(potential_fn, steps_size, inverse_mass_matrix)
```


rlouf:
I need to separate "run" from the function that updates the stan warmup, in case we want to share information across chains.

rlouf:
This is ready to review. We will just have to wait until #21 is merged to be able to test the Stan warmup with the NUTS kernel.

PR #18, opened 2021-05-21
Bug fix in NUTS sampler

junpenglao:
- change deprecated index_update ops (https://jax.readthedocs.io/en/latest/jax.ops.html#indexed-update-functions-deprecated)
- bug fix of u turn check (it was not checking on the resulting merged trajectory)
- refactored the merging of 2 trajectory so it is easier to add robust u-turn check

### DIFF
--- a/blackjax/inference/metrics.py
+++ b/blackjax/inference/metrics.py
@@ -133,6 +133,7 @@ def gaussian_euclidean(
         velocity_left = matmul(inverse_mass_matrix, m_left)
         velocity_right = matmul(inverse_mass_matrix, m_right)
 
+        # rho = m_sum
         rho = m_sum - (m_right + m_left) / 2
         turning_at_left = jnp.dot(velocity_left, rho) <= 0
         turning_at_right = jnp.dot(velocity_right, rho) <= 0
--- a/blackjax/inference/termination.py
+++ b/blackjax/inference/termination.py
@@ -50,10 +50,7 @@ def iterative_uturn_numpyro(is_turning):
         r_ckpts, r_sum_ckpts = jax.lax.cond(
             step % 2 == 0,
             (r_ckpts, r_sum_ckpts),
-            lambda x: (
-                x[0].at[ckpt_idx_max].set(r),
-                x[1].at[ckpt_idx_max].set(r_sum)
-            ),
+            lambda x: (x[0].at[ckpt_idx_max].set(r), x[1].at[ckpt_idx_max].set(r_sum)),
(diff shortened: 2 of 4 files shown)


junpenglao:
As expect the robust u-turn check is more sensitive - but I will construct some example to test this explicitly.

fehiepsi:
I guess as long as things work, it is fine. My claim there is the topic issue happens just because the formula used in the implementation is not quite correct. The fix I proposed is simple: replace
+ z_n - z_0 ~ step_size * (r_0 + ... + r_n)

by
+ z_n - z_0 ~ step_size * (r_0 / 2 + r1 + ... + r_{n-1} + r_n / 2)

But not many people agree with my argument. :D

rlouf:
I agree with your approximation but I still fail to see how it solves the original problem. While I draw some trajectories to understand, do you remember whether the performance were improved significantly enough to notice the difference?

junpenglao:
Thanks @fehiepsi 
For me I have 2 main questions around the implementation in Numpyro NUTS is that:
- is the trajectory building already doing robust(-like) u turn check
from your replies in the Stan Discourse that seems to be the case, where numpyro is computing `z_n - z_0 ~ step_size * (r_0 / 2 + r1 + ... + r_{n-1} + r_n / 2)`
Could you point me to the code path of this implementation?

- is the robust u-turn check also happens when combining the 2 trajectory?
My assessment is no, because you need to do additional 2 u-turn check for that (left most state of the **left** trajectory vs left most state of the **right** trajectory; right most state of the **left** trajectory vs right most state of the **right** trajectory), which this PR added to Blackjax

fehiepsi:
@rlouf I did some [tests](https://gist.github.com/fehiepsi/f0b1b4b2e987d8dde3993adceaa52063) back then but the code might not work anymore. You can test by changing [this term](https://github.com/blackjax-devs/blackjax/blob/master/blackjax/inference/metrics.py#L136) to `rho = m_sum`. Probably the issue will happen with another step_size...

@junpenglao We didn't implement the proposed solution by stan dev in numpyro. As above, I just replace `rho = m_sum` by `rho = m_sum - (m_right + m_left) / 2` [here](https://github.com/blackjax-devs/blackjax/blob/master/blackjax/inference/metrics.py#L136). We'll try the solution here if we get similar problems, though the side-effect "Decrease in antithetic behavior for component means in correlated models." in stan PR makes me a bit concern. Did you observe any performance improvement with this new check?

junpenglao:
Per discussion with @rlouf offline, we will hold off the robust U-turn check until further clarification.

PR #29, opened 2021-05-26
Add number of leapfrog count to NUTS diagnostic

junpenglao:


blackjax/nuts.py
```diff
@@ -53,6 +53,7 @@ class NUTSInfo(NamedTuple):
     trajectory_leftmost_state: integrators.IntegratorState
     trajectory_rightmost_state: integrators.IntegratorState
     num_trajectory_expansions: int
+    num_leapfrogs: int
```
rlouf, line 56: Would it be possible to replace mentions of `leapfrogs` by `integration_steps`, which is more general? BlackJAX ca n be used with other integrators than leapfrog...

rlouf (changes requested):
Great! To simplify the code I was envisioning this as being a property `num_states` of the trajectory which has the advantage of not having to pass extra state around. Does that make sense?

rlouf:
Just one minor comment to keep a uniform API, ready to merge when it is fixed.

junpenglao:
@rlouf I think it is fixed now - what do you think about this change?

rlouf:
Strictly speaking length would be num_states * step size. So yeah maybe `num_states` makes more sense. Sorry for being picky about the choice of words, I think it is a big part of what makes a codebase readable.

junpenglao:
That's a good point. Done.

rlouf:
Looks great now, merging!

PR #30, opened 2021-05-27
NUTS bug fix

junpenglao:
This PR fix a few major bugs in NUTS implementation
- Proposal weight need to be accumulated regardless accept or reject
- During tree doubling, the trajectory of a new tree should not include the state that initialise the new tree
- There is a bug that during trajectory building integration was starting from the proposal position instead of the previous position in the trajectory
- Energy differences should be compare with the initial state of the step, not the proposal state or the state that initialise the new tree
- Proposal should comes from a subtree (not including the state that initialise the new tree)

### DIFF
--- a/blackjax/inference/proposal.py
+++ b/blackjax/inference/proposal.py
@@ -2,6 +2,7 @@ from typing import Callable, NamedTuple, Tuple
 
 import jax
 import jax.numpy as jnp
+import numpy as np
 
 from blackjax.inference.integrators import IntegratorState
 
@@ -16,11 +17,14 @@ class Proposal(NamedTuple):
     weight:
         Weight of the proposal. It is equal to the logarithm of the sum of the canonical
         densities of each state :math:`e^{-H(z)}` along the trajectory.
+    log_p_accept:
+        Average acceptance probabilty across entire trajectory.
     """
 
     state: IntegratorState
     energy: float
     weight: float
+    log_p_accept: float
 
 
 def proposal_generator(
@@ -28,7 +32,7 @@ def proposal_generator(
 ) -> Tuple[Callable, Callable]:
     def init(state: IntegratorState) -> Proposal:
         energy = state.potential_energy + kinetic_energy(state.position, state.momentum)
-        return Proposal(state, energy, 0.0)
+        return Proposal(state, energy, 0.0, -np.inf)
 
     def update(initial_energy: float, state: IntegratorState) -> Tuple[Proposal, bool]:
         """Generate a new proposal from a trajectory state.
@@ -55,14 +59,17 @@ def proposal_generator(
         delta_energy = jnp.where(jnp.isnan(delta_energy), -jnp.inf, delta_energy)
         is_transition_divergent = jnp.abs(delta_energy) > divergence_threshold
 
-        # The weight of the new proposal is equal to H(z) - H(z_new)
+        # The weight of the new proposal is equal to H0 - H(z_new)
         weight = delta_energy
+        # Acceptance statistic min(e^{H0 - H(z_new)}, 1)
+        log_p_accept = jnp.minimum(delta_energy, 0.0)
 
         return (
             Proposal(
                 state,
                 new_energy,
                 weight,
+                log_p_accept,
             ),
             is_transition_divergent,
         )
@@ -78,12 +85,12 @@ def proposal_generator(
 def static_binomial_sampling(rng_key, proposal, new_proposal):
     """Accept or reject a proposal based on its weight.
 
-    In the static setting, the `log_weight` of the proposal will be equal to the
-    difference of energy between the beginning and the end of the trajectory. It
+    In the static setting, the `log_p_accept` of the proposal will be equal to the
+    difference of energy between the beginning and the end of the trajectory (truncated at 0.). It
     is implemented this way to keep a consistent API with progressive sampling.
 
     """
-    p_accept = jnp.clip(jnp.exp(proposal.weight), a_max=1)
+    p_accept = jnp.exp(new_proposal.log_p_accept)
     do_accept = jax.random.bernoulli(rng_key, p_accept)
 
     return jax.lax.cond(
@@ -108,6 +115,7 @@ def progressive_uniform_sampling(rng_key, proposal, new_proposal):
     p_accept = jax.scipy.special.expit(new_proposal.weight - proposal.weight)
     do_accept = jax.random.bernoulli(rng_key, p_accept)
     new_weight = jnp.logaddexp(proposal.weight, new_proposal.weight)
+    new_log_p_accept = jnp.logaddexp(proposal.log_p_accept, new_proposal.log_p_accept)
 
     return jax.lax.cond(
         do_accept,
@@ -115,11 +123,13 @@ def progressive_uniform_sampling(rng_key, proposal, new_proposal):
             new_proposal.state,
             new_proposal.energy,
             new_weight,
+            new_log_p_accept,
         ),
         lambda _: Proposal(
             proposal.state,
             proposal.energy,
             new_weight,
+            new_log_p_accept,
         ),
         operand=None,
     )
@@ -136,6 +146,7 @@ def progressive_biased_sampling(rng_key, proposal, new_proposal):
     p_accept = jnp.clip(p_accept, a_max=1.0)
     do_accept = jax.random.bernoulli(rng_key, p_accept)
     new_weight = jnp.logaddexp(proposal.weight, new_proposal.weight)
+    new_log_p_accept = jnp.logaddexp(proposal.log_p_accept, new_proposal.log_p_accept)
 
     return jax.lax.cond(
         do_accept,
@@ -143,11 +154,13 @@ def progressive_biased_sampling(rng_key, proposal, new_proposal):
             new_proposal.state,
             new_proposal.energy,
             new_weight,
+            new_log_p_accept,
         ),
         lambda _: Proposal(
             proposal.state,
             proposal.energy,
             new_weight,
+            new_log_p_accept,
         ),
         operand=None,
     )
--- a/blackjax/inference/proposal.py
+++ b/blackjax/inference/proposal.py
@@ -17,14 +17,14 @@ class Proposal(NamedTuple):
     weight:
         Weight of the proposal. It is equal to the logarithm of the sum of the canonical
         densities of each state :math:`e^{-H(z)}` along the trajectory.
-    cumsum_log_mh_accpet:
+    sum_log_p_accept:
         cumulated Metropolis-Hastings acceptance probabilty across entire trajectory.
     """
 
     state: IntegratorState
     energy: float
     weight: float
-    cumsum_log_mh_accpet: float
+    sum_log_p_accept: float
 
 
 def proposal_generator(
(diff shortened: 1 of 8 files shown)


junpenglao:
I will add some test shortly.

rlouf:
Converted to draft PR to make it easier to follow. You can convert to normal PR when ready for review!

sethaxen:
> Proposal weight need to be accumulated regardless accept or reject

Can you clarify why this is needed?

junpenglao:
> > Proposal weight need to be accumulated regardless accept or reject
> 
> Can you clarify why this is needed?

The idea here is that since all states from the trajectory (even the sub trajectory with U turn) could have been a Metropolis proposal, we just average over the corresponding Metropolis acceptance probabilities as the expectation.

I have a longer doc written a while ago on this - i will clean it up and share here.

sethaxen:
> > > Proposal weight need to be accumulated regardless accept or reject
> > 
> > 
> > Can you clarify why this is needed?
> 
> The idea here is that since all states from the trajectory (even the sub trajectory with U turn) could have been a Metropolis proposal, we just average over the corresponding Metropolis acceptance probabilities as the expectation.
> 
> I have a longer doc written a while ago on this - i will clean it up and share here.

This would seem to violate the reversibility requirement described in section A.4.1 of [Betancourt's paper](https://arxiv.org/pdf/1701.02434v2.pdf), namely, that reversibility is violated when the proposed trajectory could not have been generated by any one of the points contained therein; therefore, all trajectory expansions containing any subtrees that would have triggered a termination must be rejected, and the points contained in such expansions cannot be considered as valid proposals.

Anyways, this all only seems to make sense to me when I am in the throes of implementing adaptive HMC, so I'd be interested in seeing the write-up to hone my own understanding.

junpenglao:
I agree with you and this is not intuitive to me as well, but this seems to be the most common current implementation (for example in Stan: https://github.com/stan-dev/stan/blob/07256d04e6fbeada660c63b0958b0bd877e536d9/src/stan/mcmc/hmc/nuts/base_nuts.hpp#L197-L199)

rlouf:
I think this refers to sampling states from the trajectory as per A.3.1. and A.3.2.

The weight is the sum of all weights along the current trajectory. Even if we reject a proposal we are still growing the trajectory so we need to update the sum.

rlouf (changes requested):
I've only had the time to review `proposals.py` so far

tests/test_hmc.py
```diff
@@ -56,7 +56,7 @@ def test_hmc(inv_mass_matrix):
     )
     kernel = hmc.kernel(potential, params)
 
-    rng_key = jax.random.PRNGKey(19)
+    rng_key = jax.random.PRNGKey(23)
```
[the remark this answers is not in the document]
junpenglao: i know LOL.
But for some reason HMC is failing even though I did not touch it - we should make the test more robust.

blackjax/inference/proposal.py
```diff
@@ -110,15 +114,24 @@ def static_binomial_sampling(rng_key, proposal, new_proposal):
 def progressive_uniform_sampling(rng_key, proposal, new_proposal):
     p_accept = jax.scipy.special.expit(new_proposal.weight - proposal.weight)
```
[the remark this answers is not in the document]
junpenglao: You are right -  i got confused myself here as well as I should be using weight and convert it into p_accept
How about naming `log_p_accept` as `cumsum_log_mh_accpet`?

blackjax/inference/proposal.py
```diff
@@ -110,15 +114,24 @@ def static_binomial_sampling(rng_key, proposal, new_proposal):
...
     do_accept = jax.random.bernoulli(rng_key, p_accept)
-
-    updated_proposal = Proposal(
-        new_proposal.state,
-        new_proposal.energy,
-        jnp.logaddexp(proposal.weight, new_proposal.weight),
-    )
+    new_weight = jnp.logaddexp(proposal.weight, new_proposal.weight)
```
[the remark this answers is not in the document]
junpenglao: We should accumulate the weight because every time you do a comparison it is winner takes all

blackjax/inference/proposal.py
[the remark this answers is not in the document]
junpenglao: Actually, we should do `jnp.clip(jnp.exp(new_proposal.weight - proposal.weight), a_max=1)` as well.

tests/test_hmc.py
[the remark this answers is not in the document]
junpenglao: Oh I did touch HMC in proposal - let me change it back.

blackjax/inference/proposal.py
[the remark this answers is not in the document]
junpenglao: Also, we dont need to clip here because expit is already result in [0, 1]

blackjax/inference/proposal.py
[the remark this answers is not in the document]
rlouf: It's not the impression I had reading Bettencourt's paper but I'll check again.

blackjax/inference/proposal.py
```diff
@@ -17,14 +17,14 @@ class Proposal(NamedTuple):
     weight:
         Weight of the proposal. It is equal to the logarithm of the sum of the canonical
         densities of each state :math:`e^{-H(z)}` along the trajectory.
-    log_p_accept:
-        Average acceptance probabilty across entire trajectory.
+    cumsum_log_mh_accpet:
```
rlouf, line 20: `sum_log_p_accept` simply ?

junpenglao: sure thing.

blackjax/inference/trajectory.py
```diff
@@ -199,30 +217,47 @@ def integrate(
...
                 _,
                 _,
             ) = expansion_state
-            _, rng_key = jax.random.split(rng_key)
+            rng_key, proposal_key = jax.random.split(rng_key)
+
+            new_state = integrator(previous_state, direction * step_size)
+            new_proposal, is_diverging = generate_proposal(initial_energy, new_state)
+
+            new_trajectory, sampled_proposal = jax.lax.cond(
```
rlouf, line 246: Could we treat the first step separately to avoid going through that `lax.cond` at every step? The code would be cleaner and I assume this is inefficient in JAX.

blackjax/inference/trajectory.py
```diff
@@ -199,30 +217,47 @@ def integrate(
...
                 trajectory,
                 termination_state,
                 _,
                 _,
             ) = expansion_state
-            _, rng_key = jax.random.split(rng_key)
+            rng_key, proposal_key = jax.random.split(rng_key)
+
+            new_state = integrator(previous_state, direction * step_size)
+            new_proposal, is_diverging = generate_proposal(initial_energy, new_state)
```
rlouf, line 244: Why is that relative to the first energy in the trajectory and not relative to the previous state?

blackjax/inference/trajectory.py
```diff
@@ -275,6 +313,162 @@ def add_one_state(expansion_state):
     return integrate
 
 
+def dynamic_recursive_integration(
+    integrator: Callable,
+    kinetic_energy: Callable,
+    uturn_check_fn: Callable,
+    divergence_threshold: float,
+    use_robust_uturn_check: bool = False,
```
rlouf, line 321: Do you see any scenario where we might want to turn this off?

blackjax/inference/trajectory.py
```diff
@@ -275,6 +313,162 @@ def add_one_state(expansion_state):
...
+            The direction in which to expand the trajectory.
+        tree_depth
+            The depth of the binary tree doubling.
+        step_size
+            The step size of the symplectic integrator.
+        initial_energy
+            Initial energy H0 of the HMC step (not to confused with the initial energy of the subtree)
+
+        """
+        if tree_depth == 0:
```
rlouf, line 376: Does that play well with `jit` ?

blackjax/inference/trajectory.py
```diff
@@ -275,6 +313,162 @@ def add_one_state(expansion_state):
...
+    sample_proposal = progressive_uniform_sampling
+
+    def buildtree_integrate(
+        rng_key: jax.numpy.DeviceArray,
+        initial_state: IntegratorState,
+        direction: int,
+        tree_depth: int,
+        step_size,
+        initial_energy: float,
+    ):
```
rlouf, line 356: A plain english description of the algorithm, like for `dynamic_multiplicative_expansion` would help lot understanding what's happening below :)

blackjax/inference/trajectory.py
junpenglao: No as it is a recursive

blackjax/inference/trajectory.py
junpenglao: Setting it False here to make sure the progressive implementation is equivalent to the recursive. Just for clarity the recursive implementation should not be used for serious work as it cannot be jitted and thus likely slow.

blackjax/inference/trajectory.py
junpenglao: Yeah that's a good idea.

blackjax/inference/trajectory.py
junpenglao: It's the same idea as HMC that you compare energy differences with the initial state.

blackjax/inference/trajectory.py
```diff
@@ -370,7 +564,7 @@ def expand_once(expansion_state):
                 new_trajectory,
                 termination_state,
                 is_diverging,
-                has_terminated,
+                is_turning_subtree,
                 num_steps,
```
rlouf, line 568: `num_steps` is no longer necessary. I am trying to find a solution for all this state passing that is becoming quite ugly.

blackjax/inference/proposal.py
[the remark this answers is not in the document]
rlouf: Actually it is more `log_sum_p_accept`, no ?

blackjax/inference/trajectory.py
junpenglao: Yeah it is in fact redundant with the `num_states` in `Trajectory`

blackjax/inference/trajectory.py
junpenglao: added - wdyt?

blackjax/inference/trajectory.py
junpenglao: With that change now GLOBAL["count"] is 2 instead of 1 in test_compilation.py for NUTS - is this a big issue?

blackjax/inference/trajectory.py
```diff
@@ -379,55 +573,34 @@ def expand_once(expansion_state):
...
+                        proposal.cumsum_log_mh_accpet, new_proposal.cumsum_log_mh_accpet
+                    ),
+                )
+
             sampled_proposal = jax.lax.cond(
-                is_diverging | has_terminated,
-                proposal,
-                lambda x: x,
-                (rng_key, proposal, new_proposal),
+                is_diverging | is_turning_subtree,
```
rlouf, line 603: strictly speaking `is_tuning_subtree` -> `has_terminated` to stay consistent with `termination_criterion`.

blackjax/inference/trajectory.py
rlouf: That's enough for someone reading the code. Someone will do another pass one day™ when documenting the module.

blackjax/nuts.py
```diff
@@ -54,6 +59,7 @@ class NUTSInfo(NamedTuple):
     trajectory_rightmost_state: integrators.IntegratorState
     num_trajectory_expansions: int
     integration_steps: int
+    avg_accept_prob: float
```
rlouf, line 62: Would you mind calling this `acceptance_probability` for now (and leave a comment next to it)? I need this parameter in `Info` for step size adaptation. This is not perfect but it will do for now.

junpenglao: Sure thing.

blackjax/inference/trajectory.py
rlouf: As long as it's not a multiple of `num_samples` we're good.

rlouf:
Awesome! Leaving minor refactoring for a future PR before 0.2.

PR #40, opened 2021-06-02
Add tempered SMC

AdrienCorenflos:
This is not meant to work yet, but should have most of the features so that we can discuss.

Inspired by https://github.com/AdrienCorenflos/tempered-smc-optimization


AdrienCorenflos:
Related issue: https://github.com/blackjax-devs/blackjax/issues/41

blackjax/tempered_smc.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def _normalize(log_weights):
+    """ Normalize log-weights into weights
+    """
+    w = jnp.exp(log_weights - jnp.max(log_weights))
+    return w / w.sum()
+
+
+def kernel(potential_fn: Callable, mcmc_kernel: Callable, new_mcmc_state: Callable, resampling_method: Callable,
```
rlouf, line 44: Here I am torn between your solution of passing `mcmc_parameters` as an argument to this function, or the following:

```python
mcmc_kernel_factory = lambda potential: mcmc.new_kernel(potential_fn, parameters)

def kernel(potential_fn, mcmc_kernel_factory, new_mcmc_state, resampling_method, mcmc_step):
    raise NotImplementedError
```

blackjax/tempered_smc.py
```diff
@@ -0,0 +1,116 @@
...
+            The new state of the tempered SMC algorithm
+         info: SMCInfo,
+            Additional information on the SMC step
+        """
+        n_particles = jax.tree_flatten(state.particles)[0].shape[0]
+
+        def tempered_potential_fn(pytree):
+            return -state.lamda * potential_fn(pytree)
+
+        step_mcmc_kernel = jax.vmap(mcmc_kernel(tempered_potential_fn, mcmc_parameters), in_axes=(0, 0))
```
rlouf, line 94: How about?

```python
tempered_potential_fn = lambda position: -state.lmbda * potential_fn(position)
kernel = mcmc_kernel_factory(tempered_potential_fn)
```

blackjax/tempered_smc.py
```diff
@@ -0,0 +1,116 @@
...
+        keys = jax.random.split(rng_key, mcmc_steps * n_particles)
+        keys = keys.reshape((mcmc_steps, n_particles))
+
+        def mcmc_loop_body(mcmc_state, body_keys):
+            mcmc_state, _ = step_mcmc_kernel(body_keys, mcmc_state)
+            return mcmc_state, None
+
+        initial_mcmc_states = jax.vmap(new_mcmc_state, in_axes=(0, None))(state.particles, tempered_potential_fn)
+        last_mcmc_state, _ = jax.lax.scan(mcmc_loop_body, initial_mcmc_states, keys)
+        particles = last_mcmc_state.position
```
rlouf, line 105: Let me try to simplify this a little bit:

```python
mcmc_key, resampling_key = jax.random.split(rng_key, 2)

def mcmc_step(states, rng_key):
    keys = jax.random.split(rng_key, num_particules)
    states, info = jax.vmap(kernel)(states)
    return states, (states, info)

keys = jax.random.split(mcmc_key, num_mcmc_steps)
initial_states = jax.vmap(new_mcmc_state, in_axes=(0, None))(state.particles, tempered_potential_fn)
last_states, _ = jax.lax.scan(mcmc_step, initial_states, keys)
```

blackjax/tempered_smc.py
```diff
@@ -0,0 +1,116 @@
...
+
+class SMCState(NamedTuple):
+    """ Current state for the tempered SMC algorithm. This encodes the current set of sampled particles
+    as well as the corresponding tempering parameter :math:`\lambda`
+    """
+    particles: PyTree[Array]
+    lamda: float
+
+
+class SMCInfo(NamedTuple):
```
rlouf, line 25: We should propagate the info that comes from the MCMC algorithm, in a `mcmc_info` field. Can be useful for debugging.

blackjax/tempered_smc.py
```diff
@@ -0,0 +1,116 @@
...
+    """Additional information on the tempered SMC step.
+
+    weights: jnp.ndarray
+        The weights after the MCMC pass.
+    """
+    weights: jnp.ndarray
+
+
+new_state = base.new_hmc_state
```
rlouf, line 34: I don't see this used anywhere

blackjax/tempered_smc.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def _normalize(log_weights):
+    """ Normalize log-weights into weights
+    """
+    w = jnp.exp(log_weights - jnp.max(log_weights))
+    return w / w.sum()
```
rlouf, line 41: I would leave this at the end of the file

rlouf (changes requested):
You got the idea ! I made some suggestions as to how the code might be simplified.

blackjax/tempered_smc.py
```diff
@@ -0,0 +1,116 @@
...
+        Number of MCMC steps to run per SMC step
+
+    Returns
+    -------
+    A callable that takes a rng_key and a Pytree that contains the current state
+    of the chain and that returns a new state of the chain along with
+    information about the transition.
+    """
+
+    def one_step(rng_key: jnp.ndarray, state: SMCState, delta: float) -> Tuple[SMCState, SMCInfo]:
```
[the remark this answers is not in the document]
AdrienCorenflos: It's not exactly propagated is it? lambda is, but delta is used to increment lambda only.

blackjax/tempered_smc.py
AdrienCorenflos: I am not a fan of splitting keys within a loop, it is less efficient for very little added visibility. Also in the case of tempered SMC, there is no reason to store the intermediate states of the MCMC chain like you just did, (except for waste free I guess, but the implementation would be wildly different)

blackjax/tempered_smc.py
AdrienCorenflos: np

blackjax/tempered_smc.py
AdrienCorenflos: Yup, it's a copy paste residual

blackjax/tempered_smc.py
AdrienCorenflos: OK, I tend to prefer having auxiliary functions at the beginning, but your repo your rules!

blackjax/tempered_smc.py
AdrienCorenflos: I prefer your version. It makes the interface cleaner

blackjax/tempered_smc.py
AdrienCorenflos: More interestingly what do you think of the vmap here. Should it be the responsibility of the user (this way we allow for manual batching), or abstracted away in the code like I've done? It's pretty hard to catch the fact that the kernel may not be batched, so if we leave it as a user responsibility, returning readable errors saying "hey you forgot to batch" is probably gonna be impossible

blackjax/tempered_smc.py
rlouf: I don't like my novels to start with all the subplots, and I end up reading code a lot. But that's a matter of taste, I guess.

blackjax/tempered_smc.py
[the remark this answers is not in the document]
AdrienCorenflos: But I guess this decision can come at the end of the dev work as it would be a minimal change anyway, so maybe we should keep it for later when the rest of the design is clearer? Or is the reason why you want it in the state specific to the fact you want one_step to only take key + state?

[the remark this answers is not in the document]
AdrienCorenflos: the pb with having delta on the state is that it asks the user to manipulate the state in between two steps, essentially doing smthg like:
```python 
particles, weights, delta, lamda = state
state = State(particles, weights, new_delta, lamda)
```
which feels a bit wasteful for no reason

blackjax/tempered_smc.py
rlouf: - I haven't noticed much difference between splitting all at once and splitting within the loop, but I haven't tried in a while.
- We'll come to users batching themselves later I think; you wouldn't batch NUTS like you do HMC for instance, since NUTS chains can take very different number of steps at each round. For now we can batch inside as I don't know what the API will look like.

AdrienCorenflos:
Thanks for the comments overall @rlouf I'm working on some paper code today but will continue the implementation soonish

blackjax/tempered_smc.py
rlouf: And this pattern of partially applying comes back often in the library, eg for adaptation.

blackjax/tempered_smc.py
AdrienCorenflos: That depends how much of a PEP purist you want to be. Assigning lambda functions to variables is now advised against as it messes up with the stack trace at debug time.

rlouf: I'm a purist within reason. This specific check is disabled on this repo. Since the codebase is purely functional, disallowing lambda function assignment would make the code harder to read.

blackjax/tempered_smc.py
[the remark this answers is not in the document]
rlouf: Actually, could you add the instruction where you change the value of  `delta`? I need the full picture to be able to judge.

AdrienCorenflos:
@rlouf I added a function for choosing delta (non ideal, but will do). The resulting code is then something like 

```python
from ... import ...

target_ess = 0.75

smc_kernel = kernel(potential_fn, ...)

def tempering_body(carry):
    key, smc_state = carry
    key, subkey = split(key)
    delta = ess_solver(potential_fn, smc_state, target_ess)
    smc_state = smc_kernel(subkey, smc_state, next_delta)
    return key, smc_state

def tempering_stop_cond(carry):
    _, smc_state = carry
    return smc_state.lamda < 1

_, final_state = jax.lax.while_loop(tempering_stop_cond, tempering_body, ...)
```

rlouf:
This is looking very good! I'm confident we'll soon converge to something we can merge :)

blackjax/tempered_smc.py
```diff
@@ -0,0 +1,251 @@
...
+        stop_criterion = lambda carry: mcmc_stop_criterion(carry[0], carry[1])
+
+        def mcmc_loop_body(carry):
+            i, mcmc_state, key = carry
+            key, body_key = jax.random.split(key, 2)
+            op_keys = jax.random.split(body_key, n_particles)
+            mcmc_state, _ = step_mcmc_kernel(op_keys, mcmc_state)
+            return i + 1, mcmc_state, key
+
+        initial_mcmc_states = jax.vmap(new_mcmc_state, in_axes=[0, None])(state.particles, tempered_potential_fn)
```
rlouf, line 93: We would get rid of a lot of code by letting `particles` be the MCMC state. Since all MCMC states have a `position` argument, we could safely pass `position.state` to every function that only needs the position.

So this line would go away.

blackjax/tempered_smc.py
```diff
@@ -0,0 +1,251 @@
...
+        total_steps, last_mcmc_state, _ = jax.lax.while_loop(stop_criterion,
+                                                             mcmc_loop_body,
+                                                             (0, initial_mcmc_states, rng_key))
+        particles = last_mcmc_state.position
+
+        log_weights = -delta * jax.vmap(potential_fn, in_axes=[0])(particles)
```
rlouf, line 99:
```python
log_weights = -delta * jax.vmap(potential_fn, in_axes=[0])(particles.position)
```

blackjax/tempered_smc.py
```diff
@@ -0,0 +1,251 @@
...
+            key, body_key = jax.random.split(key, 2)
+            op_keys = jax.random.split(body_key, n_particles)
+            mcmc_state, _ = step_mcmc_kernel(op_keys, mcmc_state)
+            return i + 1, mcmc_state, key
+
+        initial_mcmc_states = jax.vmap(new_mcmc_state, in_axes=[0, None])(state.particles, tempered_potential_fn)
+        total_steps, last_mcmc_state, _ = jax.lax.while_loop(stop_criterion,
+                                                             mcmc_loop_body,
+                                                             (0, initial_mcmc_states, rng_key))
+        particles = last_mcmc_state.position
```
rlouf, line 97: goes away

blackjax/tempered_smc.py
```diff
@@ -0,0 +1,251 @@
...
+
+        def mcmc_loop_body(carry):
+            i, mcmc_state, key = carry
+            key, body_key = jax.random.split(key, 2)
+            op_keys = jax.random.split(body_key, n_particles)
+            mcmc_state, _ = step_mcmc_kernel(op_keys, mcmc_state)
+            return i + 1, mcmc_state, key
+
+        initial_mcmc_states = jax.vmap(new_mcmc_state, in_axes=[0, None])(state.particles, tempered_potential_fn)
+        total_steps, last_mcmc_state, _ = jax.lax.while_loop(stop_criterion,
```
rlouf, line 94:
```python
particles, _ = jax.lax.scan(...)
```

(we'll think about what we pass as info later)

blackjax/tempered_smc.py
```diff
@@ -0,0 +1,251 @@
...
+    next_lamda = smc_state.lamda + estimated_delta
+
+    delta = jax.lax.cond(next_lamda > 1,
+                         lambda _: 1 - smc_state.lamda,
+                         lambda _: estimated_delta,
+                         operand=None)
+    return delta
+
+
+def systematic(weights: jnp.ndarray, key: jnp.ndarray):
```
rlouf, line 230: Could go in `inference/smc/resampling.py` since there will likely be several resampling schemes in the future. I would move everything hmc related to `inference/hmc`.

blackjax/tempered_smc.py
```diff
@@ -0,0 +1,251 @@
...
+        Returns
+        -------
+        state: SMCState,
+            The new state of the tempered SMC algorithm
+         info: SMCInfo,
+            Additional information on the SMC step
+        """
+        n_particles = jax.tree_flatten(state.particles)[0][0].shape[0]
+
+        tempered_potential_fn = lambda pytree: -state.lamda * potential_fn(pytree)
```
rlouf, line 79: s/pytree/position?

blackjax/tempered_smc.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+def _normalize(log_weights):
+    """ Normalize log-weights into weights
+    """
+    w = jnp.exp(log_weights - jnp.max(log_weights))
+    return w / w.sum()
+
+
+def _log_ess(log_weights):
```
rlouf, line 118: Everything ess-related could go in a `inference/smc/ess.py`

tests/test_tempered_smc.py
```diff
@@ -0,0 +1,91 @@
...
+    logpdf += stats.norm.logpdf(preds, y, scale)
+    return -jnp.sum(logpdf)
+
+
+def inference_loop(rng_key, kernel, initial_state, delta_solver):
+    def cond(carry):
+        _, state, _ = carry
+        return 1 - state.lamda > 1e-6
+
+    def body(carry):
```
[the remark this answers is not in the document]
rlouf: Actually, how about computing the tempered potential in this loop as well so the SMC kernel above is very general? In which case you could move it to `inference/smc/kernel.py`.

The general SMC state would reduce to the MCMC state and here you could create a TemperedSMC state that includes the SMC state and delta.

blackjax/tempered_smc.py
```diff
@@ -0,0 +1,251 @@
...
+            Additional information on the SMC step
+        """
+        n_particles = jax.tree_flatten(state.particles)[0][0].shape[0]
+
+        tempered_potential_fn = lambda pytree: -state.lamda * potential_fn(pytree)
+
+        step_mcmc_kernel = jax.vmap(mcmc_kernel_factory(tempered_potential_fn), in_axes=(0, 0))
+        rng_key, resampling_key = jax.random.split(rng_key, 2)
+
+        stop_criterion = lambda carry: mcmc_stop_criterion(carry[0], carry[1])
```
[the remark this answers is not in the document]
AdrienCorenflos: okok

tests/test_tempered_smc.py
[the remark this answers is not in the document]
AdrienCorenflos: Yeah I was thinking about adding a level in the module hierarchy too to be able to split responsibilities.
Plan was indeed to have the delta calculation inside the tempering, but I put it outside as you wanted to see what it would be then. More than happy to move it back inside.

blackjax/tempered_smc.py
AdrienCorenflos: ok

blackjax/tempered_smc.py
AdrienCorenflos: ok

blackjax/tempered_smc.py
AdrienCorenflos: ok

blackjax/tempered_smc.py
AdrienCorenflos: agreed re the indo

blackjax/tempered_smc.py
AdrienCorenflos: ok

blackjax/tempered_smc.py
AdrienCorenflos: ok

AdrienCorenflos:
I think we are in agreement. I was a bit unhappy having everything in the same file anyway

tests/test_tempered_smc.py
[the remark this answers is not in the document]
rlouf: The code structure would look something like

```python
def tempered_smc(...):
    # all the tempering logic here
    particles, info = general_smc_kernel(...)
    return particles, info
```

AdrienCorenflos:
@rlouf  What do you think? I guess we are getting closer?

rlouf:
We're getting really close.

rlouf:
Let me know when the tests pass and I'll have a closer look. It would also be nice to showcase the algorithm in a notebook, show a situation where it really shines as opposed to HMC alone.

AdrienCorenflos:
I forgot to take the prior into consideration in the code, hence the garbage numbers (it is arbitrary in the optimisation context, which is what my code was doing originally, so I didn't think about it). Need to think about it, but probably won't happen before Monday.

AdrienCorenflos (commented):
> Let me know when the tests pass and I'll have a closer look. It would also be nice to showcase the algorithm in a notebook, show a situation where it really shines as opposed to HMC alone.

Example will probably be a good old multimodal distribution.

rlouf:
No problem, take your time.

And yes for the multimodal distribution! I'm curious to see how this performs compared to parallel tempering with HMC.

AdrienCorenflos:
Apart from cosmetics, this should do I believe!

AdrienCorenflos:
@rlouf I don't trust the tests failures not gonna lie. Imports are well sorted.

AdrienCorenflos:
> @rlouf I don't trust the tests failures not gonna lie. Imports are well sorted.

Looks like this is due to this isort thing. This doesn't feel like a standard python practice, what is it exactly/what's the purpose?

rlouf:
This looks great @AdrienCorenflos ! I'll find some time to dive in your code this week.

The purpose is normalizing the code as much as possible to make it easier to read. It's useful on codebases on which more than one person works, and pretty standard afaik.

rlouf:
PS: could you give me [push access](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to your branch? That way I can make cosmetic changes without bothering you. For everything else than cosmetic I'll leave comments and questions here.

AdrienCorenflos:
> This looks great @AdrienCorenflos ! I'll find some time to dive in your code this week.
> 
> The purpose is normalizing the code as much as possible to make it easier to read. It's useful on codebases on which more than one person works, and pretty standard afaik.

Honestly first time I see it as a test on an open source package. Doesn't seem that well supported at least by pycharm

AdrienCorenflos:
Err, all these tests pass locally...

AdrienCorenflos:
> PS: could you give me [push access](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to your branch? That way I can make cosmetic changes without bothering you. For everything else than cosmetic I'll leave comments and questions here.

You already have it
image

blackjax/inference/smc/ess.py
```diff
@@ -0,0 +1,78 @@
...
+    ess: float
+        The effective sample size
+    """
+
+    log_weights = log_weights - jnp.max(log_weights)
+    w = jnp.exp(log_weights)
+
+    if log:
+        w2: jnp.ndarray = jnp.exp(2 * log_weights)
+        return 2 * jnp.log(w.sum()) - jnp.log(w2.sum())
```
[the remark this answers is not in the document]
AdrienCorenflos: This will be stable I believe, I removed the max element from the logw already. Also can't use logaddexp here with these. maybe 2 * logsumexp(logw) - logsumexp(2 logw), but it is wasteful, and less precise, as it does the same thing as I did but doesn't take into account the fact that max(logw) simplifies.

blackjax/inference/smc/resampling.py
```diff
@@ -0,0 +1,80 @@
+""" All things resampling. """
+from functools import partial
+
+import jax
+import jax.numpy as jnp
+
+
+def _resampling_func(func, name, desc=""):
```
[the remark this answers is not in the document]
AdrienCorenflos: added residual this morning :)

blackjax/tempered_smc.py
```diff
@@ -0,0 +1,174 @@
...
+
+    n_iter: int
+        Current tempered SMC step
+    smc_state: SMCState
+        The state of the inner SMC sampler.
+    lamda: float
+        Current value of the tempering parameter.
+    """
+    n_iter: int
+    smc_state: SMCState
```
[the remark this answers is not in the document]
AdrienCorenflos: Feel free to change the cosmetics the way you want.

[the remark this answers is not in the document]
AdrienCorenflos: I kept it like this to also keep the SMC independent so as to be able to implement sampling of other forms (SSM, etc), later on.

blackjax/inference/smc/smc.py
```diff
@@ -0,0 +1,110 @@
...
+
+        step_mcmc_kernel = jax.vmap(mcmc_kernel, in_axes=(0, 0))
+        rng_key, resampling_key = jax.random.split(rng_key, 2)
+
+        def mcmc_loop_body(mcmc_state, step_key):
+            mcmc_keys = jax.random.split(step_key, n_particles)
+            mcmc_state, _ = step_mcmc_kernel(mcmc_keys, mcmc_state)
+            return mcmc_state, mcmc_state
+
+        initial_mcmc_states = jax.vmap(new_mcmc_state, in_axes=[0])(state.particles)
```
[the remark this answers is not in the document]
AdrienCorenflos: I don't see how this is complicated as it is using your high level interface directly.

rlouf:
Heu don't worry about the imports for now, I'll fix it. You need to make sure to use the `--profile=black` flag with `isort`.

AdrienCorenflos:
I've put in a somewhat illustrative example. You can make it fairer it by running parallel chains for HMC and NUTS, but I felt it was a bit unecessary.

AdrienCorenflos:
@rlouf just checking in to see when we can wrap this up and call it a success :D

rlouf:
Sorry, I've been traveling lately and been busy with other things. I will go over the PR next week so we don't lose too much momentum :)

rlouf:
I have done a little refactoring (mostly to follow more closely the existing naming conventions/code structure) at the higher level and will make another pass next week. So far so good!

rlouf:
I removed the `SMCState` NamedTuple as it only contained the particles. I changed `test_smc` and `test_tempered_smc` accordingly, but I might have done something wrong as they now all break. Here is what is left to do for me on this part:

- [ ] Fix the broken tests
- [ ] Add a `SamplerState` type and use instead of `PyTree` for particles
- [ ] Understand what's going on with the separate prior

Regarding the high level level algorithms, `fixed_schedule_tempered_smc` is not needed. Indeed, we could write the base kernel so it has the following signature:

```python
def one_step(rng_key, particles, lambda):
```

`fixed_schedule_tempered_smc` is then obtained by using `jax.lax.scan` to scan over the array of lambda values. `adaptive_tempered_smc` is then obtained by calling `one_step` with the computed value of `lambda` at every step.

AdrienCorenflos:
> I removed the `SMCState` NamedTuple as it only contained the particles. I changed `test_smc` and `test_tempered_smc` accordingly, but I might have done something wrong as they now all break. Here is what is left to do for me on this part:
> 
> * [ ]  Fix the broken tests
> * [ ]  Add a `SamplerState` type and use instead of `PyTree` for particles
> * [ ]  Understand what's going on with the separate prior
> 
> Regarding the high level level algorithms, `fixed_schedule_tempered_smc` is not needed. Indeed, we could write the base kernel so it has the following signature:
> 
> ```python
> def one_step(rng_key, particles, lambda):
> ```
> 
> `fixed_schedule_tempered_smc` is then obtained by using `jax.lax.scan` to scan over the array of lambda values. `adaptive_tempered_smc` is then obtained by calling `one_step` with the computed value of `lambda` at every step.

To explain the separate prior, it comes from the fact that you only temper the likelihood term. A good way to see that there is no way around it is essentially to ask yourself what would happen for $\lambda=0$ otherwise.

AdrienCorenflos:
Also why do you want to move the tempering schedule out of the algo? From my POV, it feels a bit like if you were asking the users to handle the symplectic integrator in the sampling loop too.

rlouf:
Thanks for fixing everything I broke ! Looks like there are still a couple errors because I forgot to correct the paths in the files that define the adaptation algorithms.

1. Yes I re-read Chopin's book; I am so used to HMC and never separating the likelihood and prior that it kind of surprised me.
2. Yes we should make sure that the notebooks can run after every change. We should open a new issue for that.
3. I don't see how providing the array before iterating _simplifies_ anything. First you have to keep track of a number of iterations, which is awkward. You also need to keep track of how many times you've applied the kernel to make sure you don't go past the size of the array of temperatures.
All you are doing is "consuming" an array of temperatures, using `scan` does exactly that and you don't have to bother with indexing.

AdrienCorenflos:
So what are you saying we do? Get the ess solver out and somehow ask the
user to configure it themselves to then pass it to the sampling loop?

AdrienCorenflos:
IMO if you do that you just make the algo unusable for people who aren't
already experts of SMC and of annealing. Or maybe I don't have enough
perspective and you have a good idea of how to make this somehow user
friendly?

rlouf:
From what I understood, finding a "good" temperature profile is the hardest thing about SMC. We can thus safely assume that people who are not familiar with annealing or SMC will use `adaptive_tempered_smc`. We thus leave this algorithm as is.

The important difference between `adaptive_tempered_smc` and the original version of `fixed_schedule_tempered_smc` is that in the latter the kernel holds a state which determines the maximum number of iterations. This is what bothers me the most, because it is neither idiomatic JAX code (using `scan` is recommended instead), nor is it justified. I think it is better to let users iterate over the array of temperatures themselves.

@junpenglao what do you think?

AdrienCorenflos:
Ok I can get behind getting rid of the fixed schedule algo. I was worried
you wanted to get rid of the adaptive one.

rlouf:
Great, I'll fix the remaining tests and formatting. Then we will wait until #66 is merged and add a new test for this algorithm before merging.

rlouf:
I did some further refactoring. In particular, we don't need to pass `new_mcmc_state` function if we work with the RWMH/HMC/etc states directly. There might be some nitpicking stuff left. Now let's wait for #66 to be merged so we can add a new test to check sampling accuracy.

rlouf:
I just rebased on the main branch. All we need now is a test like [this one](https://github.com/AdrienCorenflos/blackjax/blob/f71e8c38dcce86217144cf7c4434bdd658f0e2fe/tests/test_sampling.py#L212) to further check the validity of the samples. Then we're good to merge!

rlouf:
You can test using high dimension gaussian, like for HMC and NUTS.

AdrienCorenflos:
> You can test using high dimension gaussian, like for HMC and NUTS.

Done.

rlouf:
Great job @AdrienCorenflos! I made some minor modifications (such as types) and made sure all tests pass. Thank you!

AdrienCorenflos:
Yeeeah

PR #42, opened 2021-06-02
Small refactor so we do less control flow

junpenglao:
Always append the state to the rightmost during progressive integration and only flip the 2 end when integration is terminated.

### DIFF
--- a/blackjax/inference/trajectory.py
+++ b/blackjax/inference/trajectory.py
@@ -50,9 +50,7 @@ class Trajectory(NamedTuple):
     num_states: int
 
 
-def append_to_trajectory(
-    trajectory: Trajectory, state: IntegratorState
-) -> Trajectory:
+def append_to_trajectory(trajectory: Trajectory, state: IntegratorState) -> Trajectory:
     """Append a state to the (right of the) trajectory to form a new trajectory."""
     momentum_sum = jax.tree_util.tree_multimap(
         jnp.add, trajectory.momentum_sum, state.momentum
--- a/blackjax/inference/trajectory.py
+++ b/blackjax/inference/trajectory.py
@@ -50,10 +50,10 @@ class Trajectory(NamedTuple):
     num_states: int
 
 
-def rightmost_append_to_trajectory(
+def append_to_trajectory(
     trajectory: Trajectory, state: IntegratorState
 ) -> Trajectory:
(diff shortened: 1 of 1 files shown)


blackjax/inference/trajectory.py
```diff
@@ -289,18 +280,29 @@ def add_one_state(expansion_state):
...
-            _,
             trajectory,
             termination_state,
             is_diverging,
             has_terminated,
         ) = jax.lax.while_loop(
             do_keep_integrating, add_one_state, initial_integration_state
         )
+        # In the while_loop we always extend on the right most direction.
+        new_trajectory = jax.lax.cond(
```
junpenglao, line 291: @rlouf we can also fold this into `reorder_trajectories`, WDYT?

rlouf:
Do you see any notable perf difference?

junpenglao:
> Do you see any noticeable perf difference?

I feel the test runs faster but likely it is from your PR last round :-)

blackjax/inference/trajectory.py
```diff
@@ -50,21 +50,15 @@ class Trajectory(NamedTuple):
     num_states: int
 
 
-def append_to_trajectory(
-    direction: int, trajectory: Trajectory, state: IntegratorState
+def rightmost_append_to_trajectory(
```
rlouf, line 53: For every language I know of  `append`  = `add to the right` so we can probably name it `append_to_trajectory`.

blackjax/inference/trajectory.py
```diff
@@ -50,21 +50,15 @@ class Trajectory(NamedTuple):
...
+    trajectory: Trajectory, state: IntegratorState
 ) -> Trajectory:
-    """Append a state to the trajectory to form a new trajectory."""
```
rlouf, line 56: """Append a state to the (right of the) trajectory to form a new trajectory."""

blackjax/inference/trajectory.py
rlouf: I think that would make it uselessly complicated, we do not gain anything in terms of computation and it is a lot less explicit.

rlouf:
Looks good to me. I only have two minor comments on the naming, ready to merge when it's fixed. Then I'll refactor the state passing and we can release.

rlouf:
Great, merging.

PR #44, opened 2021-06-05
fix Stan warmup

rlouf:
While the warmup performs correctly on most examples, it fails on a simple gaussian target. The goal of this PR is to diagnose and solve the issue.


rlouf:
Closing for now

Issue #46, opened 2021-06-07
HMC fails to 'draw' from standard gaussian target


wiep:
blackjax currently fails to target the standard normal distribution with HMC. in my example, the empirical mean is close to 0  but the empirical variance is close to 0.5. similar values arise when employing nuts with and without stan warmup.

running
```py
import numpy as np
import jax
import jax.numpy as jnp
import blackjax.hmc as hmc
import matplotlib.pyplot as plt

potential = lambda x: -jax.scipy.stats.norm.logpdf(x, loc=0.0, scale=1.0).squeeze()
initial_position = np.array([1.0,])
initial_state = hmc.new_state(initial_position, potential)
initial_state

inv_mass_matrix = 0.1 * jnp.ones_like(initial_position)
num_integration_steps=100
step_size=1e-2

hmc_kernel = hmc.kernel(
    potential,
    step_size=step_size,
    inverse_mass_matrix=inv_mass_matrix,
    num_integration_steps=num_integration_steps
)

hmc_kernel = jax.jit(hmc_kernel)

def inference_loop(rng_key, kernel, initial_state, num_samples):
    def one_step(state, rng_key):
        state, _ = kernel(rng_key, state)
        return state, state

    keys = jax.random.split(rng_key, num_samples)
    _, states = jax.lax.scan(one_step, initial_state, keys)

    return states

rng_key = jax.random.PRNGKey(0)
states = inference_loop(rng_key, hmc_kernel, initial_state, 50_000)

samples = states.position.block_until_ready()
print(np.mean(samples, axis=0))
print(np.var(samples, axis=0))
plt.plot(samples)
plt.show()
```
gives
```
[-0.00325776]
[0.47831735]
```

it seems the behaviour changed with 7be282232d87d93a15d5c5ff07648e2eb6f144c6. in particular, with the change in `blackjax/inference/proposal.py`.
using `p_accept = jnp.clip(jnp.exp(proposal.weight), a_max=1)` gives with the above code
```
[-0.00980166]
[0.9992622]
```

similar, with the parent commit (i.e., c6f75e91217f3336c6b0cea33d53efa5c8c843b5)
```py
import numpy as np
import jax
import jax.numpy as jnp
import blackjax.hmc as hmc
import matplotlib.pyplot as plt

potential = lambda x: -jax.scipy.stats.norm.logpdf(x, loc=0.0, scale=1.0).squeeze()
initial_position = np.array([1.0,])
[...]

PR #47, opened 2021-06-07
Fix HMC failure on simple targets

rlouf:
See #46 for a description of the issue.

### DIFF
--- a/blackjax/inference/proposal.py
+++ b/blackjax/inference/proposal.py
@@ -30,8 +30,9 @@ class Proposal(NamedTuple):
 def proposal_generator(
     kinetic_energy: Callable, divergence_threshold: float
 ) -> Tuple[Callable, Callable]:
-    def init(state: IntegratorState) -> Proposal:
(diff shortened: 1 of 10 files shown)


blackjax/inference/proposal.py
```diff
@@ -30,8 +30,8 @@ class Proposal(NamedTuple):
 def proposal_generator(
     kinetic_energy: Callable, divergence_threshold: float
 ) -> Tuple[Callable, Callable]:
-    def init(state: IntegratorState) -> Proposal:
-        energy = state.potential_energy + kinetic_energy(state.position, state.momentum)
+    def new(state: IntegratorState) -> Proposal:
+        energy = state.potential_energy + kinetic_energy(state.momentum)
```
junpenglao, line 34: Damn, this was dump... how was this slip in lol.

junpenglao (approved):
LGTM

rlouf:
I added a test with a simple Gaussian target, although we need to think about more serious tests sooner rather than later.

PR #48, opened 2021-06-09
Use chex.dataclass instead of Namedtuple

junpenglao:
chex.dataclass behaves quite similar to Namedtuple, with the exception that:
- you cannot do positional initialization
- you cannot unpack it like Namedtuple


rlouf:
What do you think?

Issue #27, opened 2021-05-24
Add example of how to run on PyMC3 models


twiecki:
https://github.com/pymc-devs/pymc3/blob/master/pymc3/sampling_jax.py should serve as a good blue-print.

rlouf:
This can be started now but will need to wait until #26 is merged to be completed.

twiecki:
@kaustubh any interest in looking into this?

kc611:
Sure, 

Just to be clear, by running `BlackJAX` on `PyMC3` models we mean to convert an existing `pymc3.Model` completely into it's `BlackJax` (or `Jax` ?) variant, (along with it's inputs) similar to how we first approached the `Jax` conversion in PyMC, right? Or is there some sort of integration of `BlackJax` with `PyMC` already present.

twiecki:
@kc611 No, we just want to compile a PyMC3 model to JAX through Aesara's existing JAX backend and then pass that to `blackjax`. This should only require a couple of lines of code. You can take a look at https://github.com/pymc-devs/pymc3/blob/main/pymc3/sampling_jax.py#L157 how we do this for running a PyMC3 JAX model using numpyro's jax samplers. This here should look quite similar.

kc611:
Ah okay, got it.

kc611:
So, This is something that I came up with. The original model is from @twiecki 's Radon hierarchical model notebook (from PyMC docs). 

https://colab.research.google.com/drive/1lUXPRynGCuusHLpBTnC1tPw-2zw_6NwG?usp=sharing

But I don't know if it's working as expected. (Like, I haven't tried plotting the variables yet)

rlouf:
Looks good ! I think it would be nice to add a comparison in terms of sampling time with `pm.sample` and `pm.sampling_jax`? Using only one chain for fair comparison.

903124:
For PyMC v4 with Aesara backend here is the error: 
````
from aesara.link.jax.dispatch import jax_funcify
from aesara.graph.fg import FunctionGraph

seed = jax.random.PRNGKey(1234)
chains = 1


fgraph = FunctionGraph(model.free_RVs, [model.logpt])
fns = jax_funcify(fgraph)
logp_fn_jax = fns[0]

rv_names = [rv.name for rv in model.free_RVs]
init_state = [model.test_point[rv_name] for rv_name in rv_names]
[...]

PR #58, opened 2021-06-17
Added example of how to run on PyMC3 models

kc611:
This PR simply adds the notebook from #27.

### DIFF
--- a/notebooks/use_with_pymc3.ipynb
+++ b/notebooks/use_with_pymc3.ipynb
@@ -1,0 +1,424 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "397995ab"
+   },
+   "source": [
+    "# Use BlackJAX with PyMC3\n",
+    "Author: Kaustubh Chaudhari"
+   ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


rlouf:
Looks good to me. Can you add a header with `Author:` and your name before merging ?

twiecki:
This is great! I think it would be better to use the same model as from the other examples, so that we can also do speed comparisons. That can also be done in addition to the one here, but it doesn't seem to converge well.

rlouf:
@kc611 Do you need any help with this?

kc611:
Hey, sorry for (very) late response, got caught up in other stuff. I built the recommended 8-schools model in https://colab.research.google.com/drive/1lUXPRynGCuusHLpBTnC1tPw-2zw_6NwG?usp=sharing and ran into that strange error over there, I think it's due to the runtime environment (the code runs locally). Can you please confirm that ?

Also where do I add the Author header you mentioned ?

rlouf:
Indeed it runs locally so it must be an environment issue. Just add `Author: <your name>` under the title, it should do. Let me know and I'll merge.

rlouf:
Merging. Thank you for contributing!

PR #62, opened 2021-06-29
Update README.md

canyon289:
Makes it easier for people who are new to the field. Given its a front page readme feel like its appropriate. Feel free to close if not


rlouf:
Good point, merging. Thank you !

Issue #64, opened 2021-07-24
Add coupled MCMC methods


AdrienCorenflos:
Ideally implement these two different papers

- https://arxiv.org/abs/1709.00404
- https://arxiv.org/abs/2104.05134

They both build on this https://rss.onlinelibrary.wiley.com/doi/full/10.1111/rssb.12336 and can be used to compute unbiased estimates of expectations. I don't really thing these methods are implemented in any Python library (I don't think it's in Stan either?) and would be a cool show case.

rlouf:
Really cool ! You shouldn't have to reimplement the HMC logic for the simplest coupled algorithms, let me know if you find yourself doing that.

rlouf:
Very close to #38. There I think the authors run two long coupled chains and keep all samples; here if i understand well we run many coupled chains and only keep the last sample. Correct me if I'm wrong.

AdrienCorenflos:
What do you mean by "we only keep the last sample"? Of the chain or of the
integrator? 'Cause we keep the whole chains for sure.

rlouf:
You're right. I'll need to re-read the paper, afair the difference is that in Heng & Jacob's paper they want "to obtain estimators that are consistent in the limit of the number of replicates" while in the paper referenced in #38 they run two coupled chains for a long time.

But it does not matter here !

AdrienCorenflos:
Well, this swindled algo will require to open up the method to pass in
antithetic moments, so I would suggest keeping it the way I have put it if
you are planning on implementing it (although I agree that for my purpose I
could simply have run two chains with CRNs, albeit at the price of some
efficiency I presume).

rlouf:
> Well, this swindled algo will require to open up the method to pass in
> antithetic moments, 

We thought about it a while ago with @junpenglao and if I remember well the solution is to allow to pass the direction to the kernel.

AdrienCorenflos:
Isn't this complexifying the interface quite a bit though?

rlouf:
Not so much since you can _configure_ the kernel to integrated in a given direction in time when building it.

rlouf:
CC @pierrejacob

rlouf:
https://arxiv.org/abs/2104.05134

rlouf:
https://arxiv.org/abs/2206.07202

Issue #38, opened 2021-06-01
Add HMC Swindles


rlouf:
There are several algorithms in https://arxiv.org/abs/2001.05033 that should be simple to implement. The following code is not guaranteed to work and is here for inspiration.

 - *HMC-COUPLED* is a kernel that runs two identical kernels with the same `rng_key`;

```python
def hmc_coupled(rng_key, states):
    states, infos = jax.vmap(kernel, in_axis=(None, 0))(rng_key, states)
    return states, infos
```

 - *HMC-ANTITHETIC* is a kernel that runs two kernels with opposite momenta and identical `rng_key`.

rlouf:
Seemingly unrelated, but Metropolis-within-Gibbs has been shown to work well (https://github.com/blackjax-devs/blackjax/discussions/275). So this is definitely feasible.

PR #65, opened 2021-07-24
Coupled HMC methods

AdrienCorenflos:
This is a first crack at implementing coupled Markov chains as per this issue.

https://github.com/blackjax-devs/blackjax/issues/64

Apart from a big clean-up it will also require implementing Glynn-Rhee types of expectations estimates in order to effectively remove the MCMC bias. See essentially 2.3 in https://rss.onlinelibrary.wiley.com/doi/full/10.1111/rssb.12336 for the "easy" version for MH.

### DIFF
--- a/blackjax/coupled/hmc.py
+++ b/blackjax/coupled/hmc.py
@@ -11,7 +11,7 @@ import blackjax.inference.integrators as integrators
 import blackjax.inference.metrics as metrics
 import blackjax.inference.proposal as proposal
 import blackjax.inference.trajectory as trajectory
-from blackjax.hmc import generate as hmc_generate
+from blackjax.hmc import generate as hmc_generate, HMCInfo
 
 Array = Union[np.ndarray, jnp.DeviceArray]
 PyTree = Union[Dict, List, Tuple]
@@ -25,8 +25,8 @@ class CoupledHMCState(NamedTuple):
 
 
 class CoupledHMCInfo(NamedTuple):
-    info_state_1: base.HMCState
-    info_state_2: base.HMCState
+    info_state_1: HMCInfo
+    info_state_2: HMCInfo
 
 
 def new_state(position_1: PyTree, position_2: PyTree, potential_fn: Callable) -> CoupledHMCState:
--- a/blackjax/coupled/rwmh.py
+++ b/blackjax/coupled/rwmh.py
@@ -116,7 +116,7 @@ def reflected_gaussians(rng_key, inverse_mass_matrix: jnp.DeviceArray,
     Example:
     --------
     >>> rng_key = jax.random.PRNGKey(42)
-    >>> rng_keys = jax.random.split(rng_key, 1000)
+    >>> rng_keys = jax.random.split(rng_key, 10)
     >>> inverse_mass_matrix = jnp.ones((1,))
     >>> position_1 = -jnp.ones((1,))
     >>> position_2 = jnp.ones((1,))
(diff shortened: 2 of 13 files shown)


AdrienCorenflos:
I need to have a good think about refactoring to be able to reuse HMC primitives more.

AdrienCorenflos:
That was a bit harder than I expected but I think I'm fairly close to a prototype (though I'd rather we finished the SMC method before getting excited by this :D)

AdrienCorenflos:
Note that I have modified a number of core functions to make them reusable for my purpose, so this will probably need a review in terms of library coherence from you @rlouf

rlouf:
Can't you get the simpler coupling with HMC work just by sharing the same `rng_key` between the two kernels ?

rlouf:
That would make a great addition ! We should aim to make this as simple as possible and re-use as much code as possible:

- For the HMC version, we can call the same `hmc` kernel with the same `rng_key` twice. We should make this usable with any kernel in the HMC family, and create a `HMCKernel` type so users downstream get a warning if trying to use it with a different kind of kernl. We can see the coupling procedure as a higher-level kernel that takes a primitive kernel and builds a new one.
- RWMH is a good addition, better implemented in a separate PR

AdrienCorenflos:
I don't think that is gonna do. I'm also setting my eyes on the NUTS
equivalent which is slightly more complex and requires thinner handling of
the internals of the integrator.

rlouf:
Ok! I'll let you do your thing while I work on the SMC PR 🙂

AdrienCorenflos:
Yeah I probably overcomplicated it. The coupled RWMH though needs a very
specific max coupling and can't really be done the same way (the reflected
one happens to be the most efficient and easy implementation of it). But I
do think that I will make the biased coupled hmc disappear somehow.

AdrienCorenflos:
I'll have a go at this a bit more this end of week I think. Please feel free to send suggestions on impl my way in the meantime though. I am not claiming to understand it fully. I literally read the papers I'm citing over breakfast a few months back, so I may make it more convoluted that it needs to be (or too simple too!).
There is some code that was made public by @pierrejacob too, but sadly it is in R and I genuinely can't read R, so I'm coding from the (very well written) paper only. Maybe when we are done here we can somewhat find a way to compare with the original implementation somehow.

rlouf:
From what I've read, there are 3 different algorithms and 1 estimator that you would want to implement here:
1. Coupled RWMH
2. Coupled HMC (basically running two chains with the same rng key)
3. Something that couples HMC and RWMH

Here's how I would go about it:
- First implement (in this PR) coupled HMC + the estimator since it is the easiest;
- While you're at it implement in a next PR the other algorithm in #38 (COUPLED-ANTITHETIC)
- Implement RWMH in its own PR so we can design it carefully
- Implement coupled RWMH
- And finally couple HMC and RWMH

That should be enough work for the next few weeks :)

pierrejacob:
Hi all,

I am very pleased to see your interest in these methods! Let me know if you think I can be helpful in any way. Happy to chat about any of this at least.

P

AdrienCorenflos:
> Hi all,
> 
> I am very pleased to see your interest in these methods! Let me know if you think I can be helpful in any way. Happy to chat about any of this at least.
> 
> P

Hi!

I guess when I'm kinda done with the implementation (probably a few weeks) it would be good if you could step in to help with designing sanity checks/unittests, etc. 
In the meantime you can also shout at me if you see anything wrong in the thread!

rlouf:
Hey - what's the status on this ?

AdrienCorenflos:
I think I got the initial design wrong. We probably should have a chat, as
I am not sure what a good api would be. This is essentially a meta mcmc
algorithm, as it has an underlying stopping time that needs to be respected
(thereby is not adapted to the "give the user a kernel and let them use it"
paradigm). I have most of the logics blocks mind you, but we should have a
proper think about the desired API.

Le jeu. 23 sept. 2021 à 13:00, Rémi Louf ***@***.***> a
écrit :

rlouf:
Ok let's add that to the meeting's agenda then.

rlouf:
@AdrienCorenflos should we close this, open one issue per algorithm and work on the design there?

PR #66, opened 2021-07-26
Add diagnostics

junpenglao:
Add basic diagnostics
- computation of r-hat
- computation of effective sample size (not based on rank, but easy to add)
- add test to validate sampler accuracy base on monte carlo CLT.

It handles shape flexibly as chain dimension and sample dimension is arbitrary. For example, arviz requires the input array to be [chain, sample, dim0], here we can do [chain, sample, dim0, dim1, ...] or [sample, chain, dim0, dim1, ...], which works both for vmap and pmap output (or even [sample, dim0, dim1, dim2, ..., chain]) 
The diagnostics can be jitted so we can potentially put it in a larger sampling/tuning routine.

### DIFF
--- a/blackjax/diagnostics.py
+++ b/blackjax/diagnostics.py
@@ -4,7 +4,7 @@ from typing import Union
 import jax
 import jax.numpy as jnp
 import numpy as np
-from scipy.fftpack import next_fast_len
+from scipy.fftpack import next_fast_len  # type: ignore
 
 Array = Union[np.ndarray, jnp.DeviceArray]
 
@@ -120,13 +120,15 @@ def effective_sample_size(
     carry_cond = jnp.ones_like(mask0[0])
     max_t = jnp.zeros_like(mask0[0], dtype=int)
 
-    def body_fn(state, mask_t):
+    def positive_sequence_body_fn(state, mask_t):
         t, carry_cond, max_t = state
         next_mask = carry_cond & mask_t
         next_max_t = jnp.where(next_mask, jnp.ones_like(max_t) * t, max_t)
         return (t + 1, next_mask, next_max_t), next_mask
 
-    (*_, max_t_next), mask = jax.lax.scan(body_fn, (0, carry_cond, max_t), mask0)
+    (*_, max_t_next), mask = jax.lax.scan(
+        positive_sequence_body_fn, (0, carry_cond, max_t), mask0
+    )
     indices = jnp.indices(max_t_next.shape)
     indices = tuple([max_t_next + 1] + [indices[i] for i in range(max_t_next.ndim)])
     rho_hat_odd = jnp.where(mask, rho_hat_odd, jnp.zeros_like(rho_hat_odd))
@@ -135,13 +137,15 @@ def effective_sample_size(
     rho_hat_even = jnp.where(mask_even, rho_hat_even, jnp.zeros_like(rho_hat_even))
 
     # Geyer's initial monotone sequence
-    def body_fn(rho_hat_sum_tm1, rho_hat_sum_t):
+    def monotone_sequence_body_fn(rho_hat_sum_tm1, rho_hat_sum_t):
         update_mask = rho_hat_sum_t > rho_hat_sum_tm1
         next_rho_hat_sum_t = jnp.where(update_mask, rho_hat_sum_tm1, rho_hat_sum_t)
         return next_rho_hat_sum_t, (update_mask, next_rho_hat_sum_t)
 
     rho_hat_sum = rho_hat_even + rho_hat_odd
-    _, (update_mask, update_value) = jax.lax.scan(body_fn, rho_hat_sum[0], rho_hat_sum)
+    _, (update_mask, update_value) = jax.lax.scan(
+        monotone_sequence_body_fn, rho_hat_sum[0], rho_hat_sum
+    )
 
     rho_hat_even_final = jnp.where(update_mask, update_value / 2.0, rho_hat_even)
     rho_hat_odd_final = jnp.where(update_mask, update_value / 2.0, rho_hat_odd)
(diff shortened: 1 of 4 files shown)


twiecki:
Why have those in JAX and not just use them from arviz?

junpenglao:
I discussed with Remi and the decision is not to add additional dependencies.

The Jax implementation have additional advantage that the shape handling is more flexible: chain dimension and sample dimension is arbitrary (instead arviz requires the input array to be [chain, sample, dimension])

twiecki:
I think adding a dependency is always better (i.e. lower cost) than reimplementing things. And it doesn't even need to be a dependency, this library could just be focused on one thing and do that well (like unix utils): great implementations of samplers implemented in JAX.

rlouf:
I agree that, in general, it is better to have a library that does one thing and does it well. There are several reasons to add some of the diagnostics here:

- By adding a couple hundred lines of codes, we avoid introducing `matplotlib`, `pandas` and `xarray` as dependencies;
- Shape handling is more flexible;
- The measures are jit-able so they can eventually be added to more sophisticated warmup schemes.

We could not do the latter easily by using Arviz.

twiecki:
Regarding the 3 points:
* I think out of those 3 libraries you mentioned, fewer than 1% of users would have an env without `pandas` or `matplotlib`. And if you generate a posterior, you probably will also want to work with it and `xarray` with `arviz` is great for that. I think this will be mostly used as a utility library together with other ones like `numpyro`, `pymc3` etc, and at least `pymc3` also has that dependency, and both of those have `pandas` and `matplotlib`. Finally, none of those are challenging to install.
* That makes sense.
* I would add this functionality here then once you actually want to implement this.

Anyway, I don't think it's a big deal to add this here but wanted to challenge it a bit because feature creep, as we all know, is an easy trap to walk into.

rlouf:
I agree point (1) is not a strong one.

blackjax/diagnostics.py
```diff
@@ -0,0 +1,164 @@
...
+import jax.numpy as jnp
+import numpy as np
+from scipy.fftpack import next_fast_len  # type: ignore
+
+Array = Union[np.ndarray, jnp.DeviceArray]
+
+__all__ = ["potential_scale_reduction", "effective_sample_size"]
+
+
+def potential_scale_reduction(
```
rlouf, line 14: So if I want to compute the Rhat value in a problem with 2 variables, do I have to use `tree_multimap` to compute the values?

blackjax/diagnostics.py
```diff
@@ -0,0 +1,164 @@
...
+    input_array: Array, chain_axis: int = 0, sample_axis: int = 1
+):
+    """Gelman and Rubin (1992)'s potential scale reduction for computing multiple MCMC chain convergence.
```
rlouf, line 17: Could you add a reference in the docstring?

blackjax/diagnostics.py
```diff
@@ -0,0 +1,164 @@
...
+        (between_chain_variance / within_chain_variance + num_samples - 1)
+        / (num_samples)
+    )
+    return rhat_value.squeeze()
+
+
+def effective_sample_size(
+    input_array: Array, chain_axis: int = 0, sample_axis: int = 1
+):
+    """Compute estimate of the effective sample size (ess).
```
rlouf, line 55: Could you add a reference to a paper/book in which this algorithm is described?

tests/test_diagnostics.py
```diff
@@ -0,0 +1,58 @@
+"""Test MCMC diagnoistics."""
```
rlouf, line 1:
```suggestion
"""Test MCMC diagnostics."""
```

tests/test_diagnostics.py
```diff
@@ -0,0 +1,58 @@
...
+        input_list.insert(loc, elem)
+    return input_list
+
+
+@pytest.mark.parametrize("case", test_cases)
+@pytest.mark.parametrize("nchain", [2, 10])
+@pytest.mark.parametrize("event_shape", [(), (3,), (5, 7)])
+def test_rhat_ess(case, nchain, event_shape):
+    rng_key = jax.random.PRNGKey(32)
+    n_samples = 5000
```
rlouf, line 41: nitpicking: `n_samples` and `nchain` use different naming conventions

rlouf: I prefer `n_chains` (or better `num_chains`)

rlouf (commented):
This is a superficial review, I focused on the form. I still have to look at the algorithms in detail.

blackjax/diagnostics.py
junpenglao: Use `tree_map` but yes the function takes arrays as input

tests/test_diagnostics.py
junpenglao: Done

rlouf:
Great work ! Everything looks good to me, merging.

OriolAbril:
I saw this today and wanted to clarify one point, not to change the outcome here but to try to avoid confusion in the future.

> arviz requires the input array to be [chain, sample, dim0]

This is not really true, but I do understand where it comes from. The first and most important thing to keep in mind is that ArviZ is built on top of xarray. It is not build on top of numpy nor jax nor pytorch nor any other array library. In xarray dimension order plays no role by design, the only thing that matters are dimension names (as shown for example in https://github.com/arviz-devs/arviz/issues/1693#issuecomment-840681873). **What ArviZ requires is that there is one dimension _named_ `chain` and another _named_ `draw`**.

When working with array data, it is first converted to xarray format and then passed to the actual function. It is in this conversion
step where the limitation is. Currently all converters use that `chain, sample, *shape` convention to _name_ he dimensions when creating the corresponding xarray objects, but he converters can be updated at any time if someone has time for it, and anyone can write custom converters at any point.

Issue #73, opened 2021-07-29
Add random walk Metropolis-Hastings


AdrienCorenflos:
Following suggestion by @rlouf I'm splitting the coupled HMC commit a bit.
This is to tag the RWMH method.

rlouf:
Closed by #74

PR #74, opened 2021-07-29
Adding RWMH 

AdrienCorenflos:
This is to add RWMH as a base algo.
See this 
https://github.com/blackjax-devs/blackjax/issues/73


blackjax/rwmh.py
```diff
@@ -0,0 +1,124 @@
...
+        Returns
+        -------
+        The next state of the chain and additional information about the current step.
+        """
+        momentum_key, proposal_key = jax.random.split(rng_key, 2)
+        momentum = momentum_generator(momentum_key, state.position, unravel=False)
+
+        ravelled_position, unravel_fn = ravel_pytree(state.position)
+        ravelled_proposed_position = ravelled_position + momentum
+        proposed_position = unravel_fn(ravelled_proposed_position)
```
rlouf, line 110: I would simplify this as follows:

```python
_, unravel_fn = ravel_pytree(state.position)
new_position = jax.tree_util.tree_multimap(lambda q, dq: q + dq, position, unravel_fn(move_proposal))
```

blackjax/rwmh.py
```diff
@@ -0,0 +1,124 @@
...
+        proposed_potential = potential_fn(proposed_position)
+
+        u = jax.random.uniform(proposal_key)
+        p_accept = jnp.exp(state.potential_energy - proposed_potential)
+        do_accept = u < p_accept
```
rlouf, line 115: How about

```python
delta = proposed_potential - state.potential_energy
delta = jnp.where(jnp.isnan(delta), -jnp.inf, delta)
p_accept = jnp.clip(jnp.exp(delta), a_max=1)
do_accept = jax.random.bernoulli(key_proposal, p_accept)
```

blackjax/rwmh.py
```diff
@@ -0,0 +1,124 @@
...
+
+        new_position, new_potential = jax.lax.cond(
+            do_accept,
+            lambda _: (proposed_position, proposed_potential),
+            lambda _: (state.position, state.potential_energy),
+            operand=None)
+        return RWMHState(new_position, new_potential), RWMHInfo(p_accept, do_accept, proposed_position)
```
rlouf, line 122: Here you could define

```python
accept_state = RWMHState(proposed_position, proposed_potential)
reject_state = RWMHState(state.position, state.potential_energy)
```

And make `lax.cond` return either of those. I find it clearer.

blackjax/rwmh.py
```diff
@@ -0,0 +1,124 @@
...
+        The initial position of the chain
+    potential_fn: Callable
+        Target potential function of the chain
+    """
+    return RWMHState(position, potential_fn(position))
+
+
+def kernel(
+        potential_fn: Callable,
+        inverse_mass_matrix: Array,
```
rlouf, line 62: We want to pass a `proposal_generator` instead of an `inverse_mass_matrix` here. We don't want to limit the algorithm to moves drawn from a normal distribution.

rlouf (changes requested):

blackjax/rwmh.py
AdrienCorenflos: I now mostly addressed this, which made a lot of the above logic disappear. What do you think about it?

blackjax/rwmh.py
AdrienCorenflos: why do you care about the nan case? nothing can compare to nan by definition, so u < jnp.nan will always be False.

AdrienCorenflos:
Isn't the code stype supposed to be autocorrected by the pre-commit actions?

blackjax/hmc.py
```diff
@@ -5,15 +5,13 @@
 import jax.numpy as jnp
 import numpy as np
 
+from blackjax.common import Array, PyTree
```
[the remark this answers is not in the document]
rlouf: Yeah so now we have pretty bad merge conflicts.

AdrienCorenflos:
Conflicts are weird. I manually resolved them already.

rlouf:
If you do `git rebase main` you'll get many conflicts. As for pre-commit it doesn't fix automatically in the CI. You have to run it locally.

rlouf:
That's a lot better! I'll go over the PR whenever I find time.

rlouf:
Great work! Refactored the code and moved things around a little, now merging. We'll freeze features until 0.3 is released. Before releasing we'll need to merge #81, possibly #44 (if #81 does not come first) and do some light refactoring / documentation improvements.

Issue #31, opened 2021-05-28
Use chex for testing


junpenglao:
Should make testing pmap etc easier: https://github.com/deepmind/chex

PR #75, opened 2021-07-29
Add an assertion to make sure the position and momentum have the same shape

AdrienCorenflos:
I noticed that nothing enforced this, which lead to me not understanding weird results while working on the RWMH linear regression unittest.

### DIFF
--- a/blackjax/inference/metrics.py
+++ b/blackjax/inference/metrics.py
@@ -34,7 +34,7 @@ EuclideanKineticEnergy = Callable[[PyTree], float]
 
 
 def gaussian_euclidean(
-    inverse_mass_matrix: jnp.DeviceArray,
+        inverse_mass_matrix: jnp.DeviceArray,
 ) -> Tuple[Callable, EuclideanKineticEnergy, Callable]:
     r"""Hamiltonian dynamic on euclidean manifold with normally-distributed momentum.
 
@@ -96,9 +96,10 @@ def gaussian_euclidean(
 
     def momentum_generator(rng_key: jnp.ndarray, position: PyTree) -> PyTree:
         ravelled_position, unravel_fn = ravel_pytree(position)
-        assert ravelled_position.shape[0] == shape[0], \
-            f"'inverse_mass_matrix' and 'position' are not compatible. 'inverse_mass_matrix' has dimension {shape}, " \
-            f"but 'position' ravelled dimension is {ravelled_position.shape[0]}."
+        if not ravelled_position.shape[0] == shape[0]:
+            raise ValueError(f"'inverse_mass_matrix' and 'position' are not compatible. "
+                             f"'inverse_mass_matrix' has dimension {shape}, "
+                             f"but 'position' ravelled dimension is {ravelled_position.shape[0]}.")
         standard_normal_sample = jax.random.normal(rng_key, shape)
         momentum = dot(mass_matrix_sqrt, standard_normal_sample)
         momentum_unravel = unravel_fn(momentum)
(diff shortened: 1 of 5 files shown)


rlouf:
I don't know what @junpenglao thinks, but I am reluctant to add an assertion that is called each time a new momentum is drawn. This is also likely traced out when you use `jax.jit`.

AdrienCorenflos:
No, shapes are static in JAX so that will be evaluated under jitting but
only once at compilation time (aka only once when run in a scan body), and
not every time it is drawn.

rlouf:
At tracing time you mean?

AdrienCorenflos:
yeah

Le jeu. 29 juil. 2021 à 20:10, Rémi Louf ***@***.***> a
écrit :

blackjax/inference/metrics.py
```diff
@@ -95,7 +95,10 @@ def gaussian_euclidean(
         )
 
     def momentum_generator(rng_key: jnp.ndarray, position: PyTree) -> PyTree:
-        _, unravel_fn = ravel_pytree(position)
+        ravelled_position, unravel_fn = ravel_pytree(position)
+        assert ravelled_position.shape[0] == shape[0], \
+            f"'inverse_mass_matrix' and 'position' are not compatible. 'inverse_mass_matrix' has dimension {shape}, " \
+            f"but 'position' ravelled dimension is {ravelled_position.shape[0]}."
```
MarcoGorelli, line 101: this would be skipped in optimised mode, perhaps just check if they're the same and raise an error if not?

MarcoGorelli (changes requested):

blackjax/inference/metrics.py
AdrienCorenflos: What do you mean skipped in optimized mode? Also I that what I am doing is exactly checking that they are the same and raising an error otherwise no?

AdrienCorenflos:
Otherwise how would you make this happen? It _needs_ to fail as a silent error is way worse.

junpenglao:
Agree that we should probably add some assertion to safeguard the input args, but probably not this way:
- Assertion is extra Ops, we should add validate_args similar to TFP so user could decide whether they want it or not.
- we should not use python assert as it is unreliable, IIUC when jitted it only check once and the subsequent call it is useless

So I would suggest we put this on hold, and wait until we introduce chex (https://github.com/blackjax-devs/blackjax/issues/31) to the code base. Using chex to assert this will be much better (see eg: https://github.com/deepmind/distrax/blob/6da616249953dbf6aaefaeed4a8c4394e64887e4/distrax/_src/distributions/categorical.py#L52-L54 and https://github.com/deepmind/distrax/blob/6da616249953dbf6aaefaeed4a8c4394e64887e4/distrax/_src/distributions/bernoulli.py#L59)

AdrienCorenflos:
But shapes are static in JAX, so it really doesn't matter that it is only called once which as at program verification (tracing) anyway.

Also @rlouf @junpenglao JAX source does it absolutely everywhere, so I really don't understand why you are having second thoughts.

E.g.

https://github.com/google/jax/blob/59e6efdaef493d0735a4c2df68c71c18cfcabae3/jax/_src/numpy/linalg.py#L68

 https://github.com/google/jax/blob/59e6efdaef493d0735a4c2df68c71c18cfcabae3/jax/_src/numpy/linalg.py#L118

MarcoGorelli:
> What do you mean skipped in optimized mode?

see https://stackoverflow.com/a/2083834/4451315 - if this isn't just an internal check done to help the static type checker, 
I think it'd be better to explicitly raise the assertion

MarcoGorelli:
> > What do you mean skipped in optimized mode?
> 
> see https://stackoverflow.com/a/2083834/4451315 - if this isn't just an internal check done to help the static type checker,
> I think it'd be better to explicitly raise the assertion

Came across a bug related to this today: https://github.com/pandas-dev/pandas/issues/42866

AdrienCorenflos:
I didn't know this optimised thing! Learning new things everyday :D
I will raise it explicitly.

rlouf:
Alright, could you print out the jaxpr of the function just for me to be certain the condition is traced out?

PR #79, opened 2021-07-29
Gather all types in a single place

AdrienCorenflos:


blackjax/common.py
```diff
@@ -0,0 +1,6 @@
+from typing import Dict, List, Tuple, Union
```
rlouf, line 1: Could we name it `types.py` ?

rlouf:
Thank you for opening this PR. Now we are working on meta-algorithms (SMC, parallel tempering, coupling) it is time that we start thinking about types across the library.

blackjax/common.py
```diff
@@ -0,0 +1,6 @@
...
+
+import jax.numpy as jnp
+
+Array = Union[jnp.ndarray, jnp.DeviceArray, float]
```
junpenglao, line 5:
```suggestion
Array = Union[np.ndarray, jnp.DeviceArray]
```

rlouf (changes requested):
I think we can merge once the `PRNGKey` type is added.

AdrienCorenflos:
You sure you want to do this? It feels like it may be confusing for end
users...

rlouf:
How would it be confusing ? Because the name of the function that returns a key is `jax.random.PRNGKey`?

It may be for a split second for someone who edits the code, but I think would be very readable for others.

rlouf:
Interested in knowing why you thought this would be confusing.

rlouf:
Looks like you need to import the names in many files still, and fix the linting. After that we're good to merge.

junpenglao:
Nice! Thanks @AdrienCorenflos!

PR #80, opened 2021-07-29
Make sure tests run on CPU locally.

AdrienCorenflos:


rlouf:
Yeah, I didn't think about users who are using JAX on GPU. This will be deprecated with #31 that will make use of [Chex](https://github.com/deepmind/chex) to test JAX code.

rlouf:
Thank you ! Merging.

PR #81, opened 2021-07-30
Introduce chex for testing and typing

junpenglao:
- Replace Union[np.ndarray, jnp.DeviceArray] with chex.Array
- Start migrating to unittest and chex.TestCase so we can test all variants (jit, vmap, ect)


junpenglao:
This superseded #48, introducing chex.DataClass is probably better done after we are sure all the tests are in good shape

requirements-jax.txt
```diff
@@ -1,2 +1,3 @@
+chex
```
rlouf, line 1: I think that for now (in this PR) we should limit `chex` to testing, and create our own types for `Array` and `PRNGKey`. The last two are too simple to justify a dependency of the _library_ (not the tests) on `chex`. In which case the dependency name would go in `requirements-dev.txt`.

tests/test_compilation.py
```diff
@@ -2,27 +2,25 @@
...
-
-
-def potential(x):
-    GLOBAL["count"] += 1
-    return jscipy.stats.norm.logpdf(x)
-
 
 def test_hmc():
-    """The reason why this works is because JAX only reads the potential once when compiled?"""
+    @chex.assert_max_traces(n=1)
```
rlouf, line 15: much, much cleaner

tests/test_diagnostics.py
```diff
@@ -33,26 +37,48 @@ def insert_list(input_list, loc, elem):
...
+
+    def setUp(self):
+        super().setUp()
+        self.num_samples = 5000
+        self.test_seed = 32
+
+    @chex.all_variants
+    @parameterized.parameters(
+        itertools.product(test_cases, [2, 10], [(), (3,), (5, 7)])
+    )
```
rlouf, line 51: Does `absl` come with `chex`?

rlouf:
Chex is definitely useful for testing. Strategically, what part of the codebase do you think should be tested for all variants?

requirements-jax.txt
junpenglao: Sure, `Array` and `PRNGKey` are both pretty simple type alias. 
We can make a decision later whether we want chex as main dependency (e.g., is the Dataclass really that useful)

junpenglao:
Currently there are a few tests that we test jit and none-jit, those for sure.
Otherwise, It seems the rest is also straightforward to test all variant so we should do that
We should also have pmap test, maybe only the end to end sampling test.

requirements-jax.txt
rlouf: Sure, that sounds good.

rlouf: Let's deal with the types in #79 then.

junpenglao: +1, I will wait for #79 to merge

PR #100, opened 2021-08-06
Change Tempered SMC example for Rastrigin

AdrienCorenflos:


rlouf:
Why not just add? 🙃

rlouf:
I'll change the notebook to add this to the previous example and merge. This is a great illustration of the use of SMC vs NUTS.

rlouf:
Recovered the bimodal example and added it back. Great addition !

PR #107, opened 2021-08-19
Small simplification in tempered SMC

AdrienCorenflos:
Simply passing in original argument instead of readding them


AdrienCorenflos:
Interesting! Behaviour differs based on versions! TBH my test was a bit wrong, the log-likelihood is biased below in average (Jensen), but still, this should probably not happen. Do you know what changes in the python version could have an impact?

AdrienCorenflos:
This was incorporated in waste free smc

PR #111, opened 2021-08-23
Expose `target_accept` arguments to `stan_warmup.run`

### LINKED ISSUE #109: expose `target_accept` to `stan_warmup`?
For dealing with divergent transitions it can be useful to increase the target acceptance probability, e.g. in Stan's using `adapt_delta`, or in NumPyro with `target_accept_prob`. Am I right in thinking that this isn't possible with your current API?  The relevant argument seems to be set to a default value in the function,
```
def find_reasonable_step_size(
    ...
    target_accept: float = 0.65,
) -> float:
```
but this is not exposed to `stan_warmup`.

Is there a reason for this choice and/or another approach you recommend for dealing with divergences? Alternatively, if you agree to exposing `target_accept` to `stan_warmup`, I'd be happy to put in a small PR.

Thanks for the great library!

prashjet:
This PR exposes two arguments for step size adaptation to the function `stan_warmup.run`:
- `target_accept_initial` for initial step size adaptation i.e. passed to `find_reasonable_step_size`,
- `target_accept` for step size adaptation using dual averaging i.e. passed to `dual_averaging_adaptation`

Closes #109

### DIFF
--- a/blackjax/adaptation/step_size.py
+++ b/blackjax/adaptation/step_size.py
@@ -89,6 +89,8 @@ def dual_averaging_adaptation(
         Controls the weights of past steps in the current update. The scheme will
         quickly forget earlier step for a small value of `kappa`. Introduced
         in [2]_, with a recommended value of .75
+    target:
+        Target acceptance rate.
 
     Returns
     -------
@@ -181,7 +183,7 @@ def find_reasonable_step_size(
     value for the step size starting from any value, choosing a good first
     value can speed up the convergence. This heuristics doubles and halves the
     step size until the acceptance probability of the HMC proposal crosses the
-    .5 value.
+    target value.
 
     Parameters
     ----------
--- a/blackjax/stan_warmup.py
+++ b/blackjax/stan_warmup.py
@@ -31,7 +31,9 @@ def run(
     num_steps: int = 1000,
     *,
     is_mass_matrix_diagonal: bool = True,
-    initial_step_size: float = 1.0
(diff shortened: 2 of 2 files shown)


rlouf:
Sorry for being awol, I will review asap !

prashjet:
Hi @rlouf. I have no example where I'd use different values, but I thought that _someone_ might, and keeping the two arguments separate was more transparent.

I'm happy to discuss changing this - possible to default the value of `target_accept_init` to that of `target_accept`...?

rlouf:
Yes I think that for now we shouldn't anticipate a need that might not arise. We can only take `target_accept` as an argument and use its value everywhere?

If people need something very specific, they can already use the building blocks we provide directly.

prashjet:
> for now we shouldn't anticipate a need that might not arise.

Sage advice that I should heed in all my projects! I agree - it is done in the last commit.

rlouf:
This looks great ! I know `target_accept` is a common short name for `target_acceptance_rate`, but I prefer `blackjax` to be explicit when it comes to names. Would you mind rename `target_accept` to `target_acceptance_rate  ? And then squash your commits? This should be good to merge afterwards.

prashjet:
The rename is done. Could you advise me how to squash the commits myself? I know there is a 'squash and merge' option when you merge the PR, but is there some way I should squash beforehand?

rlouf:
You're right, I'll do it myself. Thank you again for contributing !

PR #124, opened 2021-09-26
Requirements proposal

canyon289:
Was a bit confused about requirements files notable

* Empty requirements file in root
* "hidden" requirements file in notebooks

Making proposal. As always my feelings wont be hurt if this gets closed without a merge


canyon289:
Fixing `requirements.txt` to match `setup.py` https://github.com/blackjax-devs/blackjax/blob/main/setup.py#L40

rlouf:
We copied what [Haiku](https://github.com/deepmind/dm-haiku) does for dependencies: installing JAX is different whether you want to use the GPU or not so we let the user install it manually. This explains why we have an empty `requirements.txt` file (a placeholder until we have a dependency other than `JAX`) and `requirements-jax.txt`. This should stay as is.

Agreed for the example requirements as one usually never goes in the `examples` folder and starts the notebook server in the root folder instead.

canyon289:
Thanks for the explainer

PR #150, opened 2021-11-16
Periodic Orbital MCMC

albcab:
Draft implementation of periodic orbital MCMC, see #146.

Still need to debug, write tests and documentation but this is the basic structure I had in mind for the algorithm. It allows for the integrators in inference/hmc/integrators.py to be used as bijections for creating the orbit. Also other continuous functions can be added to inference/periodic_orbit/bijections.py for the same purpose, as long as they output a function with input `(IntegratorState, float)` and output `IntregratorState`.


rlouf:
Great ! I'll have a look asap and see if there is anything I can do to help with structure etc

albcab:
Works only for bijections/integrators that preserve volume.

rlouf:
Hey! Is there anything I can do to help?

albcab:
Hi! 
Your opinion on these decision would be really helpful:
How general should I make the augmented target density, i.e. p(x,v)? The code now allows only for p(x,v)=p(x)p(v) since usually v is Gaussian independent from x. However, more generally we can have p(x,v) = p(x)p(v|x). This wont allow me to use the code in hmc/metrics.py. Every example in the paper uses p(x,v)=p(x)p(v).
How general should I make the bijection, i.e. f(x,v)? The code now allows only for volume preserving transformations, allowing me to use the integrators from hmc/integrators.py as bijections. If I want to have transformations that change volume I would need an object similar to `IntegratorState` that carries information about the change in volume. Then the functions in hmc/integrators.py couldn't be used directly without some modification. 
I think allowing only p(x,v)=p(x)p(v) should be fine but f(x,v) that changes volume could be useful and is worth creating a new State object similar to `IntegratorState` that carries more information.
Any other thoughts on the implementation would be great. After the aforementioned changes I can write the tests and the pull request can be reviewed and merged.

rlouf:
Here are a few comments/remarks. I haven't finished to read the paper, so these comments are mostly from a software engineering perspective. I'll make sure to read the paper in depth before I next review the PR.

> How general should I
> make the augmented target density, i.e. p(x,v)? The code now allows only for
> p(x,v)=p(x)p(v) since usually v is Gaussian independent from x. However, more
> generally we can have p(x,v) = p(x)p(v|x). 

We should limit ourselves to the simplest case for now and get something to work. Happy to include the more complicated stuff later :)

> How general should I make the bijection, i.e. f(x,v)? The code now allows only for volume
> preserving transformations, allowing me to use the integrators from
> hmc/integrators.py as bijections. 

Same. Not because it's not interesting, but because I'd like to see a simple version run first!

> Any other thoughts on the implementation would be
> great. After the aforementioned changes I can write the tests and the pull
> request can be reviewed and merged.

A few remarks:
- Make sure that arguments in the `kernel` function are passed in the same order as they are in `nuts` and `hmc` !
- Wouldn't work with the `log_weight` in the state simplify things a little?
- For the design in `periodic_orbital.py` I would take inspiration from HMC rather than NUTS, and enclose the `periodic_orbital_proposal` function in `kernel` as `generate`

And question:
- I haven't read the paper in depth, but looking at `bijections.py` it looks like a specific HMC dynamic. Would it make sense to move everything to the `hmc` folder since we use functions and types from there?
- Same caveat, but could `orbit generation` be considered as some kind of trajectory integration? (cf `hmc/trajectory.py`)

albcab:
The code now allows for simple p(x,v)=p(x)p(v) but general volume changing bijection f(x,v), hence:

> I haven't read the paper in depth, but looking at bijections.py it looks like a specific HMC dynamic. Would it make sense to move everything to the hmc folder since we use functions and types from there?

`bijections.py`'s state `BijectionState` is the same as `IntegratorState` except it includes `log_det_jac` which allows for volume changing transformations. It is a very minor modification but I think it could be useful (even though all the bijections coded now preserve volume).
 
> Make sure that arguments in the kernel function are passed in the same order as they are in nuts and hmc !

I'll fix this.

> Wouldn't work with the log_weight in the state simplify things a little?

I work with `log_weights` until the `PeriodicOrbitalState`, where I take their exponents and normalize, in order for the user to be able to use them directly for inference. 

> For the design in periodic_orbital.py I would take inspiration from HMC rather than NUTS, and enclose the periodic_orbital_proposal function in kernel as generate

I'm not sure what you mean, `periodic_orbital.py` is implemented exactly as `hmc,py` except it generates an orbit instead of a state.

> Same caveat, but could orbit generation be considered as some kind of trajectory integration? (cf hmc/trajectory.py)

Orbit generation is pretty much a trajectory, then I could just build the whole orbit at the `PeriodicOrbitalState` from a leftmost and rightmost state. Not sure if it will make it faster because I need to build the whole orbit at the end of each iteration anyway, but it could be moved to `trajectory.py`. I'll have a look at making the `Orbit` a `Trajectory` and let you know how it goes.

> And the example notebook is awesome !

Thanks! Glad you find it interesting!

rlouf:
> I work with `log_weights` until the `PeriodicOrbitalState`, where I take their
> exponents and normalize, in order for the user to be able to use them directly
> for inference.

Wouldn't it make sense for `new_orbit` to return `weight` instead of `log_weight` so we don't have to take the exponent in `periodic_orbital.py`? 

It doesn't feel right as it currently is. I would much prefer having something like:

``` python
# the output and input state have the same type
state, info = generate_orbit(direction, state)

info = PeriodicOrbitalInfo(
    info.momentum,
    jnp.mean(state.weigts),
    jnp.var(state.weigts),
) 

return state, info
``` 

> I'm not sure what you mean, `periodic_orbital.py` is implemented exactly as
> `hmc,py` except it generates an orbit instead of a state.

Actually it's fine as it is, I think I was confused by the name of the `generate` function which should be named `proposed` for the sake of having the same conventions; this makes the code easier to read.

> Orbit generation is pretty much a trajectory, then I could just build the whole
> orbit at the `PeriodicOrbitalState` from a leftmost and rightmost state. Not
> sure if it will make it faster because I need to build the whole orbit at the
> end of each iteration anyway, but it could be moved to `trajectory.py`. I'll
> have a look at making the `Orbit` a `Trajectory` and let you know how it goes.

We can leave as is for now, I need to have a much better understanding of the algorithm before even considering moving things to the `/hmc` folder. The structure is good.

albcab:
> Wouldn't it make sense for new_orbit to return weight instead of log_weight so we don't have to take the exponent in periodic_orbital.py?

Yes it would. I'll make the changes, including the name changes for readability.

tests/test_sampling.py
```diff
@@ -212,7 +250,7 @@ def logprob_fn(x):
 
     def mcse_test(self, samples, true_param, p_val=0.01):
         posterior_mean = jnp.mean(samples, axis=[0, 1])
-        ess = diagnostics.effective_sample_size(samples, chain_axis=1, sample_axis=0)
+        ess = diagnostics.effective_sample_size(samples, chain_axis=0, sample_axis=1)
```
albcab, line 253: There was an error on the original test_sampling.py on which was the axis of chain and which of samples. Also change in line 289.

albcab:
Once we decide if we make `bijections.py`/`orbit.py` part of `hmc/integrators.py`/`hmc/trajectory.py` or leave it as is, I can either move the code or make tests.

rlouf:
You can leave them where they are for now !

rlouf:
Is this good to review and/or do you need my help with anything?

rlouf:
Awesome! We'll address the design change in these PRs, don't worry about it. Would be great if you could rebase on `main` and add the kernel and notebooks to the docs :)

albcab:
Rebased to main, added kernel and notebook to sphinx docs and included sampling tests. Not sure if these tests are enough? Maybe should add more, let me know :)

albcab:
Refactored to match the new API. Also, I've simplified the API: deleted `orbit.py` or `bijection.py`. `orbit.py` is done in `periodic_orbital.py` and opted to just use the `integrators.py` API for bijections, if people want to use bijections which change volume it can be included in the future. Finally, I opted for creating the ellipsis bijection in the notebook to showcase how the API can be customized to use any bijection (that doesn't change volume, for now).

.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: "Run tests"
 
 on:
   pull_request:
-    branches [main]
+    branches: [main]
```
rlouf, line 5: Can you revert this? The fix has been merged on `main`.

blackjax/kernels.py
```diff
@@ -535,3 +535,76 @@ def step_fn(rng_key: PRNGKey, state):
...
+        is minus the potential function.
+    step_size
+        The value to use for the step size in the symplectic integrator.
+    inverse_mass_matrix
+        The value to use for the inverse mass matrix when drawing a value for
+        the momentum and computing the kinetic energy.
+    period
+        The total steps used to build the orbit.
+    bijection
+        (algorithm parameter) The symplectic integrator to use to integrate the trajectory.\
```
rlouf, line 574: nitpick: `\` at the end of the line

blackjax/kernels.py
```diff
@@ -535,3 +535,76 @@ def step_fn(rng_key: PRNGKey, state):
...
+    logprob_fn
+        The logprobability density function we wish to draw samples from. This
+        is minus the potential function.
+    step_size
+        The value to use for the step size in the symplectic integrator.
+    inverse_mass_matrix
+        The value to use for the inverse mass matrix when drawing a value for
+        the momentum and computing the kinetic energy.
+    period
+        The total steps used to build the orbit.
```
rlouf, line 572: Do you mean "The number of steps used to build the orbit"?

blackjax/mcmc/periodic_orbital.py
```diff
@@ -0,0 +1,279 @@
...
+    ----------
+    position:
+        the current values of the random variables whose posterior we want to
+        sample from. Can be anything from a list, a (named) tuple or a dict of
+        arrays. The arrays can either be Numpy or JAX arrays.
+    logprob_fn:
+        a function that returns the value of the log posterior when called
+        with a position.
+    period:
+        total steps used to build the orbit.
```
rlouf, line 60: same remark as above

blackjax/mcmc/periodic_orbital.py
```diff
@@ -0,0 +1,279 @@
...
+import blackjax.mcmc.metrics as metrics
+from blackjax.types import Array, PRNGKey, PyTree
+
+
+class PeriodicOrbitalState(NamedTuple):
+    """State of the periodic orbital algorithm.
+
+    The periodic orbital algorithm takes one orbit with weights,
+    samples from the points on that orbit according to their weights
+    and returns another weighted orbit of the same period.
```
rlouf, line 17: A high-level description of the parameters would be nice here

blackjax/mcmc/periodic_orbital.py
```diff
@@ -0,0 +1,279 @@
...
+    ----------
+    bijection
+        transformation used to build the orbit (given a step size).
+
+    Returns
+    -------
+    A kernel that takes a rng_key and a Pytree that contains the current state
+    of the chain and that returns a new state of the chain along with
+    information about the transition.
+
```
rlouf, line 105: Please add a reference to the relevant literature

blackjax/mcmc/periodic_orbital.py
```diff
@@ -0,0 +1,279 @@
...
+        )
+        direction = directions[choice_indx]
+        period = jnp.max(directions) + 1
+        direction = jnp.mod(direction + jnp.array(period / 2, int), period)
+        potential_energy = potential_energies[choice_indx]
+        potential_energy_grad = jax.tree_util.tree_map(
+            lambda p_energy_grad: p_energy_grad[choice_indx], potential_energies_grad
+        )
+
+        # #Alternatively we can sample uniformly the direction
```
rlouf, line 178: This comment is not documenting the code; please remove.

blackjax/mcmc/periodic_orbital.py
```diff
@@ -0,0 +1,279 @@
...
+        (
+            positions,
+            weights,
+            directions,
+            potential_energies,
+            potential_energies_grad,
+        ) = state
+
+        choice_indx = jax.random.choice(key_choice, len(weights), p=weights)
+        position = jax.tree_util.tree_map(
```
rlouf, line 167: plural?

rlouf:
I've made a few cosmetic comments and still need to go over the code. Great work, the PR has clearly been carefully crafted which is always appreciated!

The notebook is awesome, and it would be even better if you could add a couple of lines about why we would want to use this method over more commonly used ones (HMC, NUTS) :)

rlouf:
@albcab As per discussion could you rename the algorithm `orbital_hmc`? I think it is good to merge after that.

albcab:
@rlouf Renamed the higher level kernel and added a couple of lines to the documentation that hopefully make the naming clear.

PR #152, opened 2021-12-03
Metropolis Adjusted Langevin algorithm

jeremiecoullon:
Here's a first attempt at merging sgmcmcjax into blackjax. Sorry it took ages to get something done! I've had loads of work and paper deadlines/revisions/etc.. :( 

I added the following:

1. diffusions function:
    - `state` includes `gradient_state`. This is to store stuff that will be needed for fancier gradient estimators (SVRG for example)
    - iteration number i is set to 0. This should be stored in some kind of state
2. Gradient estimators
    - `build_grad_log_post` builds the gradient function that takes in parameters and the data (which might be a minibatch)
    - `build_gradient_estimation_fn`: returns the gradient estimator as well as the function that initialises the gradient state (the gradient at the first iteration will be based on a minibatch of data)
3. Build kernels:
    - Build ULA kernel
    - Build SGLD kernel: same as ULA but you pass in a batch size as well as a random key to initialise the gradient

I did some copy paste from sgmcmcjax, but tried to follow the API that @rlouf suggested. But I'm not sure what the best way is to define state. We need a state for the chain (like for standard mcmc), but we also need state for the gradients (not for this simple gradient estimator, but for SVRG for example). 

Anyways have a look at this is let me know what your thoughts are. I added a notebook that runs boths algorithms in `notebooks/sgld_ula.ipynb`.

### DIFF
--- a/blackjax/mcmc/diffusion.py
+++ b/blackjax/mcmc/diffusion.py
@@ -1,0 +1,39 @@
+"""Solvers for Langevin diffusions."""
+from typing import NamedTuple
+
+import jax
+import jax.numpy as jnp
+
+from blackjax.types import PRNGKey, PyTree
+
+__all__ = ["overdamped_langevin"]
+
+
+class DiffusionState(NamedTuple):
+    position: PyTree
+    logprob_grad: PyTree
+
+
+def generate_gaussian_noise(rng_key: PRNGKey, position):
+    position_flat, unravel_fn = jax.flatten_util.ravel_pytree(position)
+    noise_flat = jax.random.normal(rng_key, shape=jnp.shape(position_flat))
+    return unravel_fn(noise_flat)
+
+
+def overdamped_langevin(logprob_grad_fn):
+    """Euler solver for overdamped Langevin diffusion."""
+
+    def one_step(rng_key, state: DiffusionState, step_size: float, batch: tuple = ()):
+        position, logprob_grad = state
+        noise = generate_gaussian_noise(rng_key, position)
+        position = jax.tree_util.tree_multimap(
+            lambda p, g, n: p + step_size * g + jnp.sqrt(2 * step_size) * n,
+            position,
+            logprob_grad,
+            noise,
+        )
+
+        logprob_grad = logprob_grad_fn(position, *batch)
+        return DiffusionState(position, logprob_grad)
+
+    return one_step
--- a/blackjax/mcmc/mala.py
+++ b/blackjax/mcmc/mala.py
@@ -1,0 +1,103 @@
+"""Public API for Metropolis Adjusted Langevin kernels."""
+from typing import Callable, NamedTuple, Tuple
+
+import jax
+import jax.numpy as jnp
+
+from blackjax.mcmc.diffusion import overdamped_langevin
+from blackjax.types import PRNGKey, PyTree
+
+__all__ = ["MALAState", "MALAInfo", "init", "kernel"]
+
+
+class MALAState(NamedTuple):
+    """State of the MALA algorithm.
+
+    The MALA algorithm takes one position of the chain and returns another
+    position. In order to make computations more efficient, we also store
+    the current log-probability density as well as the current gradient of the
+    log-probability density.
+
+    """
+
+    position: PyTree
+    logprob_grad: PyTree
+
+
+class MALAInfo(NamedTuple):
+    """Additional information on the MALA transition.
+
+    This additional information can be used for debugging or computing
+    diagnostics.
+
+    acceptance_probability
+        The acceptance probability of the transition.
+    is_accepted
+        Whether the proposed position was accepted or the original position
+        was returned.
+
+    """
+
+    acceptance_probability: float
+    is_accepted: bool
+
+
+def init(position: PyTree, logprob_fn: Callable) -> MALAState:
+    grad_fn = jax.grad(logprob_fn)
+    logprob_grad = grad_fn(position)
+    return MALAState(position, logprob_grad)
+
+
+def kernel():
+    """Build a MALA kernel.
+
+    Returns
+    -------
+    A kernel that takes a rng_key and a Pytree that contains the current state
+    of the chain and that returns a new state of the chain along with
+    information about the transition.
+
+    """
+
+    def transition_probability(state, new_state, step_size):
+        """Transition probability to go from `state` to `new_state`"""
+        theta = new_state.position - state.position - step_size * state.logprob_grad
+        return -0.25 * (1.0 / step_size) * jnp.dot(theta, theta)
+
+    def one_step(
+        rng_key: PRNGKey, state: MALAState, logprob_fn: Callable, step_size: float
+    ) -> Tuple[MALAState, MALAInfo]:
+        """Generate a new sample with the MALA kernel.
+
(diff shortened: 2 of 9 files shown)


blackjax/sgmcmc.py
```diff
@@ -0,0 +1,17 @@
+from jax import random
```
rlouf, line 1: I am going to start the review at the user-interface as it is fairly easy to define. I don't know if you've noticed but the namespace in blackjax is so you can import kernels as `blackjax.algorithm_name`. I would thus call this file `sgld.py`. And `build_SGLD_kernel` would become `kernel`.

blackjax/sgmcmc.py
```diff
@@ -0,0 +1,17 @@
...
+
+def build_SGLD_kernel(key, step_size, loglikelihood, logprior, params_IC, data, batch_size):
+    grad_log_post = build_grad_log_post(loglikelihood, logprior, data)
+    estimate_gradient, init_gradient = build_gradient_estimation_fn(grad_log_post, data, batch_size)
+    mygrad, gradient_state = init_gradient(key, params_IC)
+    state = (params_IC, mygrad, gradient_state)
+
+    one_step = overdamped_langevin(estimate_gradient)
+
+    def kernel(rng_key, state):
```
rlouf, line 14: When I am working with SgMCMC I expect to be able to pass data iteratively instead of the whole dataset, and thus the following kernel signature:

```python
def kernel(rng_key, state, batch)
```

where `batch` contains the new data points.

blackjax/mcmc/langevin.py
```diff
@@ -0,0 +1,43 @@
...
+        noise = generate_noise(rng_key, position)
+        position = jax.tree_util.tree_multimap(
+            lambda p, g, n: p + step_size * g + jnp.sqrt(2 * step_size) * n,
+            position,
+            grad_state.logprob_grad,
+            noise,
+        )
+
+        new_state = DiffusionState(position, grad_state)
+        grad_state = GradientState(logprob_grad_fn(new_state, *batch))
```
rlouf, line 40: We could pass `jax.grad(logprob_fn)` as `logprob_grad_fn` and call `logprob_grad_fn(position, *batch)`

blackjax/mcmc/ula.py
```diff
@@ -0,0 +1,38 @@
...
+class UlaInfo(NamedTuple):
+    pass
+
+
+
+def kernel(logprob_fn: Callable, step_size: float):
+    _grad_fn = jax.grad(logprob_fn)
+
+    def grad_fn(state: DiffusionState) -> PyTree:
+        return _grad_fn(state.position)
```
rlouf, line 21: cf my comment about `logprob_grad_fn` in `langevin.py`

blackjax/mcmc/ula.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def one_step(rng_key, state):
+        info = None
+        return integrator(rng_key, state, step_size), info
+
+    return one_step
+
+
+
+def init(position: PyTree, logprob_fn: Callable) -> DiffusionState:
```
rlouf, line 33: We've added `init` before `kernel` in the other kernel files, we should follow this convention here too.

rlouf:
Looks great ! The example in the notebook can probably be a test in the test suite instead (we need a test).

blackjax/mcmc/langevin.py
jeremiecoullon: I pass in `DiffusionState` rather than position because SVRG will need the current centering value to estimate the gradient. This will be in `grad_state: GradientState` in `DiffusionState`.

If we pass in only the position here then the diffusion can't be re-used.
What do you think? Is there perhaps another solution?

blackjax/mcmc/ula.py
jeremiecoullon: (same reply as in langevin.py comment)

rlouf:
@jeremiecoullon I simplified the interface a lot to be able to merge MALA asap. We will need to change a few things for SgLD of course, but let's move more incrementally from one stable state to the next. I just need to add some documentation and this should be ready to merge.

blackjax/mcmc/mala.py
```diff
@@ -0,0 +1,63 @@
...
+import jax
+import jax.numpy as jnp
+
+from blackjax.mcmc.diffusion import overdamped_langevin
+from blackjax.types import PRNGKey, PyTree
+
+__all__ = ["MALAState", "MALAInfo", "init", "kernel"]
+
+
+class MALAState(NamedTuple):
```
jeremiecoullon, line 13: Nice! It makes sense to keep this simple for now rather than worry too much about accomodation SVRG later on.

blackjax/mcmc/mala.py
```diff
@@ -0,0 +1,63 @@
...
+        """Transition probability to go from `state` to `new_state`"""
+        theta = new_state.position - state.position - step_size * state.logprob_grad
+        return -0.25 * (1.0 / step_size) * jnp.dot(theta, theta)
+
+    def one_step(rng_key: PRNGKey, state: MALAState) -> Tuple[MALAState, MALAInfo]:
+
+        new_state = integrator(rng_key, state, step_size)
+
+        delta = (
+            logprob_fn(new_state.position)
```
jeremiecoullon, line 43: Ideally, the log-probability is calculated in the `integrator` call above alongside the gradient (as it would be cheaper computationally to calculate both in one go). This would mean adding `log_probability` to `DiffusionState`, and well as using `value_and_grad` in `overdamped_langevin`.

Do you think this should be done now or as an imrovement later on?

rlouf: Yep that's a good idea, just made the change.

blackjax/mcmc/diffusion.py
```diff
@@ -0,0 +1,39 @@
...
+
+__all__ = ["overdamped_langevin"]
+
+
+class DiffusionState(NamedTuple):
+    position: PyTree
+    logprob_grad: PyTree
+
+
+def generate_noise(rng_key: PRNGKey, position):
```
junpenglao, line 17: nit:
```suggestion
def generate_gaussian_noise(rng_key: PRNGKey, position):
```

blackjax/mcmc/mala.py
```diff
@@ -0,0 +1,103 @@
...
+        """Transition probability to go from `state` to `new_state`"""
+        theta = new_state.position - state.position - step_size * state.logprob_grad
+        return -0.25 * (1.0 / step_size) * jnp.dot(theta, theta)
+
+    def one_step(
+        rng_key: PRNGKey, state: MALAState, logprob_fn: Callable, step_size: float
+    ) -> Tuple[MALAState, MALAInfo]:
+        """Generate a new sample with the MALA kernel.
+
+        TODO expand the docstring.
```
junpenglao, line 72: Are you planning to resolve this before submitting?

junpenglao (approved):

blackjax/mcmc/mala.py
rlouf: Not yet, I am not sure the level of detail we should get into here.

rlouf:
I need to close because I deleted a commit and github did not take it into account in the PR (but it did on the originating branch),

PR #149, opened 2021-11-05
Fix SMC notebook as per issue https://github.com/blackjax-devs/blackjax/issues/148

AdrienCorenflos:

PR #153, opened 2021-12-27
WIP: Refactor nuts and hmc kernels for better jit speedup

zaxtax:
The following is a work-in-progress to speed-up blackjax following a proposal outlined at https://github.com/blackjax-devs/blackjax/wiki/Meeting-minutes#new-kernel-api

Not all tests have been rewritten to pass yet. But enough works that early feedback is warranted.


rlouf:
I think that if we manipulate a more general kernel such as the one implemented in #149 we should have finer control over the `jit` compilation and performance should thus improve? I am not a big fan of the closure pattern for parameters anymore.

Issue #154, opened 2021-12-30
Contribute ! :hugs: 


rlouf:
## Documentation :book: 

* We need documentation *now*.
* It would be nice if the documentation also acts like a literature review. Ie: give the pros/cons of each algorithm and tuning recommendations. Similar to [sgmcmcjax's sampler page](https://sgmcmcjax.readthedocs.io/en/latest/all_samplers.html) or to this [list of MCMC algorithms](https://m-clark.github.io/docs/ld_mcmc/index_onepage.html). The idea is that BlackJAX could attract users even just with the documentation/lit review.

## Examples :ferris_wheel: 

* Any comparison with optimization
* Comparisons between sampling algorithm
* Anything with many dimensions / lots of datapoints
* Anything that explains how the algorithms work

## Algorithms :hammer: 

- Likelihood-free inference, in particular one-shot SMC ABC
- Riemannian HMC (implicit and explicit integration)
- Slice sampling
- Variational inference:
    - Mean field approximation
    - [SVGD](https://arxiv.org/abs/1608.04471)
- Some kind of "ensemble" method (note that ensemble can mean many different things), such as [Ensemble preconditing MCMC](https://arxiv.org/abs/1607.03954)

## Adaptation :sparkles: 

In practice adaptation algorithms (that compute reasonable values for the sampling algorithms' parameters) are as important as samplers.

- [ChEEs](https://proceedings.mlr.press/v130/hoffman21a)
- [Empirical HMC](https://arxiv.org/abs/1810.04449)

## Meta-algorithms :cloud: 

- Gibbs sampling
- HMC coupling
- Parallel tempering (repplica exchange)

## Testing :mag_right: 

- [Simulation Based Calibration](https://arxiv.org/pdf/1804.06788.pdf)
- [Kernel goodness-of-fit test](http://proceedings.mlr.press/v48/chwialkowski16.pdf)

PR #155, opened 2021-12-31
Sphinx documentation

horaceg:
Hello, 

I don't know much about Blackjax but thought I could help setup a sphinx doc seing https://github.com/blackjax-devs/blackjax/issues/154

Build with e.g. `sphinx-build -b html docs docs/_build/html` and then `open docs/_build/html/index.html`

Current problems:
- The modules structure is a bit large, so I only documented part of it for now (see `docs/api.rst`).
- NamedTuple: the parameters are thrice in the documentation: once as properties, once as parameters, and once for type hints
- References don't work 
- The HierarchicalBNN notebook contains base64-encoded images that don't play well with myst-nb. I thus removed it from the doc for now, but it is still parsed so it slows the process down.
-  ~The `use_with_pymc3` notebook has incorrect header sections, so it messes with the notebook stuff~
- The Github README is included but with the name `Blackjax`, so it is not very explicit


rlouf:
Wow that's amazing! Is there any chance you could add a github workflow to build the doc automatically whenever something is pushed on main? There should be a recipe already in https://github.com/rlouf/mcx

rlouf:
I see you are using `myst-nb` to include notebooks in the documentation. Is there anyway we could do the same for notebooks in a jupytext format? 

I eventually want to get rid of `.ipynb` files in the the repo *but* want to keep a rendered version of the example notebooks somewhere.

horaceg:
> I see you are using `myst-nb` to include notebooks in the documentation. Is there anyway we could do the same for notebooks in a jupytext format?
> 
> I eventually want to get rid of `.ipynb` files in the the repo _but_ want to keep a rendered version of the example notebooks somewhere.

I think it is possible given https://myst-nb.readthedocs.io/en/latest/examples/custom-formats.html#using-jupytext

However in jupytext format there is no output so they will need to be run every time the documentation is generated, which I guess is not very convenient if you need to run them on beefy machines or if the execution time is large. Is it the case ?

There are some options to do caching of this stuff, but I guess it would be hard to share the cache across machines : https://myst-nb.readthedocs.io/en/latest/use/execute.html#triggering-notebook-execution

It seems that JAX is doing this, but it adds a layer of complexity to the process https://jax.readthedocs.io/en/latest/developer.html#update-documentation

rlouf:
It does add a layer of complexity, but having `.ipynb` in the repo is already hard to manage collaboration-wise. I suggest we do it in another PR as this is a separate concern. I was just curious.

rlouf:
@horaceg I finally had time to look at your PR in details. This looks great for a first pass! Since I am a bit more familiar woth BlackJAX I'll take it from there for the content of the API reference and the front page. I'll have a look at the outstanding issues as well. Thank you for your help.

rlouf:
It is not perfect but it is already a net improvement to the library, so merging. Thank you again for your help.

rlouf:
Just a quick note to say that I've played with the documentation recently, and `MyST-NB` was a very good pick. Thank you again!

PR #159, opened 2022-01-22
Simplify the user API & create a more general kernel

### LINKED ISSUE #133: Make kernel factories return NamedTuples
[Optax](https://github.com/deepmind/optax) has this very nice design where their optimizer factories returns the following NamedTuple:

```python
class GradientTransformation(NamedTuple):
  """Optax transformations consists of a function pair: (initialise, update)."""
  init: TransformInitFn
  update: TransformUpdateFn
```

See [the code](https://github.com/deepmind/optax/blob/3d666d92211dc0e2689c3acf82656d6286952733/optax/_src/base.py#L47). This would fit the design of our kernel factories quite well and simplify the API. I was thinking:

```python
class SamplingAlgorithm(NamedTuple):
    init: Callable
    sample: Callable

hmc = blackjax.hmc(logprob_fn, **params)

state = hmc.init(position)
new_state, info = hmc.sample(rng_key, state)
```

Opening the issue to get your opinion on the design **and** the naming. Here we don't have an *update* since we keep every sample.

rlouf:
In this PR I'm exploring a new way for users to interface with the code. To use the `hmc` algorithm we currently have to write:

```python
import blackjax.hmc as hmc

state = hmc.new_state(logprob_fn)
kernel = hmc.kernel(logprob_fn, **params)
```
Which not only requires to pass the `logprob_fn` as an argument  twice (not a big problem), but also makes the internal organization less flexible. For instance we would like to have all HMC-related algorithms in the same module instead of needing one module per algorithm. It also makes it obvious that users need to first build a state from the position and pass said state to the kernel.

After this PR we would have:

```python
import blackjax

hmc = blackjax.hmc(logprob_fn, **params)
state = hmc.init(position)
new_state, info = hmc.update(rng_key, state)
```

We also slightly refactor the interface for the window adaptation:

```python
window_adaptation = blackjax.window_adatation(blackjax.nuts, logprob_fn, num_steps=1000)
state, kernel, _ = window_adaptation.run(rng_key, position)
new_state, info = kernel(rng_key, state)
```
Closes #133, #168, #169

## Progress

- [x] Add a `base.py` file that describes the types
- [x] Add an `mcmc.py` file that contains the high-level interface with the MCMC algorithms
    - [x] The HMC kernel
    - [x] The NUTS kernel
    - [x] The Gaussian Random walk kernel
- [x] Add a wrapper around `DeviceArray`s to be able to specify arrays as static arguments to JIT
      (not as simple as initially thought)
- [x] Update the design of the warmup (`adaptation.py`)
- [x] Add regresion tests on the number of times the logprob is traced
- [x] Add tests for high-level API
- [x] ~~Pass `num_steps` to the `run` function for the warmup~~ (faster passed before)
- [x] Polish the internal API of the adaptation
- [x] Add simple performance test to prevent regressions
- [x] Update the design of SMC kernels (`smc.py`)
- [x] update the notebooks:
    - [x] Intrduction
    - [x] BNN
    - [x] Linear Regression
    - [x] Tempered SMC
    - [x] TFP
    - [x] Numpyro
    - [x] PyMC
- [x] Tidy the internals
- [x] Update the README
- [x] Update the documentation
- [x] Restore tests in CI
- [x] Clean the commit history


rlouf:
I have successfully added the base interface types and updated HMC, NUTS and random walk RMH. One would now use HMC in the following way:

```python
import blackjax

hmc = blackjax.hmc(logprob_fn, step_size, inverse_mass_matrix, num_integration_steps)
state = hmc.init(position)
new_state, info = hmc.update(rng_key, state)
```

There is an important issue with this design when we use adaptation algorithms to get values for the algorithms' parameters. Consider the NUTS algorithm (see `HierarchicalBNN` notebook):

```python
import blackjax

initial_state = nuts.new_state(initial_position, logprob_fn)

kernel_generator = lambda step_size, inverse_mass_matrix: nuts.kernel(
    logprob_fn, step_size, inverse_mass_matrix
 )

final_state, (step_size, inverse_mass_matrix), info = stan_warmup.run(
    warmup_key, kernel_generator, initial_state, num_warmup
)

nuts_kernel = jax.jit(nuts.kernel(logprob_fn, step_size, inverse_mass_matrix))
```

While we currently only need `logprob_fn` to build the initial state, the new design also requires the value of other parameters: the step size, the inverse mass matrix, etc. Most of the time, however, we do not want to have to specify these values and use adaptation algorithms instead. This is not practical.

rlouf:
@junpenglao Part of this may be a bit ouf of scope for the current PR, but I think the following higher-level interface would be nice, and closer to Optax's :

```python
import blackjax

# Option 1: The most flexible. We only specify `logprob_fn` at init but need to pass the value of parameters at every step.
nuts = blackjax.nuts(logprob_fn)
state = nuts.init(position)
new_state, info  = nuts.update(rng_key, state, step_size, inverse_mass_matrix)

# Option 2: We specify the algorithm parameters at init, either manually or using the window adaptation.
# This would be the most used interface
warmup = blackjax.window_adptation(blackjax.nuts, logprob_fn)
(step_size, inverse_mass_matrix), state, info = warmup.run(rng_key, state)

nuts = blackjax.nuts(logprob_fn, step_size, inverse_mass_matrix)
state = nuts.init(position)
new_state, info  = nuts.update(rng_key, state)
```

We do not support intermediate solutions. Internals would look like:

```python
from typing import Callable, NamedTuple

class SamplingAlgorithm(NameTuple):
    init: Callable
    update: Callable

def nuts(logprob_fn: Callable, step_size: Optional, inverse_mass_matrix: Optional):
    if step_size and inverse_mass_matrix:
        return lambda key, state: kernel(rng_key, state, inverse_mass_matrix, step_size)
    elif not step_size and not inverse_mass_matrix:
        return kernel
    else:
        raise AttributeError("Blackjax allows you to build algorithm specifying either all parameters at initialization or none, but specifying only one parameter is not supported.")

def kernel(rng_key, state, inverse_mass_matrix, step_size):
    pass
```

rlouf:
Option 1 above is out of scope for this PR so I implemented a workaround for now:

```python
import blackjax

# Option 1: The most flexible. We only specify `logprob_fn` at init but need to pass the value of parameters at every step.
nuts = blackjax.nuts(logprob_fn)
state = nuts.init(position)
new_state, info  = nuts.kernel(step_size, inverse_mass_matrix)(rng_key, state)

# Option 2: We specify the algorithm parameters at init, either manually or using the window adaptation.
# This would be the most used interface
nuts = blackjax.nuts(logprob_fn, step_size, inverse_mass_matrix)
state = nuts.init(position)
new_state, info  = nuts.step(rng_key, state)
```

junpenglao:
I see, so the internal call pattern is always
```python
... = kernel(rng_key, state, inverse_mass_matrix, step_size)
```
and we provide another api to generate closure around it.

rlouf:
That's correct, I think that *internally* we should use this general `kernel(rng_key, state, **parameters)` pattern, and provide a high-level API (purely for convenience) that specifies these values when the algorithm is initialized. What do you think?

junpenglao:
Yeah I am 100% onboard with the general `kernel(rng_key, state, **parameters)` pattern.
Plus it is not _that_ difficult to have high level api handles intermedia, maybe we can even use the new case pattern https://www.python.org/dev/peps/pep-0634/

rlouf:
I agree, although I'm not sure that we *should* do it event if it is *simple*. Sticking to a strict batteries-included vs you're on your own philosophy will save us all many headaches I believe.

rlouf:
@junpenglao I implemented the new interface. `nuts(logprob_fn, **parameters)` returns a namedtuple with an `init` and a `step` function. `nuts_kernel` return a more general `kernel(rng_key, state, logprob_fn, step_size, inverse_mass_matrix)` kernel. Not sure how to interface this with the warmup yet, but it would be great if you could have a look at it already.

blackjax/hmc.py
```diff
@@ -50,26 +64,61 @@ class HMCInfo(NamedTuple):
...
+
+    def init_fn(position: PyTree):
+        return jax.jit(hmc_init, static_argnums=(1,))(position, logprob_fn)
+
+    def step_fn(rng_key: PRNGKey, state: HMCState) -> Tuple[HMCState, HMCInfo]:
+        # `np.ndarray` and `DeviceArray`s are not hashable and thus cannot be used as static arguments.`
+        # Workaround: https://github.com/google/jax/issues/4572#issuecomment-709809897
+        kernel_fn = jax.jit(
+            kernel,
+            static_argnums=(2, 3, 5),
```
junpenglao, line 88:
```suggestion
            static_argnames=['logprob_fn', 'step_size', 'num_integration_steps'],
```

blackjax/base.py
```diff
@@ -0,0 +1,98 @@
...
+from typing_extensions import Protocol
+
+from .types import PRNGKey, PyTree
+
+Position = PyTree
+State = NamedTuple
+Info = NamedTuple
+
+
+class InitFn(Protocol):
```
junpenglao, line 12: Why typing Protocal here and not below?

blackjax/nuts.py
```diff
@@ -53,23 +56,41 @@ class NUTSInfo(NamedTuple):
...
+
+    kernel = nuts_kernel(integrator, divergence_threshold, max_num_doublings)
+
+    def init_fn(position: PyTree):
+        return jax.jit(hmc_init, static_argnums=(1,))(position, logprob_fn)
+
+    def step_fn(rng_key: PRNGKey, state: HMCState) -> Tuple[HMCState, NUTSInfo]:
+        # `np.ndarray` and `DeviceArray`s are not hashable and thus cannot be used as static arguments.`
+        # Workaround: https://github.com/google/jax/issues/4572#issuecomment-709809897
+        kernel_fn = jax.jit(kernel, static_argnums=(2, 3))
```
junpenglao, line 77:
```suggestion
        kernel_fn = jax.jit(kernel, static_argnames=['logprob_fn', 'step_size'])
```

blackjax/rmh.py
```diff
@@ -26,7 +24,38 @@ def kernel(logprob_fn: Callable, sigma: Array):
...
+
+    def init_fn(position: PyTree):
+        return jax.jit(rmh_init, static_argnums=(1,))(position, logprob_fn)
+
+    def step_fn(
+        rng_key: PRNGKey, state: base.RMHState
+    ) -> Tuple[base.RMHState, base.RMHInfo]:
+        # `np.ndarray` and `DeviceArray`s are not hashable and thus cannot be used as static arguments.`
+        # Workaround: https://github.com/google/jax/issues/4572#issuecomment-709809897
+        kernel_fn = jax.jit(kernel, static_argnums=(2))
```
junpenglao, line 43:
```suggestion
        kernel_fn = jax.jit(kernel, static_argnames=['logprob_fn', 'sigma'])
```

blackjax/hmc.py
```diff
@@ -50,26 +64,61 @@ class HMCInfo(NamedTuple):
...
+
+def hmc_init(position: PyTree, logprob_fn: Callable):
+    def potential_fn(x):
+        return -logprob_fn(x)
+
+    potential_energy, potential_energy_grad = jax.value_and_grad(potential_fn)(position)
+    return HMCState(position, potential_energy, potential_energy_grad)
+
+
+def hmc_kernel(
```
junpenglao, line 110: Could you add some test case of calling `hmc_kernel` directly? (Same for NUTS and RMH)

blackjax/base.py
rlouf: That's an excellent question, I forgot and I think that's why I'm getting type errors.

rlouf:
I've made the modifications we discussed. The high level kernels are in `blackjax/mcmc.py` and you can now use `blackjax` the following way:

```python
import jax
import jax.numpy as jnp
import numpy as np

import blackjax

observed = np.random.normal(10, 20, size=1_000)
def logprob_fn(x):
  logpdf = stats.norm.logpdf(observed, x["loc"], x["scale"])
  return jnp.sum(logpdf)
```

Use the HMC kernel:

```python
step_size = 1e-3
inverse_mass_matrix = jnp.array([1., 1.])
num_integration_steps = 10

hmc = blackjax.hmc(logprob_fn, step_size, inverse_mass_matrix, num_integration_steps)
```

```python
initial_position = {"loc": 1., "scale": 2.}
state = hmc.init(initial_position)
new_state, info = hmc.step(rng_key, state)
```

Use the NUTS kernel:

```python
nuts = blackjax.nuts(logprob_fn, step_size, inverse_mass_matrix)
state = nuts.init(initial_position)
new_state, info = nuts.step(rng_key, state)
```

And the gaussian random walk:

```python
sigma = jnp.array([0.1, 0.1])

rmh = blackjax.rmh(logprob_fn, sigma)
state = rmh.init(initial_position)
new_state, info = rmh.step(rng_key, state)
```

I will now update the adaptation API (in `blackjax/adaptation.py`), and the SMC API (in `blackjax/smc.py`).

blackjax/mcmc.py
```diff
@@ -0,0 +1,160 @@
...
+import jax
+
+from blackjax.base import SamplingAlgorithm
+from blackjax.hmc_base import HMCInfo, HMCState, hmc_init, hmc_kernel
+from blackjax.inference.hmc import integrators
+from blackjax.nuts_base import NUTSInfo, nuts_kernel
+from blackjax.rmh_base import RMHInfo, RMHState, rmh_init, rmh_kernel
+from blackjax.types import Array, PRNGKey, PyTree
+
+__all__ = ["hmc", "nuts"]
```
junpenglao, line 12:
```suggestion
__all__ = ["hmc", "nuts", "rmh"]
```

rlouf:
Here is the adaptation API we're aiming for:

```python
warmup = blackjax.stan_warmup(algorithm, logprob_fn, *args)
new_state, (step_size, inverse_mass_matrix), info = warmup.run(rng_key, state, num_steps)
```

Where `args` is such that we can later build a new kernel with:

```
algo = algorithm(logprob_fn, step_size, inverse_mass_matrix, *args)
```

junpenglao:
Suggestion:
```
warmup = blackjax.stan_warmup(algorithm, logprob_fn, ***args)
new_state, updated_args, info = warmup.run(rng_key, state, num_steps)
algo = algorithm(logprob_fn, ***updated_args)
```

blackjax/__init__.py
```diff
@@ -1,14 +1,18 @@
...
     "hmc",
     "nuts",
+    "rmh",
     "adaptive_tempered_smc",
     "tempered_smc",
-    "rmh",
-    "stan_warmup",
+    "window_adaptation",
+    "ess",
+    "ehat",
```
junpenglao, line 15:
```suggestion
    "rhat",
```

blackjax/hmc_base.py
```diff
@@ -50,26 +63,26 @@ class HMCInfo(NamedTuple):
...
-def kernel(
-    logprob_fn: Callable,
-    step_size: float,
-    inverse_mass_matrix: Array,
-    num_integration_steps: int,
-    *,
+def hmc_kernel(
     integrator: Callable = integrators.velocity_verlet,
-    divergence_threshold: int = 1000,
+    divergence_threshold: float = 100,
```
junpenglao, line 76: Is changing this to 100 intended?

blackjax/inference/hmc/metrics.py
```diff
@@ -34,9 +34,7 @@
 EuclideanKineticEnergy = Callable[[PyTree], float]
 
 
-def gaussian_euclidean(
-    inverse_mass_matrix: Array,
-) -> Tuple[Callable, EuclideanKineticEnergy, Callable]:
+def gaussian_euclidean() -> Tuple[Callable, EuclideanKineticEnergy, Callable]:
```
junpenglao, line 37: I think we should keep the origin signature here that takes `inverse_mass_matrix` as input and dispatch 
```python
dot, matmul = jnp.dot, jnp.matmul
```
in the if-else.

blackjax/hmc_base.py
rlouf: Nope, good catch!

rlouf:
What would you think of the following structure:

```
blackjax/
  adaptation.py  # aliases
  mcmc.py  # aliases
  smc.py  # aliases
  kernel/  # raw kernels
     hmc.py
     nuts.py
     tempered_smc.py
     adaptive_tempered_smc.py
     orbital_mcmc.py
     mala.py
     ula.py
     sgld.py
  internals/  # all the building blocks
    adaptation/
        window_adaptation.py
     hmc/
        integrator.py
        trajectory.py
     smc/
     sgmcmc/

```

zaxtax:
I don't really like having an `internals/` directory. I'd rather have the directory `hmc/` and have the public kernels live in `__init__.py`

rlouf:
Maybe this structure fits the problem better:

```
blackjax/
  alias.py  # high level aliases
  adaptation/
     step_size.py
     mass_matrix.py
     window_adaptation.py
  smc/
     smc.py
     ess_solver.py
     solver.py
  mcmc/ 
     hmc.py
     nuts.py
     tempered_smc.py
     adaptive_tempered_smc.py
     orbital_mcmc.py
     mala.py
     ula.py
     sgld.py
     trajectory.py
     integrator.py
     metrics.py
```

zaxtax:
That looks better, though I don't think it's so bad to have HMC specific things in their own directory.

rlouf:
I've found that the distinction between `hmc`, `langevin`, `random_walk` etc is not that relevant actually. They only differ in the way they build a proposal, and some algorithms mix different methods.

blackjax/inference/hmc/hmc.py
```diff
@@ -79,21 +92,46 @@ def kernel(
 
     """
 
-    def potential_fn(x):
-        return -logprob_fn(x)
+    def kernel(
```
junpenglao, line 95: I think we should rename kernel here to something else (`kernel_fn` to indicate it is a PyTree -> PyTree function?) or the top level function to `gen_kernel`

blackjax/mcmc.py
```diff
@@ -0,0 +1,189 @@
...
+    .. code:
+
+       kernel = blackjax.hmc.new_kernel(integrators.yoshida)
+       state = blackjax.hmc.init(position, logprob_fn)
+       state, info = kernel(rng_key, state, logprob_fn, step_size, inverse_mass_matrix, num_integration_steps)
+
+    """
+
+    new_kernel = staticmethod(hmc_base.kernel)
+    init = staticmethod(hmc_base.init)
```
junpenglao, line 49: NICE!

blackjax/mcmc.py
```diff
@@ -0,0 +1,189 @@
...
+        integrator: Callable = integrators.velocity_verlet,
+        divergence_threshold: int = 1000,
+    ) -> SamplingAlgorithm:
+
+        kernel = cls.new_kernel(integrator, divergence_threshold)
+
+        def init_fn(position: PyTree):
+            return cls.init(position, logprob_fn)
+
+        def step_fn(rng_key: PRNGKey, state):
```
junpenglao, line 67: Jit it?

tests/test_compilation.py
```diff
@@ -1,61 +1,123 @@
...
+        step = jax.jit(kernel.step)
 
         for _ in range(10):
             rng_key, sample_key = jax.random.split(rng_key)
-            state, _ = kernel(sample_key, state)
+            state, _ = step(sample_key, state)
 
     def test_nuts(self):
         # Log probability function was traced twice as we call it
         # at Step 0 when building a new trajectory in tree doubling.
```
junpenglao, line 50: Remove this comment

junpenglao (approved):

blackjax/adaptation/window_adaptation.py
```diff
@@ -448,4 +380,17 @@ def stan_warmup_schedule(
...
+    def schedule_fn(step):
+        """Return the current window and whether this step is the last in the window.
+
+        Passed the supposed last step we stay in the first window.
+
+        """
+        try:
+            return schedule[step]
+        except IndexError:
+            return (0, False)
```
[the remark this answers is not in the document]
rlouf: Yes and with hindsight I think it makes more sense to `scan` over an array that contains schedule information since the number of steps is known in advance.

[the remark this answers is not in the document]
junpenglao: Yeah I think that should be faster actually.

blackjax/mcmc.py
rlouf: :grin:

blackjax/mcmc_adaptation.py
```diff
@@ -0,0 +1,108 @@
...
+    window_adaptation_base,
+    window_adaptation_schedule,
+)
+from blackjax.base import AdaptationAlgorithm
+from blackjax.types import Array, PRNGKey, PyTree
+
+__all__ = ["window_adaptation"]
+
+
+def window_adaptation(
```
junpenglao, line 16: Nice. Is this function tested anywhere?

blackjax/mcmc.py
rlouf: I don't think we should actually.

blackjax/mcmc_adaptation.py
rlouf: That's the one used in `test_sampling.py`

blackjax/inference/hmc/hmc.py
rlouf: Renamed it to `one_step`. Will do for now.

blackjax/base.py
```diff
@@ -0,0 +1,112 @@
...
+        parameters and returns a new state and some information about the
+        transition.
+
+    """
+
+    init: InitFn
+    step: UpdateFn
+
+
+class RunFn(Protocol):
```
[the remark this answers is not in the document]
rlouf: This API is in flux for now, it will probably change as we add more adaptation algorithms and if we add a sampling loop.

blackjax/mcmc_adaptation.py
```diff
@@ -0,0 +1,108 @@
...
+        keys = jax.random.split(rng_key, num_steps)
+        last_state, warmup_chain = jax.lax.scan(
+            one_step,
+            (init_state, init_warmup_state),
+            keys,
+        )
+        last_chain_state, last_warmup_state = last_state
+
+        step_size, inverse_mass_matrix = final(last_warmup_state)
+        kernel = kernel_factory(step_size, inverse_mass_matrix)
```
junpenglao, line 104: This means that we cannot jit the `run` function, did you benchmark this? Is if faster to jit `one_step` like what you are doing here or jit the `run` function and move the kernel creation outside of the warmup?

rlouf: Thanks for pointing this out. I haven't benchmarked it, and should before merging.

tests/test_benchmarks.py
```diff
@@ -0,0 +1,78 @@
+"""Simple benchmarks to track potential performance regressions.
+
+(TODO) This is only a first draft. We should add the ESS per gradient evaluation, ESS /
+second and other metrics to make sure that the results are "correct", and
+obviously more models. It should also be run in CI.
```
junpenglao, line 5: Not for this PR but we should wrap it with `chex.variant`.

blackjax/mcmc_adaptation.py
rlouf: I just tried the change necessary to jit the whole `run` function and:
- I don't find it noticeably faster
- If you jit-compile the `run` function you cannot return a kernel

I think we should keep the current version, maybe refactor the internals a little bit?

rlouf:
Last round of review @junpenglao !

blackjax/kernels.py
```diff
@@ -0,0 +1,465 @@
...
+        keys = jax.random.split(rng_key, num_steps)
+        last_state, warmup_chain = jax.lax.scan(
+            one_step,
+            (init_state, init_warmup_state),
+            (keys, schedule),
+        )
+        last_chain_state, last_warmup_state = last_state
+
+        step_size, inverse_mass_matrix = final(last_warmup_state)
+        kernel = kernel_factory(step_size, inverse_mass_matrix)
```
junpenglao, line 402: I guess the issue of not being able to jit `run` will be addressed in https://github.com/blackjax-devs/blackjax/issues/171?

junpenglao (approved):

blackjax/kernels.py
rlouf: Indeed!

junpenglao:
giphy (2)

PR #164, opened 2022-02-02
Update to fold in all trajectory building into a single while_loop

junpenglao:
Pro:
- this avoids calling the integrator outside of the while_loop, and cuts the compilation half (now trace only once like HMC)

Con:
- additional conditional to handle step 0 but it is unavoidable.


tests/test_sampling.py
```diff
@@ -187,7 +187,7 @@ class MonteCarloStandardErrorTest(chex.TestCase):
 
     def setUp(self):
         super().setUp()
-        self.key = jax.random.PRNGKey(2351235)
+        self.key = jax.random.PRNGKey(20220203)
```
rlouf, line 190: :grimacing: this does not look good. Test passing depends on the seed?

tests/test_compilation.py
```diff
@@ -37,7 +37,7 @@ def logprob_fn(x):
     def test_nuts(self):
         # Log probability function was traced twice as we call it
         # at Step 0 when building a new trajectory in tree doubling.
-        @chex.assert_max_traces(n=2)
+        @chex.assert_max_traces(n=1)
```
rlouf, line 40: I like that! Any idea of the performance difference:
- For small `n_samples` where time is dominated by compilation
- For very large `n_samples` where time is dominated by running time of the jitted kernel

blackjax/inference/hmc/trajectory.py
```diff
@@ -205,8 +205,20 @@ def add_one_state(loop_state):
             new_state = integrator(trajectory.rightmost_state, direction * step_size)
             new_proposal, is_diverging = generate_proposal(initial_energy, new_state)
 
-            new_trajectory = append_to_trajectory(trajectory, new_state)
-            sampled_proposal = sample_proposal(proposal_key, proposal, new_proposal)
+            # At step 0, we always accept the proposal, since we
```
rlouf, line 208: I remember you added the extra step to correct a bug in my implementation, but I don't see any of that in the `numpyro` implementation. Did I miss something?

rlouf (commented):
Excited to see the difference in performance between this version and the current one! I am slightly concerned that you had to use a different seed for the tests to pass; shouldn't the results be exactly the same?

tests/test_sampling.py
junpenglao: the NUTS one is slightly flaky 😔
I think I choose this seed before for the test to pass so the problem is not limit to this PR. We should randomized the seed in the test (so each time it ran it use different seed but in a reproducible way), and make this test more robust.

blackjax/inference/hmc/trajectory.py
junpenglao: I believe here is where numpyro apply the same logic https://github.com/pyro-ppl/numpyro/blob/26905216c0d24b35eee2581fb2a5648730e311e8/numpyro/infer/hmc_util.py#L1018-L1030

Explaination:
```
# After the first doubling we have
o -- o

# Suppose we extend on the right side, the end result would be
         new tree
            /\ 
           /  \
o -- o -- o -- o

# To build the new tree, we take the rightmost from previous tree
o -- x
     ^ select this leaf state

# before this change, we do 1 leapfrog outside of the while_loop
o -- x --> o
           ^ enter the while_loop for tree building, append new leaf after

# now
o -- x
     ^ enter the while_loop for tree building, but at step 0 (the first step) 
       we generate the new tree (ie leave out x) with its associated proposal
```

junpenglao:
> I am slightly concerned that you had to use a different seed for the tests to pass; shouldn't the results be exactly the same?

Yeah that test is a bit flaky... good point about why the result is not exactly the same - I am actually not sure but maybe compiling the first step in the while_loop change something?

tests/test_sampling.py
rlouf: Is it the "experimental proof" or the MCSE?

blackjax/inference/hmc/trajectory.py
rlouf: Got it. We can soon write a book on sampling algorithms ^^

tests/test_compilation.py
rlouf: I think we should benchmark on a simple example like the linear regression in the tests to see if the change is overall better (decreases compile time & execution time stays the same) or if there is a trade-off.

rlouf:
Awesome, merging!

PR #165, opened 2022-02-03
Update use_with_numpyro.ipynb

junpenglao:
Update notebook to get the same diagnostic information from both numpyro and blackjax NUTS, and also setting some parameters to be the same.

This also help benchmarking as now both sampler ran in the same parameter setting.


junpenglao:
To use this notebook as simple benchmark, i first set the flag in the notebook `RUN_BENCHMARK = True`, then from terminal:
```sh
(blackjax) junpenglao-macbookpro% ipython examples/use_with_numpyro.ipynb

Running 1000 warmup steps and 5000000 sampling steps.
CPU times: user 2.76 s, sys: 40.7 ms, total: 2.8 s
Wall time: 2.79 s
CPU times: user 29.6 s, sys: 658 ms, total: 30.2 s
Wall time: 29.9 s
Blackjax:
Acceptance rate: 0.87
0.04% divergent transitions
CPU times: user 31.3 s, sys: 539 ms, total: 31.8 s
Wall time: 31.4 s
Numpyro:
Acceptance rate: 0.90
0.00% divergent transitions
Blackjax average 12.08 leapfrog per iteration.
Numpyro average 10.86 leapfrog per iteration.
```

rlouf:
Clearly the window adaptation is not optimal, cf #44. But we can already see `blackjax` is faster / gradient evaluation :)

PR #179, opened 2022-02-09
Moved dual_averaging.py to /adaptation and renamed it to optimizers.py

Bnux256:
I deleted the previous PR and started from scratch with rlouf's feedback. dual_averaging.py is now in /adaptation called optimizers.py. All tests have passed.


rlouf:
Thank you so much for your help! In the future you don't need to close your PR to modify its content, you can just push new commits and rebase at the end :)

PR #180, opened 2022-02-14
Fix example notebook to use pymc 4

zaxtax:
Updates this notebook to use pymc4 correctly. Also removes the negation in the density function as blackjax does not expect a potential function but the actual density function.

### DIFF
--- a/examples/use_with_pymc.ipynb
+++ b/examples/use_with_pymc.ipynb
@@ -7,7 +7,7 @@
     "id": "397995ab"
    },
    "source": [
-    "# Use BlackJAX with PyMC3\n",
+    "# Use BlackJAX with PyMC\n",
     "Author: Kaustubh Chaudhari"
    ]
   },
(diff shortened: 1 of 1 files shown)


twiecki (approved):

examples/use_with_pymc.ipynb
```diff
@@ -273,16 +175,27 @@
...
    "outputs": [],
    "source": [
+    "from pymc.sampling_jax import get_jaxified_graph\n",
+    "\n",
     "rvs = [rv.name for rv in model.value_vars]\n",
     "init_position_dict = model.compute_initial_point()\n",
     "init_position = [init_position_dict[rv] for rv in rvs]\n",
     "\n",
-    "logprob_fn = pm.sampling_jax.get_jaxified_logp(model)"
+    "def get_jaxified_logp(model):\n",
```
rlouf, line 189: Why not reuse the available `get_jaxified_logp` which is easier to use (and remember)?

zaxtax: `get_jaxified_logp` returns the negation of the function blackjax is expecting. There is some refactoring that needs to happen on the pymc side but this is the cleanest way to make this notebook correct.

rlouf:
LGTM, thanks for the fix!

Issue #82, opened 2021-07-31
`requirements-dev.txt` version of deps


junpenglao:
Is the version for deps in `requirements-dev.txt` necessary? Can we remove them?

rlouf:
It's just to make sure that CI and local tests run on the exact same version of these libraries. The probability of a mismatch happening is low, but can still happen if people don't update their environment frequently. What's your opinion ?

rlouf:
The introduction of `pre-commit` makes all dependencies but `pre-commit` and what is needed for tests unnecessary. Waiting for #31 to be merged to clean `requirements-dev.txt`. We should keep the version numbers to make sure the checks/tests ran locally by developers are the same than the ones ran in CI.

PR #183, opened 2022-03-27
Elliptical slice sampler

albcab:
Elliptical slice sampler https://arxiv.org/abs/1001.0175
With a simple example of Gaussian process regression, planning on doing more complex examples with GPs in the future.


rlouf:
Awesome! Can you please make sure the checks pass before I review the PR?

albcab:
I did a couple of things to fix tests and pre-commits on local devices:
- Something between Jax release [v0.3.1](https://github.com/google/jax/releases/tag/jax-v0.3.1) and [v0.3.4](https://github.com/google/jax/releases/tag/jax-v0.3.4) broke `test_metrics.py` since the the random numbers generated changed. Resolved to using `jax.random.normal` instead of printing the outputed random number directly.
- Updated pre-commit hooks’ versions.
- Reformatted according to latests version of black.
- Changed `==` for `>=` in `requirements-dev.txt` as discussed in #82.

rlouf:
Code coverage is complaining because the sampler does not seem to have corresponding tests?

albcab:
> Code coverage is complaining because the sampler does not seem to have corresponding tests?

Yes, I'll rebase and add tests.

twiecki:
Is this sampler that useful?

albcab:
It's useful for gaussian processes, specially when the likelihood is broad. (I'm pretty sure) It scales independently with dimension (same dynamics as pCN but with slice sampling instead of MH). I think its main advantages are that its tuning free and gradient free, so easy to use and cheap on the niche of problems it works (and works well) on.

twiecki:
But if it's a GP and we run it in JAX, why wouldn't gradients ever be available, in which case NUTS is likely to outperform.

rlouf:
Great work, thank you! I'll review whenever I get a chance.

> I haven't experimented with it enough to give you a specific example where it outperforms NUTS but I'm doing some research on this so if I find anything specific I'll probably make an example for the docs.

I think a comparison with NUTS would be useful for a majority of people who will come to the library, and for whom NUTS is the gold standard. I've heard good things about this sampler from GP people, but I think it is not obvious for everyone.

rlouf:
Also note: Gradient computation is very expensive for big models; this is actually where we spend most of the time. It's supposed to be cheap and very efficient on Gaussian processes.

blackjax/mcmc/elliptical_slice.py
```diff
@@ -0,0 +1,166 @@
...
+    theta: float
+    subiter: int
+
+
+def init(position: PyTree, loglikelihood_fn: Callable):
+    loglikelihood = loglikelihood_fn(position)
+    return EllipSliceState(position, loglikelihood)
+
+
+def kernel(cov_matrix: Array, mean: Array):
```
rlouf, line 53: Could you please add a reference to the relevant literature?

blackjax/mcmc/elliptical_slice.py
```diff
@@ -0,0 +1,166 @@
...
+
+    Returns
+    -------
+    A kernel that takes a rng_key and a Pytree that contains the current state
+    of the chain and that returns a new state of the chain along with
+    information about the transition.
+
+    """
+
+    cov_matrix_sqrt = jax.lax.linalg.cholesky(cov_matrix)
```
rlouf, line 69: Is this true in the limiting case where `cov_matrix` is a scalar? Check `metrics.py`

blackjax/mcmc/elliptical_slice.py
```diff
@@ -0,0 +1,166 @@
...
+    of the chain and that returns a new state of the chain along with
+    information about the transition.
+
+    """
+
+    def generate(
+        rng_key: PRNGKey, state: EllipSliceState
+    ) -> Tuple[EllipSliceState, EllipSliceInfo]:
+        position, loglikelihood = state
+        kmomentum, kunif, ktheta = jax.random.split(rng_key, 3)
```
rlouf, line 121: We tend to prefer explicit names to shortcuts when it doesn't make the code harder to read. 

`kery_momentum`, `key_uniform` and `key_theta`?

blackjax/mcmc/elliptical_slice.py
```diff
@@ -0,0 +1,166 @@
...
+            rng, thetak = jax.random.split(rng)
+            theta = jax.random.uniform(thetak, minval=theta_min, maxval=theta_max)
+            p, m = ellipsis(position, momentum, theta, mean)
+            loglikelihood = loglikelihood_fn(p)
+            theta_min = jnp.where(theta < 0, theta, theta_min)
+            theta_max = jnp.where(theta > 0, theta, theta_max)
+            subiter += 1
+            return rng, loglikelihood, subiter, theta, theta_min, theta_max, p, m
+
+        _, loglikelihood, subiter, theta, *_, position, momentum = jax.lax.while_loop(
```
rlouf, line 146: Is this guaranteed to stop or are there circumstances where we can get stuck in an infinite loop?

blackjax/mcmc/elliptical_slice.py
```diff
@@ -0,0 +1,166 @@
...
+        return (
+            EllipSliceState(position, loglikelihood),
+            EllipSliceInfo(momentum, theta, subiter),
+        )
+
+    return generate
+
+
+def ellipsis(p, m, theta, mean):
+    p, unravel_fn = jax.flatten_util.ravel_pytree(p)
```
rlouf, line 160: `p`, `m` match the usual mathematical notations, but their meaning is not necessarily obvious to someone reading the code (and mathematical notations tend to vary form one author to the other anyway). Could you change the names to something more explicit?

rlouf: And please add a docstring for this function, I certainly understand what it does, but it will not be obvious to everyone reading the code.

blackjax/mcmc/elliptical_slice.py
```diff
@@ -0,0 +1,166 @@
...
+        new value is likely to be to the previous value.
+    """
+
+    momentum: PyTree
+    theta: float
+    subiter: int
+
+
+def init(position: PyTree, loglikelihood_fn: Callable):
+    loglikelihood = loglikelihood_fn(position)
```
rlouf, line 49: Is it really the log-likelihood only?

rlouf (changes requested):
This looks great! I added a few comments and asked a few clarifying questions, but it shouldn't take too long to address.

I like that the notebook adds a GP example to the library. There are a few things missing to be able to merge it:
- The first equations are one the same line. Use `align` instead of `equation`.
- Rename `rbf` to `squared_exponential` since this is how you call it in the text
- Show a plot with the fake data
- Show posterior predictives
- Label the axes of your figures
- Add a comparison with NUTS (maybe on a bigger example to be able to see perf difference)

rlouf:
@albcab is there anything I can do to help here?

albcab:
> @albcab is there anything I can do to help here?

I've been busy trying to meet NeurIPS deadline for submissions. I'll make the changes next week and let you know if there is anything.

blackjax/mcmc/elliptical_slice.py
albcab: As the bracket `[theta_min, theta_max]` shrinks, the proposal gets closer to the original position with likelihood `log y < loglikelihood_fn(position)`. It is guaranteed to stop in a finite number of iterations as long as the log likelihood is continuous wrt the parameter we are sampling (since the parameter is a continuous random variable with Gaussian prior I'm pretty sure its guaranteed to stop in finite iterations always).

blackjax/mcmc/elliptical_slice.py
albcab: Yes, the algorithm assumes that the prior of the parameter is Gaussian (user is only able to modify its mean and covariance matrix) and slice sampling is done only on the likelihood.

blackjax/mcmc/elliptical_slice.py
rlouf: I see. Could you add this in the docstring of the body function of the while loop?

albcab:
>     * Add a comparison with NUTS (maybe on a bigger example to be able to see perf difference)

Added comparison with NUTS+window adaptation on a much larger dimensional GP regression. The improvement in time is much more evident.

rlouf:
Thank you for this great addition, both the sampler and the example notebook !

PR #184, opened 2022-03-27
iMCMC

albcab:
Involutive MCMC framework https://arxiv.org/abs/2006.16653

I am using methods inside the `State` classes to allow for easy customization by the end user. 
Includes examples of use to create an irreversible vanilla HMC algorithm.


blackjax/kernels.py
```diff
@@ -463,3 +464,67 @@ def step_fn(rng_key: PRNGKey, state):
             )
 
         return SamplingAlgorithm(init_fn, step_fn)
+
+
+class imcmc:
```
rlouf, line 469: I think this is too general to be a high-level kernel. It is fine to only implement internals.

rlouf:
Can we close this for now?

PR #190, opened 2022-03-30
Adding progress bars to window adaptation

zaxtax:
This adapts @jeremiecoullon progress bar code to allow progress bars to be displayed during adaptation. Like in numpyro they can have an adverse effect on sampling time so should be disabled for benchmarking.

If we think progress bars are out of scope for blackjax this can be refactored to take `progress_bar_scan` shaped callback instead.

### DIFF
--- a/blackjax/progress_bar.py
+++ b/blackjax/progress_bar.py
@@ -1,0 +1,91 @@
+"""Progress bar decorators for use with step functions.
+Adapted from Jeremie Coullon's blog post [1]
+.. [1]: "How to add a progress bar to JAX scans and loops"
+        https://www.jeremiecoullon.com/2021/01/29/jax_progress_bar/
+"""
+from fastprogress.fastprogress import progress_bar
+from jax import lax
+from jax.experimental import host_callback
+
+
+def progress_bar_scan(num_samples, print_rate=None):
+    "Progress bar for a JAX scan"
+    progress_bars = {}
+
+    if print_rate is None:
+        if num_samples > 20:
+            print_rate = int(num_samples / 20)
+        else:
+            print_rate = 1  # if you run the sampler for less than 20 iterations
+
+    def _define_bar(arg, transform, device):
+        progress_bars[0] = progress_bar(range(num_samples))
+        progress_bars[0].update(0)
+
+    def _update_bar(arg, transform, device):
+        progress_bars[0].update_bar(arg)
+
+    def _update_progress_bar(iter_num):
+        "Updates progress bar of a JAX scan or loop"
+        _ = lax.cond(
+            iter_num == 0,
+            lambda _: host_callback.id_tap(
+                _define_bar, iter_num, result=iter_num, tap_with_device=True
+            ),
+            lambda _: iter_num,
+            operand=None,
+        )
+
+        _ = lax.cond(
+            # update every multiple of `print_rate` except at the end
+            (iter_num % print_rate == 0),
+            lambda _: host_callback.id_tap(
+                _update_bar, iter_num, result=iter_num, tap_with_device=True
+            ),
+            lambda _: iter_num,
+            operand=None,
+        )
+
+        _ = lax.cond(
+            # update by `remainder`
+            iter_num == num_samples - 1,
+            lambda _: host_callback.id_tap(
+                _update_bar, num_samples, result=iter_num, tap_with_device=True
+            ),
+            lambda _: iter_num,
+            operand=None,
+        )
+
+    def _close_bar(arg, transform, device):
+        progress_bars[0].on_iter_end()
+        print()
+
+    def close_bar(result, iter_num):
+        return lax.cond(
+            iter_num == num_samples - 1,
+            lambda _: host_callback.id_tap(
+                _close_bar, None, result=result, tap_with_device=True
+            ),
+            lambda _: result,
+            operand=None,
+        )
+
+    def _progress_bar_scan(func):
+        """Decorator that adds a progress bar to `body_fun` used in `lax.scan`.
+        Note that `body_fun` must either be looping over `np.arange(num_samples)`,
+        or be looping over a tuple who's first element is `np.arange(num_samples)`
+        This means that `iter_num` is the current iteration number
+        """
+
+        def wrapper_progress_bar(carry, x):
+            if type(x) is tuple:
+                iter_num, *_ = x
+            else:
+                iter_num = x
+            _update_progress_bar(iter_num)
+            result = func(carry, x)
+            return close_bar(result, iter_num)
+
+        return wrapper_progress_bar
+
+    return _progress_bar_scan
--- a/blackjax/kernels.py
+++ b/blackjax/kernels.py
@@ -1,11 +1,13 @@
 from typing import Callable, Dict, Union
 
 import jax
+import jax.numpy as jnp
 
 import blackjax.adaptation as adaptation
 import blackjax.mcmc as mcmc
 import blackjax.smc as smc
 from blackjax.base import AdaptationAlgorithm, SamplingAlgorithm
+from blackjax.progress_bar import progress_bar_scan
 from blackjax.types import Array, PRNGKey, PyTree
 
 __all__ = [
@@ -389,6 +391,7 @@ def window_adaptation(
     is_mass_matrix_diagonal: bool = True,
     initial_step_size: float = 1.0,
     target_acceptance_rate: float = 0.65,
+    progress_bar: bool = False,
     **parameters,
 ) -> AdaptationAlgorithm:
     """Adapt the parameters of algorithms in the HMC family.
@@ -417,6 +420,8 @@ def window_adaptation(
         The initial step size used in the algorithm.
     target_acceptance_rate
         The acceptance rate that we target during step size adaptation.
+    progress_bar
+        Whether we should display a progress bar.
     **parameters
         The extra parameters to pass to the algorithm, e.g. the number of
         integration steps for HMC.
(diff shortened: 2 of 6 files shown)


junpenglao (approved):

rlouf:
It looks like you forgot to commit the file that contains the code for the progress bar. I don't think progress bars are out of scope, and can be part of an effort to provide user-friendly loops for the high level API, like [emcee](https://github.com/dfm/emcee) does.

blackjax/progress_bar.py
```diff
@@ -0,0 +1,85 @@
+"""Progress bar decorators for use with step functions.
+Adapted from Jeremie Coullom's blog post [1]
```
jeremiecoullon, line 2: typo: "Coullon" :)

junpenglao:
> It looks like you forgot to commit the file that contains the code for the progress bar.

Haha I though that code was already in the repo and this is just adding the highlevel user interface

jeremiecoullon:
Great idea to add this! Just a heads up that this progress bar doesn't work on TPU (Colab) for some reason (see [this issue](https://github.com/jeremiecoullon/SGMCMCJax/issues/24)); there might be some problem with the callback. But GPU and CPU is fine.

zaxtax:
> Great idea to add this! Just a heads up that this progress bar doesn't work on TPU (Colab) for some reason (see [this issue](https://github.com/jeremiecoullon/SGMCMCJax/issues/24)); there might be some problem with the callback. But GPU and CPU is fine.

Is this a tqdm problem or are there are colab notebooks where a progress bar works with jax and a TPU?

jeremiecoullon:
> > Great idea to add this! Just a heads up that this progress bar doesn't work on TPU (Colab) for some reason (see [this issue](https://github.com/jeremiecoullon/SGMCMCJax/issues/24)); there might be some problem with the callback. But GPU and CPU is fine.
> 
> Is this a tqdm problem or are there are colab notebooks where a progress bar works with jax and a TPU?

I just tried this in [this colab](https://colab.research.google.com/github/probml/probml-notebooks/blob/main/notebooks/sg_mcmc_jax.ipynb) which runs on TPU: using tqdm normally (in a Python loop) works fine. So it's `progress_bar_scan` which is the problem (so maybe the callback bit)

zaxtax:
> > > Great idea to add this! Just a heads up that this progress bar doesn't work on TPU (Colab) for some reason (see [this issue](https://github.com/jeremiecoullon/SGMCMCJax/issues/24)); there might be some problem with the callback. But GPU and CPU is fine.
> > 
> > 
> > Is this a tqdm problem or are there are colab notebooks where a progress bar works with jax and a TPU?
> 
> I just tried this in [this colab](https://colab.research.google.com/github/probml/probml-notebooks/blob/main/notebooks/sg_mcmc_jax.ipynb) which runs on TPU: using tqdm normally (in a Python loop) works fine. So it's `progress_bar_scan` which is the problem (so maybe the callback bit)

That colab is giving me errors. Am I suppose to run it in a certain way?

blackjax/progress_bar.py
zaxtax: My apologies!

zaxtax:
For some reason the test isn't installing tqdm. Also I used tqdm without really thinking? Do we want fastprogress instead as that is the progress bar library used in pymc?

jeremiecoullon:
@zaxtax : Running the colab notebook will cause an error when sampling (due to the progress bar!); is this what you found? So to get it to run you need to add `pbar=False` when building the sampler:

> my_sampler = build_sgld_sampler(dt, loglikelihood, logprior, (X_data,), batch_size, pbar=False)

rlouf:
> For some reason the test isn't installing tqdm. Also I used tqdm without really thinking? Do we want fastprogress instead as that is the progress bar library used in pymc?

Do you know why PyMC is using `fastprogress` instead of `tqdm`? @junpenglao ?

zaxtax:
Do you think the simplicity will help resolve the issue Jeremie is having
running it on colab with TPUs?

jeremiecoullon:
@zaxtax ah ok! No, for TPU the progress_bar with host_callback doesn't seem to work (though I didn't try to fix it so it may be possible).
This might be a question for the JAX team. Though the [JAX docs](https://jax.readthedocs.io/en/latest/jax.experimental.host_callback.html) does seem to have some discussion of this module being used on TPU

rlouf:
That is a nice addition to the window adaptation code, and a first step towards including simple inference loops in the library.
Codecov complains because it seems that the window adaptation is never run with `progress_bar=True`. You should add a test for that just to prevent regressions in the future.

rlouf:
So I took the progress bar for a test drive with the following code (taken from the tests):

```python
from datetime import datetime
import functools

import blackjax
import jax
import jax.numpy as jnp

def regression_logprob(scale, coefs, preds, x):
    """Linear regression"""
    logpdf = 0
    logpdf += stats.expon.logpdf(scale, 1, 1)
    logpdf += stats.norm.logpdf(coefs, 3 * jnp.ones(x.shape[-1]), 2)
    y = jnp.dot(x, coefs)
    logpdf += stats.norm.logpdf(preds, y, scale)
    return jnp.sum(logpdf)

key_data, rng_key = jax.random.split(rng_key, 2)

# Generate the data
x_data = jax.random.normal(key_data, shape=(100_000, 1))
y_data = 3 * x_data + jax.random.normal(key_data, shape=x_data.shape)

# Generate the logprosterior
logposterior_fn_ = functools.partial(
    regression_logprob, x=x_data, preds=y_data
)
logposterior_fn = lambda x: logposterior_fn_(**x)

num_samples = 50_000
initial_position = {"scale": 1.0, "coefs": 2.0}
```

I first tried the warmup for NUTS _without the progress bar_:

``` python
start = datetime.now()

warmup_key, inference_key = jax.random.split(rng_key, 2)
warmup = blackjax.window_adaptation(
    blackjax.nuts,
    logposterior_fn,
    num_samples,
    True,
    progress_bar=False
)
state, kernel, _ = warmup.run(
    warmup_key,
    initial_position,
)
end = datetime.now()
end-start
```

And then ran the code _with the progress bar_:

``` python
start = datetime.now()

warmup_key, inference_key = jax.random.split(rng_key, 2)
warmup = blackjax.window_adaptation(
    blackjax.nuts,
    logposterior_fn,
    num_samples,
    True,
    progress_bar=True
)
state, kernel, _ = warmup.run(
    warmup_key,
    initial_position,
)
end = datetime.now()
end-start
```

Both ran in ~30 seconds on my machine, so at this refresh rate there is no noticeable performance difference.

However I have found that refreshing every `num_samples / 30` leads to frustratingly low refresh rates for a large number of samples. Ideally you'd want to refresh on a time basis: every second or some kind of sweet spot between user frustration and performance. Do you think that would be possible?

zaxtax:
> So I took the progress bar for a test drive with the following code (taken from the tests):
> 
> ```python
> from datetime import datetime
> import functools
> 
> import blackjax
> import jax
> import jax.numpy as jnp
> import jax.scipy.stats as stats
> 
> def regression_logprob(scale, coefs, preds, x):
>     """Linear regression"""
>     logpdf = 0
>     logpdf += stats.expon.logpdf(scale, 1, 1)
>     logpdf += stats.norm.logpdf(coefs, 3 * jnp.ones(x.shape[-1]), 2)
>     y = jnp.dot(x, coefs)
>     logpdf += stats.norm.logpdf(preds, y, scale)
>     return jnp.sum(logpdf)
> 
> rng_key = jax.random.PRNGKey(0)
> key_data, rng_key = jax.random.split(rng_key, 2)
> 
> # Generate the data
> x_data = jax.random.normal(key_data, shape=(100_000, 1))
> y_data = 3 * x_data + jax.random.normal(key_data, shape=x_data.shape)
> 
> # Generate the logprosterior
> logposterior_fn_ = functools.partial(
>     regression_logprob, x=x_data, preds=y_data
> )
> logposterior_fn = lambda x: logposterior_fn_(**x)
> 
> num_samples = 50_000
> initial_position = {"scale": 1.0, "coefs": 2.0}
> ```
> 
> I first tried the warmup for NUTS _without the progress bar_:
> 
> ```python
> start = datetime.now()
> 
> warmup_key, inference_key = jax.random.split(rng_key, 2)
> warmup = blackjax.window_adaptation(
>     blackjax.nuts,
>     logposterior_fn,
>     num_samples,
>     True,
>     progress_bar=False
> )
> state, kernel, _ = warmup.run(
>     warmup_key,
>     initial_position,
> )
> end = datetime.now()
> end-start
> ```
> 
> And then ran the code _with the progress bar_:
> 
> ```python
> start = datetime.now()
> 
> warmup_key, inference_key = jax.random.split(rng_key, 2)
> warmup = blackjax.window_adaptation(
>     blackjax.nuts,
>     logposterior_fn,
>     num_samples,
>     True,
>     progress_bar=True
> )
> state, kernel, _ = warmup.run(
>     warmup_key,
>     initial_position,
> )
> end = datetime.now()
> end-start
> ```
> 
> Both ran in ~30 seconds on my machine, so at this refresh rate there is no noticeable performance difference.
> 
> However I have found that refreshing every `num_samples / 30` leads to frustratingly low refresh rates for a large number of samples. Ideally you'd want to refresh on a time basis: every second or some kind of sweet spot between user frustration and performance. Do you think that would be possible?

Of course. We can set the rate to be a function of elapsed time.

zaxtax:
Why is that a merge conflict?

rlouf:
That's because of #192. I will fix it before merging, don't worry.

zaxtax:
Ok I take that back, throttling based on time is not doable without calling `host_callback` every iteration. This can take a computation that was without a progress bar takes 10s and make it take 45s. I suggest merging as is or I can add print_rate as an optional argument to the decorator.

rlouf:
That's good to know! We can probably fine-tune the refresh rate in the future, in the meantime let's merge this. Great and very useful work :)

rlouf:
Merging, thanks!

PR #191, opened 2022-03-30
Implement stochastic gradient langevin dynamics

### LINKED ISSUE #131: Add SgLd
Starting from the SgMCMCJAX implementation.

rlouf:
With the gradient estimator in Robbins & Monro (1951). Closes #131.

### DIFF
--- a/blackjax/kernels.py
+++ b/blackjax/kernels.py
@@ -5,6 +5,7 @@ import jax.numpy as jnp
 
 import blackjax.adaptation as adaptation
 import blackjax.mcmc as mcmc
+import blackjax.sgmcmc as sgmcmc
 import blackjax.smc as smc
 from blackjax.base import AdaptationAlgorithm, SamplingAlgorithm
 from blackjax.progress_bar import progress_bar_scan
@@ -17,6 +18,7 @@ __all__ = [
     "nuts",
     "orbital_hmc",
     "rmh",
+    "sgld",
     "tempered_smc",
     "window_adaptation",
 ]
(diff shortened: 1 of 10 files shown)


blackjax/mcmc/sgld.py
```diff
@@ -0,0 +1,33 @@
+"""Public API for the Stochastic gradient Langevin Dynamics kernel."""
+from typing import Callable, NamedTuple
+
+from blackjax.mcmc.diffusion import overdamped_langevin
+from blackjax.types import PRNGKey, PyTree
+
+__all__ = ["SgLDState", "SgLDInfo", "init", "kernel"]
+
+
+class SgLDState(NamedTuple):
```
jeremiecoullon, line 10: nitpick: the lowercase g looks odd to me. The standard way of writing the abreviation (in papers) is "SGLD", so `SGLDState` would feel more natural to me

rlouf:
The gradient estimator and SGLD algorithms are implemented. I also re-implemented [SGMCMCJax's MLP example](https://sgmcmcjax.readthedocs.io/en/latest/nbs/BNN.html) locally. Left to do:
- Add tests for both SGLD and the gradient estimator;
- Add a notebook with the MLP example on MNIST.

blackjax/sgmcmc/gradients.py
```diff
@@ -0,0 +1,55 @@
...
+) -> Callable:
+    """Builds a simple gradient estimator.
+
+    This estimator first appeared in [1]_.
+
+    Parameters
+    ----------
+    logprior_fn
+        The log-probability density function corresponding to the prior distribution.
+    loglikelihood_fn
```
rlouf, line 20: Here I should specify the expected signature for `loglikelihood_fn` and `logprior_fn` as it is a potential source of confusion.

tests/test_sampling.py
```diff
@@ -110,21 +110,21 @@ def test_linear_regression(self, case, is_mass_matrix_diagonal):
...
     },
     {
         "algorithm": blackjax.nuts,
-        "initial_position": jnp.array(100.0),
-        "parameters": {"step_size": 0.1, "inverse_mass_matrix": jnp.array([0.1])},
+        "initial_position": jnp.array(3.0),
+        "parameters": {"step_size": 4.0, "inverse_mass_matrix": jnp.array([1.0])},
         "num_sampling_steps": 6000,
-        "burnin": 5_000,
+        "burnin": 1_000,
```
junpenglao, line 127: Could you put these changes to a separate PR?

blackjax/kernels.py
```diff
@@ -389,6 +391,59 @@ def step_fn(rng_key: PRNGKey, state):
...
+
+    def __new__(  # type: ignore[misc]
+        cls,
+        grad_estimator_fn: Callable,
+        schedule_fn: Callable,
+    ) -> SamplingAlgorithm:
+
+        step = cls.kernel(grad_estimator_fn)
+
+        def init_fn(position: PyTree, batch):
```
junpenglao, line 432: suggestion: `batch` -> `data_batch` to emphasize this is batch of data

blackjax/sgmcmc/diffusion.py
```diff
@@ -0,0 +1,39 @@
...
+def generate_gaussian_noise(rng_key: PRNGKey, position):
+    position_flat, unravel_fn = jax.flatten_util.ravel_pytree(position)
+    noise_flat = jax.random.normal(rng_key, shape=jnp.shape(position_flat))
+    return unravel_fn(noise_flat)
+
+
+def overdamped_langevin(logprob_grad_fn):
+    """Euler solver for overdamped Langevin diffusion."""
+
+    def one_step(rng_key, state: DiffusionState, step_size: float, batch: tuple = ()):
```
junpenglao, line 26: It's better to type hint `batch` as pytree.

tests/test_sampling.py
rlouf: I feel like a separate commit is enough.

rlouf:
I have a (small) dilemma for the API. Let's assume that we're doing some supervised learning, so the data consists of a design matrix `X` and result vector `y`. Let's call the parameters of the model `position`. The `logprior_fn` is only a function of `position`:

```python
def logprior_fn(position):
   pass
```

But `loglikelihood_fn` is a function of `position`, `X` and `y`. We have several choices for the API. I think that in either case the following signature for the step function is preferable:

```python
data_batch = (X, y)
new_state = sgld.step(rng_key, state, data_batch)
```

I am just not sure which of 

```python
def loglikelihood_fn(position, X, y):
   pass
```

or 

```python
def loglikelihood_fn(position, data):
   X, y = data
   pass
```

we should ask the user to provide. I'd say the latter (consistency), but I'd like your opinion nevertheless.

PR #195, opened 2022-04-04
Bugfix in test workflow

albcab:
Github was skipping testing because of bug on test workflow.


junpenglao:
Thanks!

junpenglao:
Hmmm any idea why this is triggering `the no-commit-to-branch` pre-commit error?

rlouf:
I have no idea, but I don't see any problem with my workflow config. Removing this check in #197

Issue #201, opened 2022-04-25
API to specify forward mode gradients


FedericoV:
Hi,

Thanks for all the work into writing Blackjax: the API is really elegant, and the code is very easy to follow.

I had a quick request: currently, the main API entry point of Blackjax expects the user to pass a (jax) function that computes the log posterior probability.  This function is then internally differentiated using jax primitives, to obtain the function as well as its gradient.

I have a function where the gradients computed using forward mode autodiff are close to 10x faster than those computed with reverse mode.  Unfortunately, there's no easy way to swap between forward and reverse mode autodiff without writing your own integrators (which is not that bad, just, not ideal).

There's a few different ways such an API could go: one option would be to ask users to optionally provide their own derivative, and only calculate the derivative using jax.grad if it's not provided (this seems like the most obvious way forward to me).  Another way would be to give users a choice - and then calculate the derivative internally using `jax.jacfwd` if the user requests forward mode autodiff.

junpenglao:
Either approach sounds fine to me. I slightly prefer optional user specify gradient fun, because it applies to wider user cases (including using `jax.jacfwd` and other analytic gradient implementation)

rlouf:
The first option seems to be the best way forward since, as @junpenglao said, this also allows analytically computed gradient functions to be used.

I am generally reluctant to implement anything that complicates the API, but you're the second person to ask for this possibility, and it will lead to a more uniform API when stochastic gradient MCMC is added to the library.

FedericoV:
I started working on this issue.  Will open a pull request with where I am right now.

rlouf:
Hey @FedericoV, we're deprecating this since you can register custom gradients for your functions. May I ask what your original example was? We're writing a tutorial to explain users how to do this and need an example.

FedericoV:
Hi Remi,

The example we had was a likelihood that included the solution of stiff
ODEs in its computation.  Reverse mode gradients were much slower than
forward mode ones.

PR #205, opened 2022-05-11
API for specifying custom gradients.

FedericoV:
As discussed here: https://github.com/blackjax-devs/blackjax/issues/201

I went ahead and tried adding the option to pass custom gradients, and added some basic tests that verify that things still work with custom gradients (in this case, I just used forward mode gradients).

Feedback on API very welcome - this was the simplest design I could think of.  I haven't finished changing the code everywhere, but wanted to hear comments on API before going further.

### DIFF
--- a/blackjax/mcmc/hmc.py
+++ b/blackjax/mcmc/hmc.py
@@ -1,5 +1,5 @@
 """Public API for the HMC Kernel"""
-from typing import Callable, NamedTuple, Tuple
+from typing import Callable, NamedTuple, Optional, Tuple
 
 import jax
 
@@ -63,11 +63,21 @@ class HMCInfo(NamedTuple):
     num_integration_steps: int
 
 
-def init(position: PyTree, logprob_fn: Callable):
+def init(
+    position: PyTree, logprob_fn: Callable, logprob_grad_fn: Optional[Callable] = None
+):
     def potential_fn(x):
         return -logprob_fn(x)
 
-    potential_energy, potential_energy_grad = jax.value_and_grad(potential_fn)(position)
+    if logprob_grad_fn:
+        # Need to flip the sign the same way we do for potential_fn
+        potential_energy_grad = -logprob_grad_fn(position)
+        potential_energy = potential_fn(position)
+
+    else:
+        potential_energy, potential_energy_grad = jax.value_and_grad(potential_fn)(
+            position
+        )
     return HMCState(position, potential_energy, potential_energy_grad)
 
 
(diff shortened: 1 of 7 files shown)


blackjax/mcmc/hmc.py
```diff
@@ -63,11 +63,16 @@ class HMCInfo(NamedTuple):
...
 
 
-def init(position: PyTree, logprob_fn: Callable):
+def init(position: PyTree, logprob_fn: Callable, logprob_grad_fn: Optional[Callable] = None):
     def potential_fn(x):
         return -logprob_fn(x)
 
-    potential_energy, potential_energy_grad = jax.value_and_grad(potential_fn)(position)
+    if logprob_grad_fn:
+        potential_energy_grad = -logprob_grad_fn(position)
```
junpenglao, line 71: comment to remind us why the negative sign here.

FedericoV:
Added some more tests: I see a test failure that I believe is due to a numerical issue, but, another set of eyes is welcome.

tests/test_sampling.py
```diff
@@ -244,15 +254,21 @@ def mcse_test(self, samples, true_param, p_val=0.01):
...
         )
         num_chains = 10
         initial_positions = jax.random.normal(pos_init_key, [num_chains, 2])
-        kernel = algorithm(logprob_fn, inverse_mass_matrix=true_scale, **parameters)
+        if custom_gradients:
+            logprob_grad_fn = jax.jacfwd(logprob_fn)
+            kernel = algorithm(logprob_fn, inverse_mass_matrix=true_scale, logprob_grad_fn=logprob_grad_fn,
+                **parameters)
+        else:
+            kernel = algorithm(logprob_fn, inverse_mass_matrix=true_scale, **parameters)
```
junpenglao, line 270:
```suggestion
        if custom_gradients:
            logprob_grad_fn = jax.jacfwd(logprob_fn)
        else:
            logprob_grad_fn = None
        kernel = algorithm(logprob_fn, inverse_mass_matrix=true_scale, logprob_grad_fn=logprob_grad_fn,
                **parameters)
```

rlouf:
I made a few changes (mostly naming and linting), LGTM otherwise. Is there anything else to change in the code?

FedericoV:
There was a test that was failing when switching reverse mode gradients to forward mode gradients: do you think that's a cause for concern?  It just looked like a numerical tolerance issue to me but wanted to double check.

The only other issue is that there's a little bit of repeated code, but, it's probably not worth factoring it out yet.

junpenglao:
@rlouf How do you make the test not skip? The test are not run right now
@FedericoV We should try to get a better understanding of why test fail for sure, maybe even enable the custom_grad test for a few more places.

rlouf:
It's an issue with the github workflow that I need to fix. Tests are passing locally btw.

FedericoV:
Let me know if you'd like to add any more changes.  I'm fine with merging it whenever.

rlouf:
Thank you for contributing!

FedericoV:
Awesome, glad to help!

PR #210, opened 2022-05-24
Clean up optimization code path

junpenglao:
- [x] Move dual averaging and lbfgs into optimizers
- [x] Clean up lbfgs (use scan instead of while_loop)
- [x] Refactor pathfinder init to use vmap (as discussed in the paper, the loop could be run in parallel)


blackjax/optimizers/lbfgs.py
```diff
@@ -0,0 +1,289 @@
...
+        next_tup = lax.cond(not_converged, lbfgs_one_step, non_op, carry, it)
+        return next_tup, next_tup[0][-1]
+
+    solver = jaxopt.LBFGS(fun=fun, maxiter=maxiter, maxls=maxls, history_size=maxcor)
+    value0, grad0 = jax.value_and_grad(fun)(x0)
+    state = solver.init_state(x0)
+
+    value0, grad0 = jax.value_and_grad(fun)(x0)
+    # LBFGS update overwirte value internally, here is to set the value for checking condition
+    state = state._replace(value=value0, stepsize=state.stepsize / 1.5)
```
junpenglao, line 134: @miclegr any reason we reduce the initial step size here?

junpenglao:
@miclegr FYI

blackjax/optimizers/lbfgs.py
miclegr: No reason, I've left it there inadvertently while troubleshooting convergence issues

junpenglao: Thanks! Resolving...

miclegr:
That's much cleaner and clearer, great job!
I've also noticed that you used the latest revision of the paper, while I was using version 3 that was missing the `lbfgs_recover_alpha` step, good that you fixed that :)

rlouf (approved):
Looks good to me! Do you want to implement the API change we discussed for variational inference in this PR since it is a minimal change? AFAIR it was something along the lines of:

```python
import blackjax

pathfinder = blackjax.pathfinder(logprob_fn)
approximation = pathfinder.approximate(init_positions, max_num_steps)
samples = approximation.samples(num_samples)
```

We could even turn `approximation` into a generator.

junpenglao:
Let's hold off the refactoring of VI until we implement meanfield ADVI - right now pathfinder definitely doesnt feel VI to me

PR #211, opened 2022-05-24
Add the SGHMC algorithm

rlouf:
As described in http://proceedings.mlr.press/v32/cheni14.pdf

- [ ] Add the integrator for the corresponding diffusion
- [ ] Add function that builds the trajectory
- [ ] Add kernel
- [ ] Add  to tests
- [ ] Add to documentation
- [ ] Add to MNIST example
- [ ] Add CNN with `flax` (from SGMCMCJax)


rlouf:
@junpenglao ping

blackjax/kernels.py
```diff
@@ -475,6 +476,83 @@ def step_fn(rng_key: PRNGKey, state, data_batch: PyTree):
...
+    def __new__(  # type: ignore[misc]
+        cls,
+        grad_estimator_fn: Callable,
+        schedule_fn: Callable,
+        num_integration_steps: int,
+    ) -> SamplingAlgorithm:
+
+        step = cls.kernel(grad_estimator_fn)
+
+        def init_fn(position: PyTree, data_batch: PyTree):
```
junpenglao, line 546: Not sure about the typing here, as `data_batch` is a iterator/generator?

junpenglao: Same below

junpenglao (approved):
Looks good. A high level question:
Using a generator (as in the example notebook) only works if the training/sampling loop is in Python right? Can we put that into jax.lax.scan or jit it?

blackjax/kernels.py
rlouf: `data_batch` is a PyTree. I wrote an example where these batches are produced by a generator, but it happens outside the kernel and is not necessary.

rlouf:
That's a good question ! I don't think so as generally jax doesn't work well with statefull code. However it's very easy to modify that code and turn it into a function.

PR #216, opened 2022-06-07
Make MALA work with any PyTree

### LINKED ISSUE #214: MALA doesn't work with dictionary inputs
I apologize if I'm misunderstanding how the interface is supposed to work, but I tried the quickstart example using `blackjax.mala` instead of NUTS and I get the following error (blackjax version is `0.7.0`).

```
File .../blackjax/mcmc/mala.py:65, in kernel.<locals>.transition_probability(state, new_state, step_size)
     63 def transition_probability(state, new_state, step_size):
     64     """Transition probability to go from `state` to `new_state`"""
---> 65     theta = new_state.position - state.position - step_size * state.logprob_grad
     66     return -0.25 * (1.0 / step_size) * jnp.dot(theta, theta)

TypeError: unsupported operand type(s) for -: 'dict' and 'dict'
```

The code to reproduce this is pasted below (it is essentially just the quickstart example in the README). Note that if I set `initial_position` to an array `jax.numpy.array([1., 2.])` and modify the log probability function appropriately it seems to work. It would be nice if I could sample using MALA from a function that takes dict inputs.

```python
import jax
import jax.numpy as jnp
import numpy as np

import blackjax

observed = np.random.normal(10, 20, size=1_000)
def logprob_fn(x):
  logpdf = stats.norm.logpdf(observed, x["loc"], x["scale"])
  return jnp.sum(logpdf)

# Build the kernel
step_size = 1e-3
mala = blackjax.mala(logprob_fn, step_size)

# Initialize the state
initial_position = {"loc": 1., "scale": 2.}
state = mala.init(initial_position)

# Iterate
for _ in range(100):
    _, rng_key = jax.random.split(rng_key)
    state, _ = mala.step(rng_key, state)
```

rlouf:
Closes #214

### DIFF
--- a/blackjax/mcmc/mala.py
+++ b/blackjax/mcmc/mala.py
@@ -3,6 +3,7 @@ from typing import Callable, NamedTuple, Tuple
 
 import jax
 import jax.numpy as jnp
+from jax.flatten_util import ravel_pytree
 
 from blackjax.mcmc.diffusion import overdamped_langevin
 from blackjax.types import PRNGKey, PyTree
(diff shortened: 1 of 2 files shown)


junpenglao:
Add a small test?

junpenglao (approved):

PR #228, opened 2022-06-24
Add the marginal gradient based sampler for latent Gaussian models

AdrienCorenflos:
I would highly recommend reading the supplementary material of this to understand why the init and the kernel need to be created together https://rss.onlinelibrary.wiley.com/doi/full/10.1111/rssb.12269.

In fact, I tend to think some additional improvement should be done to the current code to allow not recomputing anything if delta does not change, but this may come as a second step.


rlouf:
Cool! Can you rebase your branch on main please?

tests/test_latent_gaussian.py
```diff
@@ -0,0 +1,65 @@
+import jax
+import jax.scipy.stats as stats
+import numpy as np
+import numpy.testing as np_testing
+import pytest
```
junpenglao, line 5: Please update the test using Chex

AdrienCorenflos:
I guess this should kind of do beside cosmetic changes?

junpenglao (approved):

tests/test_latent_gaussian.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+    def test_gaussian(self, seed, use_inverse, mean):
+        n_samples = 500_000
+
+        KEY = jax.random.PRNGKey(42)
+        D = 5
+        np.random.seed(seed)
+        C = np.random.randn(D, D**2)
+        C = C @ C.T
+        prior_mean = np.random.randn(D) if mean else None
```
junpenglao, line 24: Could you change this bit to use JAX random as well? Then everything is control by a single seed

AdrienCorenflos:
@rlouf just saw you assigned this to me. Do you need me to do anything? I'm leaving for a week tomorrow.

junpenglao:
I think this is fine with a squash and merge - @rlouf ?

rlouf:
I like to understand the code I merge in the codebase, will take the time to do that today

blackjax/__init__.py
```diff
@@ -33,4 +34,5 @@
     "ess",  # diagnostics
     "rhat",
     "pathfinder_adaptation",
+    "latent_gaussian",
```
rlouf, line 37: `latent_gaussian` refers to a class of models rather than a specific sampler. We should find a better name imo.

blackjax/kernels.py
```diff
@@ -394,6 +392,69 @@ def step_fn(rng_key: PRNGKey, state):
         return SamplingAlgorithm(init_fn, step_fn)
 
 
+class latent_gaussian:
+    """Implements the marginal sampler for latent Gaussian model of [1].
+    It uses a first order approximation to the log_likelihood of a model with Gaussian prior.
```
rlouf, line 397: A line break would be nice here.

blackjax/kernels.py
```diff
@@ -394,6 +392,69 @@ def step_fn(rng_key: PRNGKey, state):
         return SamplingAlgorithm(init_fn, step_fn)
 
 
+class latent_gaussian:
```
rlouf, line 395: `marginal_latent_gaussian` ?

rlouf (changes requested):
Code looks great! My remarks are mostly about naming / repo organization. After reading the notebook I think you should add the `calibration` function to adaptation algorithms. Which will probably give us a design headache that we'll need to take care of before v1 @junpenglao

blackjax/kernels.py
junpenglao: `marginal_latent_gaussian_sampler`? Since `marginal_latent_gaussian` could be interpreted as INLA like algorithm

blackjax/__init__.py
AdrienCorenflos: The paper calls it mGrad for "marginal gradient sampler for latent Gaussian models". I was trying a few things but it's a bit of a mouthful...

blackjax/kernels.py
```diff
@@ -631,7 +692,6 @@ def __new__(  # type: ignore[misc]
         logprob_fn: Callable,
         sigma: Array,
     ) -> SamplingAlgorithm:
-
```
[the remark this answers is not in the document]
AdrienCorenflos: Creates the kernel for this sampler.

AdrienCorenflos:
Regarding the adaptation, I would disagree, this is very ad-hoc and probably not the best way to do it, so much better to leave it as being simply illustrative IMO

blackjax/kernels.py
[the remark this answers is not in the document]
rlouf: Oh I thought the change was in `sgld.py`. My bad.

blackjax/kernels.py
rlouf: 😩

junpenglao: How about `MarginalGradSampler` or `MarginalGradientSampler`?

AdrienCorenflos: which essentially can be understood as "Marginal gradient sampler for latent Gaussian models"

AdrienCorenflos: Naming is always the worst :grimacing:

AdrienCorenflos: > marginal_latent_gaussian

INLA would be called inla though (and would probably live in VI, not MCMC)...

rlouf:
I refactored the code slightly to get match the conventions of the library a little. However the implementation is currently limited to array inputs and won't work with arbitrary PyTrees. Plus we need to fix the commit mess before merging; commits related to this PR are interleaved with commits from `main` that were merged on it.

rlouf:
Fixing the commit history will take me too long; I honestly think it would take less time to start on a new branch that's up-to-date with `main` and add the changes there. Also, I've noticed the algorithm only works with flat arrays, and not arbitrary PyTrees like the rest of the library. Is there a specific reason?

junpenglao:
@AdrienCorenflos Could you close this PR and start a fresh one from Main? Agree with @rlouf that this is a bit difficult to review.

AdrienCorenflos:
Closing in favour of https://github.com/blackjax-devs/blackjax/pull/247

Issue #220, opened 2022-06-14
MEADS


albcab:
Implementation of [MEADS](https://proceedings.mlr.press/v151/hoffman22a.html) (algorithm 3).

To do list of the components of the algorithm:

- [x] Non-reversible slice sampling for Metropolis accept/reject decisions ([Neal](https://arxiv.org/abs/2001.11950))
- [x] HMC with persistent momentum (Generalized HMC)
- [x] Tuning heuristics:

- Diagonal preconditioning
- Step size
- Damping factor 
- Slice drift
- Estimating largest eigenvalues

- [x] Parallelizable Ensemble-chain adaptation
- [x] Maximum-Eigenvalue Adaptation of Damping and Step-size

Issue #171, opened 2022-02-08
Expose the adaptation as kernel


rlouf:
The window adaptation is currently only available via a function that implements and jit-compiles the loop. While this is convenient and should be kept, we should also expose the window adaptation as a kernel (and also the slow and fast updates for consitency).

Currently we have the following runtime:

```python
import blackjax

adapt = blackjax.window_adaptation(blackjax.nuts, logprob_fn, num_warmup_steps)
state, kernel, info = adapt.run(rng_key, position)
```

While we wish to keep this design, we would like to allow separate chain and warmup state updates:

``` python
state = blackjax.nuts.init(position, logprob_fn)
warmup_state = blackjax.window_adaptation.init(num_warmup_steps)

new_state, info = blackjax.nuts.step(
    rng_key,
    state,
    logprob_fn,
    warmup_state.step_size,
    warmup_state.inverse_mass_matrix
)
new_warmup_state, warmup_info = blackjax.window_adaptation.update(
    warmup_state, new_state, info
)
```

PR #229, opened 2022-06-24
Add the Generalized HMC sampler + ECA adaptation (MEADS) 

albcab:
This is a working draft of MEADS (#220). MEADS kernel uses HMC with persistent momentum and MH using nonreversive slice sampling (GHMC) and parallel ensemble chain adaptation. GHMC uses the same `flip_momentum` function from HMC, maybe that function could be included in the public api of `blackjax.mcmc.hmc` to be used as a utility function? 

I still need to add docs and a better example of its use. The algorithm is most useful when running on multicore CPUs or GPUs.


blackjax/adaptation/parallel_eca.py
```diff
@@ -0,0 +1,64 @@
+"""Public API for (Parallelizable) Ensamble Chain Adaptations"""
```
rlouf, line 1:
```suggestion
"""Public API for (Parallelizable) Ensemble Chain Adaptations"""
```

blackjax/mcmc/ghmc.py
```diff
@@ -0,0 +1,209 @@
...
+    momentum = unravel_fn(jax.random.normal(key_mometum, p.shape))
+    slice = jax.random.uniform(key_slice, minval=-1.0, maxval=1.0)
+
+    return GHMCState(position, momentum, potential_energy, potential_energy_grad, slice)
+
+
+def kernel(
+    noise_gn: Callable = lambda _: 0.0,
+    divergence_threshold: float = 1000,
+):
```
rlouf, line 61: Isn't that from a paper by Radford Neal? If so this paper should be cited in the docstring.

blackjax/adaptation/parallel_eca.py
```diff
@@ -0,0 +1,64 @@
...
+    #         return states, info
+
+    #     states, infos = jax.lax.scan(step_fn, states, rng_keys)
+    #     return states, infos
+
+    def update(
+        rng_key: PRNGKey,
+        state: ParallelECAState,
+    ) -> Tuple[NamedTuple, Iterable, NamedTuple]:
+        parameters = jax.pmap(
```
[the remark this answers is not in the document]
albcab: Since the sampling is done on parallel, I figured the parameter estimation for each batch of chains should also be done on parallel.

blackjax/mcmc/ghmc.py
```diff
@@ -0,0 +1,209 @@
...
+            proposal.potential_energy_grad,
+            slice,
+        )
+
+        return state, info
+
+    return one_step
+
+
+def persistent_ghmc_proposal(
```
rlouf, line 114: Isn't that essentially the same as HMC's? We should re-use the code as much as possible: it simplified the codebase and make obvious the connections between the algorithms when there are some. In this case GHMC is "just" HMC with a different momentum sampling and acceptance schemes. We should make this obvious looking at the code.

blackjax/mcmc/ghmc.py
```diff
@@ -0,0 +1,209 @@
...
+            step_size,
+            divergence_threshold,
+            logprob_grad_fn,
+        )
+
+        key_momentum, key_noise = jax.random.split(rng_key)
+        position, momentum, potential_energy, potential_energy_grad, slice = state
+        momentum, unravel_fn = ravel_pytree(momentum)
+        # New momentum is persistent
+        momentum = unravel_fn(
```
rlouf, line 90: We can re-use the code in `metrics.py` for everything momentum related, no?

blackjax/mcmc/ghmc.py
```diff
@@ -0,0 +1,209 @@
...
+    def flip_momentum(state):
+        flipped_momentum = jax.tree_util.tree_map(lambda m: -1.0 * m, state.momentum)
+        return integrators.IntegratorState(
+            state.position,
+            flipped_momentum,
+            state.potential_energy,
+            state.potential_energy_grad,
+        )
+
+    def generate(
```
rlouf, line 182: Again, can't we reuse the code in `hmc.py` here?

rlouf:
Many great additions in one PR! However I am not a big fan of the warmup being ran inside the kernel; what about users who would like to run Generalized HMC with their parameters? There's a place for this kind of design, but imo `blackjax` is not where this should happen (PyMC would be, cf https://github.com/pymc-devs/pymc/issues/5930). 
`MEADS` is a warmup scheme for GHMC and should be exposed just as that, an adaptation scheme. GHMC should be exposed as a kernel. What do you think?

I'm asking for @junpenglao's opinion as well.

junpenglao:
+1 to @rlouf's point.
For example, in the case of ChEES we will add a HMC kernel that randomized the number of leapforg (Uniform(min, max) that min and max are result from ChEES tuning.

blackjax/mcmc/ghmc.py
albcab: The code now uses `metrics.py` (I've also commented a version without `metrics.py`) but it seems unnecessary. Maybe I could write a version of `gaussian_euclidean` with persistent momentum?

albcab:
> `MEADS` is a warmup scheme for GHMC and should be exposed just as that, an adaptation scheme (just like ChEEs is an adaptation scheme for HMC). GHMC should be exposed as a kernel. What do you think?

Agree. I've pushed a version with `meads` and `ghmc` as separate kernels, the former is an adaptation scheme with `run` api and the latter a sampling scheme with `init` and `step` apis.

I've also recycled code from `hmc.py` with small modifications to the original code to make it work. I've included Neal's nonreversible slice sampling style MH in `proposal.py`. I need a version of `velocity_verlet` that uses a different `step_size` for each dimension so I've included it as a separate function, I could possibly generalize the original `velocity_verlet` function or make a include a new general version in `integrators.py`. I could also include a persistent momentum version of `gaussian_euclidean` in `metrics.py` or shift the original version in `ghmc.py` as I do now. wdyt?

albcab:
> Very strange to me that they chose to do that instead of using a mass matrix but we'll have to deal with it I guess.

I think its because of the persistent momentum. Assuming initial momentum `m` has std normal distribution then new momentum `m' = sqrt(1-alpha) * m + sqrt(alpha) * e` for `e` std normal, also has std normal distribution (see calculation of MH acceptance rate). Maybe we could scale the momentum after its persistent update before numerical integration, but we would need to adjust the MH acceptance rate calculation to the scaled density. 

This could be a nice option since it would avoid generalizing the numerical integrators. Just have to make sure it works theoretically.

rlouf:
I think that once you're using `tree_multimap` that's not a huge deal. Let's see when the warmup is fully implemented!

rlouf:
Oh yes sorry I did not expect to find the warmup code in `kernels.py`! I'll need to take another look.

rlouf:
Would you mind organizing the warmup in `adaptation/meads.py` so that the following:

```python
    def max_eigen(matrix: PyTree):
        X = jnp.vstack(
            [leaf.T for leaf in jax.tree_leaves(matrix)]
        ).T  # will only work if all variables are at most vectors (not 2+ dimensional tensors)
        n, _ = X.shape
        S = X @ X.T
        diag_S = jnp.diag(S)
        lamda = jnp.sum(diag_S) / n
        lamda_sq = (jnp.sum(S**2) - jnp.sum(diag_S**2)) / (n * (n - 1))
        return lamda_sq / lamda

    def parameter_gn(batch_state, current_iter):
        batch_position = batch_state.position
        mean_position = jax.tree_map(lambda p: p.mean(axis=0), batch_position)
        sd_position = jax.tree_map(lambda p: p.std(axis=0), batch_position)
        batch_norm = jax.tree_map(
            lambda p, mu, sd: (p - mu) / sd,
            batch_position,
            mean_position,
            sd_position,
        )
        batch_grad = jax.pmap(logprob_grad_fn)(batch_position)
        batch_grad_scaled = jax.tree_map(
            lambda grad, sd: grad * sd, batch_grad, sd_position
        )
        epsilon = jnp.minimum(0.5 / jnp.sqrt(max_eigen(batch_grad_scaled)), 1.0)
        gamma = jnp.maximum(
            1.0 / jnp.sqrt(max_eigen(batch_norm)),
            1.0 / ((current_iter + 1) * epsilon),
        )
        alpha = 1.0 - jnp.exp(-2.0 * epsilon * gamma)
        delta = alpha / 2
        step_size = jax.tree_map(lambda sd: epsilon * sd, sd_position)
        return step_size, alpha, delta
```

is included in this file instead of `kernels.py` (and merge with what's in `eca.py`, at this stage there is no reason to separate the two)? `kernels.py` is supposed to be mostly interface code and plumbing.

blackjax/kernels.py
```diff
@@ -765,6 +765,148 @@ def step_fn(rng_key: PRNGKey, state):
...
+        def step_fn(rng_key: PRNGKey, state):
+            return step(
+                rng_key, state, logprob_fn, step_size, alpha, delta, logprob_grad_fn
+            )
+
+        return SamplingAlgorithm(init_fn, step_fn)  # type: ignore[arg-type]
+
+
+def meads(
+    algorithm: ghmc,
```
rlouf, line 799: Do you mean `algorihm=gmc`?

blackjax/kernels.py
```diff
@@ -765,6 +765,148 @@ def step_fn(rng_key: PRNGKey, state):
...
+    logprob_fn: Callable,
+    num_batch: int,
+    batch_size: int,
+    num_steps: int = 1000,
+    *,
+    divergence_threshold: int = 1000,
+    logprob_grad_fn: Optional[Callable] = None,
+) -> AdaptationAlgorithm:
```
rlouf, line 807: A docstring would be great. In particular I have no idea what `num_batch` and `batch_size` mean (if I haven't read the paper).

rlouf: Do the authors give reasonable default values for these parameters?

blackjax/kernels.py
```diff
@@ -765,6 +765,148 @@ def step_fn(rng_key: PRNGKey, state):
...
+                delta,
+                logprob_grad_fn=logprob_grad_fn,
+            )
+
+        return kernel_fn
+
+    if not logprob_grad_fn:
+        logprob_grad_fn = jax.grad(logprob_fn)
+
+    def max_eigen(matrix: PyTree):
```
rlouf, line 828: Docstring + maybe a more explicit name?

blackjax/kernels.py
```diff
@@ -765,6 +765,148 @@ def step_fn(rng_key: PRNGKey, state):
...
+            [leaf.T for leaf in jax.tree_leaves(matrix)]
+        ).T  # will only work if all variables are at most vectors (not 2+ dimensional tensors)
+        n, _ = X.shape
+        S = X @ X.T
+        diag_S = jnp.diag(S)
+        lamda = jnp.sum(diag_S) / n
+        lamda_sq = (jnp.sum(S**2) - jnp.sum(diag_S**2)) / (n * (n - 1))
+        return lamda_sq / lamda
+
+    def parameter_gn(batch_state, current_iter):
```
rlouf, line 839: Idem as above.

blackjax/adaptation/parallel_eca.py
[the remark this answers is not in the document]
rlouf: Correct me if I'm wrong @junpenglao but `pmap` is for execution on multiple devices (CPUs, TPUs), and `vmap` for vectorization on the same machine. There might be cases where one would want to use `pmap` here but I think in most cases we'd want to use `vmap`?

rlouf (changes requested):
There is still one big change to make, move over all the `meads` adaptation logic to `eca.py` (and rename it `meads.py`) so only the plumbing is left in `kernels.py`. Besides that, and the refactoring I suggested above we need to add docstrings, doc, makes the checks pass, and decide what we do about  `pmap` vs `vmap`.

albcab:
> There is still one big change to make, move over all the `meads` adaptation logic to `eca.py` (and rename it `meads.py`) so only the plumbing is left in `kernels.py`

I've created `meads.py` as discussed. I've also added a new file called `chain_adaptation.py` which includes functions that perform parallel ensemble chain adaptation and cross chain adaptation. Since implementing [Ch-ESS HMC](http://proceedings.mlr.press/v130/hoffman21a/hoffman21a.pdf) on blackjax's API for my research, I've noticed that MEADS Algorithm 3 is implemented as an [adaptive MCMC](http://probability.ca/jeff/ftpdir/adaptex.pdf) algorithm, in the sense that it is meant to generate correct samples as it adapts. For an algorithm like this to be ergodic it needs to follow certain criteria hence why we need parallel ensemble chain adaptation. On the other hand, Ch-ESS Algorithm 1 is not an adaptive algorithm, it is meant to adapt through a warm-up phase (generating incorrect samples), then the algorithm's parameters are fixed and correct samples can be generated. MEADS can also be run like this and not as adaptive MCMC, hence I've given the user the choice (bool parameter `eca`) of running adaptive MCMC (where samples from `warmup_states` can be used) or as adaptation returning a `ghmc` kernel with fixed parameters (same as `window_adaptation`). We have the choice to implement these options as two separate kernels or as one kernel with a bool parameter. In any case, I'll need cross chain adaptation to implement Ch-ESS, so I've included this in a new file (I also think it'll be useful for other adaptive algorithms).

> decide what we do about `pmap` vs `vmap`

I've included a `batch_fn` option for the user and fixed other minor things.

Let me know what you think about the current implementation and if there is something that can be improved. I'm on vacation the coming two weeks but after that I'll add docstrings and tests, finish this PR, open a PR for Ch-ESS and add an example of MEADS and Ch-ESS, probably reproducing the experiments of the MEADS paper or something similar.

albcab:
I've left only the version of MEADS that works like `window_adaptation` for GHMC. It does change the current API for warm-ups and from various experiments I've found no improvement with the adaptive version. 

Also there are some examples using real data that I though could be interesting to have in the documentation.

blackjax/adaptation/chain_adaptation.py
```diff
@@ -0,0 +1,78 @@
+"""Public API for General Cross-Chain Adaptations"""
```
rlouf, line 1: Is that going to be re-used for ChEes?

blackjax/adaptation/chain_adaptation.py
```diff
@@ -0,0 +1,78 @@
...
+            raise ValueError(
+                "Cross-chain adaptation got inconsistent sizes for array axes on *State. Every array's shape must be of the form (num_chain, ...)"
+            )
+        return ChainState(initial_states, 0)
+
+    def update(
+        rng_key: PRNGKey, state: ChainState, *param
+    ) -> Tuple[ChainState, PyTree, NamedTuple]:
+        parameters = parameter_gn(state.states, state.current_iter, *param)
+        kernel = batch_fn(kernel_factory(*parameters))
```
rlouf, line 73: Note for future self: We should ask the user to pass a `vmap`ped or `pmap`ped directly instead of asking for a kernel factory _and_ a batching function.

blackjax/adaptation/chain_adaptation.py
```diff
@@ -0,0 +1,78 @@
...
+    Returns
+    -------
+    init
+        Function that initializes the warmup.
+    update
+        Function that moves the warmup one step.
+    """
+
+    def init(initial_states: NamedTuple) -> ChainState:
+        check_leaves_shape = jax.tree_util.tree_leaves(
```
rlouf, line 60: I think we should trust that the user knows what she's doing and assume that the number of chains is equal to the length of the state's first dimension.

blackjax/adaptation/chain_adaptation.py
```diff
@@ -0,0 +1,78 @@
...
+    Parameters
+    ----------
+    kernel_factory
+        Function that takes as input the parameters that need to be learned and
+        outputs a kernel that generates new samples.
+    parameter_gn
+        Function that takes as input the last state of all chains stacked on the
+        first dimension, the current iteration of the warm-up phase, and other
+        optional inputs and outputs updated parameters.
+    num_chain
```
rlouf, line 46: `adaptation_fn`?

blackjax/adaptation/chain_adaptation.py
rlouf: We'd be down to 2 parameters for `cross_chain`.

blackjax/adaptation/meads.py
```diff
@@ -0,0 +1,125 @@
...
+    by the grouped samples of all chains with shape (num_chain, parameter dimension).
+    Specifically, it does Algorithm 3 of [1]_ with cross-chain adaptation instead of
+    parallel ensample chain adaptation.
+
+    Parameters
+    ----------
+    kernel_factory
+        Function that takes as input the step size, alpha and delta parameters
+        and outputs a Generalized HMC kernel that generates new samples.
+    num_chain
```
rlouf, line 34: `log_prob_grad_fn` is missing in the list of parameters.

blackjax/adaptation/chain_adaptation.py
```diff
@@ -0,0 +1,78 @@
...
+            )
+        return ChainState(initial_states, 0)
+
+    def update(
+        rng_key: PRNGKey, state: ChainState, *param
+    ) -> Tuple[ChainState, PyTree, NamedTuple]:
+        parameters = parameter_gn(state.states, state.current_iter, *param)
+        kernel = batch_fn(kernel_factory(*parameters))
+        keys = jax.random.split(rng_key, num_chain)
+        new_states, infos = kernel(keys, state.states)
```
rlouf, line 75: I know this is currently happening in `window_apdaptation`, but I've done things differently in [AeHMC](https://github.com/aesara-devs/aehmc) where sampling and adaptation happen completely separately in the sense that adaptation is not moving the state. We'll leave it like this for now but will need to adress this deisgn issue in #171

blackjax/mcmc/ghmc.py
```diff
@@ -0,0 +1,254 @@
...
+            proposal.potential_energy_grad,
+            info.acceptance_probability,
+        )
+
+        return state, info
+
+    return one_step
+
+
+def velocity_verlet(
```
rlouf, line 170: That you specify a step size per component instead of a mass matrix is unfortunate, but this will do for now.

blackjax/mcmc/ghmc.py
```diff
@@ -0,0 +1,254 @@
...
+            position,
+            momentum,
+            potential_energy,
+            potential_energy_grad,
+        )
+
+    return one_step
+
+
+def update_momentum(rng_key, position, momentum, alpha):
```
rlouf, line 233: `rng_key, state, alpha`?

blackjax/mcmc/proposal.py
```diff
@@ -98,6 +98,35 @@ def static_binomial_sampling(rng_key, proposal, new_proposal):
     )
 
 
+# --------------------------------------------------------------------
+#                   NON-REVERSIVLE SLICE SAMPLING
+# --------------------------------------------------------------------
+
+
+def nonreversible_slice_sampling(slice, proposal, new_proposal):
```
rlouf, line 106: :+1:

rlouf:
Almost there! Two things related to the notebooks:
1- We can probably skip `MEADS.ipynb`, it is not adding anything that is not in the two others;
2- The rhat diagnostic takes values greater than 1 in the sparse regression model, which means the chains are not mixing well.

blackjax/adaptation/chain_adaptation.py
albcab: Yes it will.

albcab:
Great! I've made made the corrections you've recommended and skipped `MEADS.ipynb`. I've noticed all notebooks are .md now, should I push them as .md files directly? Also, I've increased the R-hat on the sparse regression model by increasing the amount of working iterations (given the high dimension of the target space and the hierarchies, this models tends to be a hard one to sample from). Comparisons with ChESS and other parallel MCMC methods should be interesting on these examples.

rlouf:
Great! Yes would you mind converting the notebooks to `.md` following the instructions in the README? Good on my end otherwise.

@junpenglao do you want to take a look?

rlouf:
@junpenglao ping

blackjax/adaptation/meads.py
```diff
@@ -0,0 +1,125 @@
...
+        warmup state.
+
+    References
+    ----------
+    .. [1]: Hoffman, M. D., & Sountsov, P. (2022). Tuning-Free Generalized
+            Hamiltonian Monte Carlo. In International Conference on Artificial
+            Intelligence and Statistics (pp. 7799-7813). PML
+    """
+
+    def maximum_eigenvalue(matrix: PyTree):
```
junpenglao, line 55: This function should move to outside of the function scope.

junpenglao: Also, add docstring to explain that `matrix` each leaf have a single batch shape at the first axis.

blackjax/adaptation/meads.py
```diff
@@ -0,0 +1,125 @@
...
+        directly computing an unbiased estimate of the largest eigenvalue because of
+        the latter's large variance.
+        """
+
+        X = jnp.vstack(
+            [
+                leaf.reshape(leaf.shape[0], -1).T
+                for leaf in jax.tree_util.tree_leaves(matrix)
+            ]
+        ).T
```
junpenglao, line 72:
```suggestion
        X = jnp.hstack(
            [
                leaf.reshape(leaf.shape[0], -1)
                for leaf in jax.tree_util.tree_leaves(matrix)
            ]
        )
```

Or add a kwarg of unravel function, where user can pass in function to unravel (which could be `vmapping` a unravel func like in
https://github.com/blackjax-devs/blackjax/blob/main/blackjax/vi/pathfinder.py#L176-L184)

junpenglao:
Per discussion with @rlouf offline, we will merge first and make change of the API with other refactoring effort.

junpenglao (approved):

rlouf:
Great job! Thank you for another great contribution and your patience with the process.

Issue #227, opened 2022-06-20
Could LBFGS work with pytrees?


ahwillia:
Similar to this previous issue (https://github.com/blackjax-devs/blackjax/issues/214), would it be possible to modify `minimize_lbfgs` to also work with the initial guess `x0` being something more flexible like a tuple of arrays? That would be really handy so that I could use the same `logprob` function to draw samples and to find the (local) max of the posterior.

junpenglao:
We are doing a bit more matrix operation following the pathfinder paper so we would mostly likely keep working with flatten array internally, but maybe we can have a public API that does the unflatten? WDYT @rlouf ?
Proposal is `minimize_lbfgs` -> `_minimize_lbfgs` and have `minimize_lbfgs` being the public API that accepts pytree input.

rlouf:
Another option, if we don't want to have multiple functions, is to build within `lbgfs` a function that works with flattened array using `ravel_pytrees`, ravel the input and and unravel the output.

junpenglao:
A decorator is actually a great idea. One problem however is that the output structure is not necessary a pytree with the same structure (some you dont need to unravel, some you need to `vmap(unravel_fn)(...)`)

rlouf:
We can just unravel inside the function for this one and generalize later if needed. @ahwillia do you feel like submitting a PR for this?

ahwillia:
Happy to give it a shot. I am somewhat new to jax just learned that `jax.flatten_util.ravel_pytree` exists. I'm not sure I understand the cases where `vmap(unravel_fn)(...)` is needed though.

Is the following sketch what you two are thinking?

```python
def minimize_lbfgs(func, x0, **kwargs):
    if isinstance(x0, jax.numpy.DeviceArray):
        return _minimize_lbfgs(func, x0, **kwargs)
    else:
        x0_raveled, unravel_fn = ravel_pytree(init_params)
        return _minimize_lbfgs(lambda x: func(unravel_fn(x)), x0_raveled, **kwargs)
```

Where `minimize_lbfgs` -> `_minimize_lfbgs`.

rlouf:
I think you can safely ravel an array, in which case no need to bother with the `if...else`. Don't forget to unravel the output as well!

ahwillia:
Nice, that's very simple then! Unraveling the output seems a little tricky because everything is inside `LBFGSHistory` and the fields `x` and `g` are typed as `Array`. Should I change those types to `PyTree`? Is there a treemap

[...]

PR #232, opened 2022-06-27
Tweak LBFGS to allow PyTree inputs

ahwillia:
This addresses https://github.com/blackjax-devs/blackjax/issues/227.

It passes tests locally on my machine, but should probably add an explicit test for a more complex pytree input. Let me know your thoughts.

One thing I'm not sure about is whether you want to count on `LbfgsState` being a stable feature of `jaxopt`. It might make sense to only unravel the fields in `LBFGSHistory`.

### DIFF
--- a/blackjax/optimizers/lbfgs.py
+++ b/blackjax/optimizers/lbfgs.py
@@ -6,8 +6,8 @@ import jax.random
 import jaxopt
 from jax import lax
 from jax.flatten_util import ravel_pytree
-from jaxopt.base import OptStep
 from jaxopt._src.lbfgs import LbfgsState
+from jaxopt.base import OptStep
 
 from blackjax.types import Array, PyTree
 
@@ -79,7 +79,7 @@ def minimize_lbfgs(
     -------
     Optimization results and optimization path
     """
-    
+
     # Ravel pytree into flat array.
     x0_raveled, unravel_fn = ravel_pytree(x0)
     unravel_fn_mapped = jax.vmap(unravel_fn)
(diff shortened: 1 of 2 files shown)


blackjax/optimizers/lbfgs.py
```diff
@@ -77,6 +79,65 @@ def minimize_lbfgs(
...
+
+    # Unravel optimization path history.
+    history = LBFGSHistory(
+        x=unravel_fn_mapped(history_raveled.x),
+        f=history_raveled.f,
+        g=unravel_fn_mapped(history_raveled.g),
+        alpha=unravel_fn_mapped(history_raveled.alpha),
+        update_mask=unravel_fn_mapped(
+            history_raveled.update_mask.astype(x0_raveled.dtype)
+        ).astype(history_raveled.update_mask.dtype),
```
junpenglao, line 126:
```suggestion
        update_mask=jax.tree_map(
            lambda x: x.astype(history_raveled.update_mask.dtype), 
            unravel_fn_mapped(history_raveled.update_mask.astype(x0_raveled.dtype))
        ),
```

junpenglao (changes requested):

junpenglao:
You also need to change the import in Pathfinder (and other places where `minimize_lbfgs` is initially called) to 
```python
from blackjax.vi.pathfinder import _minimize_lbfgs
```

ahwillia:
Thanks for the review, I cleaned those items up.

junpenglao (changes requested):
`test_optimizers.py` also need update.

junpenglao:
Actually, we can leave the test as is (still works as unravel is a non-op there)

PR #247, opened 2022-07-05
Add Gradient based sampler for latent Gaussian models

AdrienCorenflos:
This is a clean up of the commit history of https://github.com/blackjax-devs/blackjax/pull/228, implementing the marginal sampler of https://rss.onlinelibrary.wiley.com/doi/full/10.1111/rssb.12269.

### DIFF
--- a/blackjax/__init__.py
+++ b/blackjax/__init__.py
@@ -5,6 +5,7 @@ from .kernels import (
     elliptical_slice,
     hmc,
     mala,
+    mgrad_gaussian,
     nuts,
     orbital_hmc,
     pathfinder_adaptation,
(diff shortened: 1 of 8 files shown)


blackjax/__init__.py
```diff
@@ -5,6 +5,7 @@
     elliptical_slice,
     hmc,
     mala,
+    marginal_latent_gaussian,
```
junpenglao, line 8: Should we just go with `mGrad` like in the paper?

AdrienCorenflos: It's not very standard...

AdrienCorenflos: That being said, if you want to change the name, please do. I'm not attached to it.

PR #256, opened 2022-07-20
Use jax.tree_map in place of jax.tree_multimap

jakevdp:
The latter is deprecated in jax and will be removed in https://github.com/google/jax/pull/11382

Note that in the last several JAX releases, `tree_multimap` is a direct alias of `tree_map`, so this substitution will not result in any change of behavior aside from silencing the deprecation warning (see https://github.com/google/jax/blob/5b576cb03e85f2e4d8d99f314cf5d87981ef5d42/jax/_src/tree_util.py#L198-L202).


junpenglao:
Thank you!

PR #271, opened 2022-09-03
Fix flaky test

### LINKED ISSUE #269: Some tests are flaky
As can be seen in [this run](https://github.com/blackjax-devs/blackjax/runs/8163451457?check_suite_focus=true) some tests in the test suite are flaky in the sens that they do not give the same result depending on the run. We should first understand where that randomness come from, and then assess the soundness of each test.

junpenglao:
- Flakyness in optimizer is due to maxcor too low
- Flakyness in linear regression test is due to incorrect logprob function (actually all algorithm in the same test returns bias result)

fix #269


rlouf:
LGTM, merging!

PR #273, opened 2022-09-05
Create an example for sampling multiple chains using vmap and pmap

### LINKED ISSUE #251: pmap seems to drastically improve performance in the example notebook
# Description

I am running `blackjax` in `WSL2` on a 32 core CPU.

I was playing with the example notebook (https://blackjax-devs.github.io/blackjax/examples/Introduction.html#), 
and noticed that the CPU utilization is actually quite low when running multiple chains. 

I have modified the code by first running 
```python
import numpyro as npr
npr.util.set_host_device_count(32)
```

Then I re-used the inference loop from the single-chain example, but instead of using `vmap` I used `pmap` to parallelize the execution:

```python

keys = jax.random.split(rng_key, num_chains)
inference_loop = jax.pmap(
    inference_loop, in_axes=(0, 0, None, None), static_broadcasted_argnums=(2, 3)
)

states = inference_loop(keys, initial_states, nuts.step, 1_000)
```

And this seems to cut the running from 2 **minutes** to 3 **seconds**
```
# vmap
Wall time: 2min 10s
# pmap
Wall time: 2.91 s
```

Am I doing something wrong here, or should the example actually be adjusted to use `pmap`?

# Reproducing

See full notebooks here:  
https://gist.github.com/elanmart/810f1964738b0ddd8f108b17b7969f82

# Setup

```
Python implementation: CPython
Python version       : 3.9.12
IPython version      : 8.4.0

jax     : 0.3.14
jaxlib  : 0.3.14
blackjax: 0.8.2

Compiler    : GCC 7.5.0
OS          : Linux
Release     : 5.10.102.1-microsoft-standard-WSL2
Machine     : x86_64
Processor   : x86_64
CPU cores   : 32
Architecture: 64bit
```

elanmart:
Resolves #251

I am not sure if this will work for you 
1. I've created a new notebook focusing only on sampling multiple chains, removing that section from `Introduction`. My motivation is that `pmap` requires you to set `XLA` flags **before** running any `jax` code, and it seemed confusing when I included this in the `Introduction.md`. 
2. I've included a bit of background on `vmap` and `pmap`, linking to the relevant `jax` resources, but now the example is quite wordy. Let me know if I should get rid of some of the paragraphs. 

Let me know what you think!


examples/MultipleChains.md
```diff
@@ -0,0 +1,276 @@
...
+
+In this example, we'll briefly demonstrate how you can run multiple MCMC chains using `jax` built-in constructs: `vmap` and `pmap`.
+We will use the NUTS example from the previous notebook, and compare the performance of the two approaches.
+
+## Vectorization vs parallelization
+
+`jax` provides two distinct transformations:
+- [vmap](https://jax.readthedocs.io/en/latest/jax.html?highlight=vmap#jax.vmap), used to automatically vectorize `jax` code
+- and [pmap](https://jax.readthedocs.io/en/latest/_autosummary/jax.pmap.html#jax.pmap),
+which enables paralleiziation across multiple devices, such as multiple GPUs (or, in our case, CPU cores).
```
rlouf, line 24:
```suggestion
which enables parallelization across multiple devices, such as multiple GPUs (or, in our case, CPU cores).
```

examples/MultipleChains.md
```diff
@@ -0,0 +1,276 @@
...
+    jupytext_version: 1.14.1
+kernelspec:
+  display_name: Python 3 (ipykernel)
+  language: python
+  name: python3
+---
+
+# Sampling Multiple Chains
+
+In this example, we'll briefly demonstrate how you can run multiple MCMC chains using `jax` built-in constructs: `vmap` and `pmap`.
```
rlouf, line 16:
```suggestion
In this example, we will briefly demonstrate how you can run multiple MCMC chains using `jax` built-in constructs: `vmap` and `pmap`.
```

examples/MultipleChains.md
```diff
@@ -0,0 +1,276 @@
...
+We will use the NUTS example from the previous notebook, and compare the performance of the two approaches.
```
rlouf, line 17:
```suggestion
We will use the NUTS example from the introduction notebook, and compare the performance of the two approaches.
```

examples/MultipleChains.md
```diff
@@ -0,0 +1,276 @@
...
+    parallel_fn = jax.pmap(fn)
+
+    parallel_fn(data)
+
+except Exception as e:
+    print(e)
+```
+
+This means that you'll only be able to run as many MCMC chains as you have CPU cores.
+See this [question](https://github.com/google/jax/discussions/4198) for a more discussion on the topic,
```
rlouf, line 82:
```suggestion
See this [question](https://github.com/google/jax/discussions/4198) for a more detailed discussion on the topic,
```

examples/MultipleChains.md
```diff
@@ -0,0 +1,276 @@
...
+    data = jnp.arange(1024)
+    parallel_fn = jax.pmap(fn)
+
+    parallel_fn(data)
+
+except Exception as e:
+    print(e)
+```
+
+This means that you'll only be able to run as many MCMC chains as you have CPU cores.
```
rlouf, line 81:
```suggestion
This means that you will only be able to run as many MCMC chains as you have CPU cores.
```

examples/MultipleChains.md
```diff
@@ -0,0 +1,276 @@
...
+ax.set_ylabel("loc")
+ax.legend(["Chain 1", "Chain 2"])
+```
+
+### Using `pmap`
+
+In case of `pmap`, we can simply choose to apply the transformation directly to the original `inference_loop` function.
+
+```{code-cell} ipython3
+inference_loop_multiple_chains = jax.pmap(inference_loop, in_axes=(0, None, 0, None), static_broadcasted_argnums=(1, 3))
```
rlouf, line 222: What happens if you apply `jax.vmap` to `inference_loop`?

rlouf:
Looks good to me beside a few nitpicks :) Thank you for contributing.

examples/MultipleChains.md
elanmart: Performance-wise it's exactly the same on my machine. 
IMO the code is then a bit cleaner (no need to write a new function).

Should I change the example and apply `vmap` to `inference_loop`? I didn't change the original example since I thought there might be some other reasons to do manually implement `inference_loop_multiple_chains`

rlouf:
Great work @elanmart ! I will include this in a series of `How To`s  in the documentation.

Issue #215, opened 2022-06-06
makes window_adaptation more vmap friendly


junpenglao:
Currently we cannot vmap `window_adaptation` as the one of the return is a `kernel`
https://github.com/blackjax-devs/blackjax/blob/f4221d041902b472095983cd83edffb76a395f51/blackjax/kernels.py#L584-L587

Proposal to add a kwarg `return_kernel=True` to `window_adaptation`:

```python
 step_size, inverse_mass_matrix = final(last_warmup_state) 
 if return_kernel:
     kernel = kernel_factory(step_size, inverse_mass_matrix) 
     return last_chain_state, kernel, warmup_chain 
 else:
     return last_chain_state, (step_size, inverse_mass_matrix), warmup_chain 
```

rlouf:
I've never been a big fan on this API to be honest, I expect it to return the parameters every time I use it. So maybe we should do just that.

rlouf:
Linked to #171. Should be addressed together.

PR #276, opened 2022-09-12
Refactor the adaptation kernels

rlouf:
The goal of this PR is 3-fold:

1. Make the adaptation more `vmap`-friendly (#215)
2. Decouple moving the chain / updating the parameter values (and follow what was done in [AeHMC](https://github.com/aesara-devs/aehmc), #171)
3. Refactor the internals wherever it makes it more clear what is happening.

This (mostly 1) introduce breaking changes.

## Decouple sampling and adaptation

Internally adaptation and sampling kernels are now completely decoupled, so one adaptation step is the succession of one sampling step followed by updating the parameters' values:

```python
# First update the state
# this does not happen inside of `window_adaptation` anymore
new_state, info = kernel(
    rng_key,
    state,
    logprob_fn,
    adaptation_state.step_size,
    adaptation_state.inverse_mass_matrix,
    logprob_grad_fn,
    **parameters,
)

# Then update the kernel parameters
new_adaptation_state = update(
    rng_key,
    adaptation_state,
    new_state.position,
    info.acceptance_proability,
)
```

## Make the adaptation more `vmap` friendly

Returning a kernel specialize with the values computed during adaptation was a fine idea, but that unfortunately prevents us from vectorizing the adaptation, so here's the updated version:

```python
adaptation = blackjax.window_adaptation(blackjax.nuts, logprob_fn, 1000)
state, params, _ = adaptation.run(rng_key, init_position)
kernel = blackjax.nuts(logprob_fn, *params).step
```

_Note:_ Returning a kernel was pretty neat and I don't like the current state of affairs. A compromise that doesn't involve kwargs would be to return a `State`, `Kernel`, `AdaptationState` and let people pick ˋAdaptationState.step_size` and `AdaptationState.inverse_mass_matrix` should they want to parallelise the algorithm.

## Status

- [x] Window adaptation
- [x] MEADS
- [x] Ensemble Chain Adaptation
- [x] Pathfinder adaptation
- [ ] Check API on examples


.pre-commit-config.yaml
```diff
@@ -33,13 +33,3 @@ repos:
   rev: v0.942
   hooks:
     - id: mypy
-- repo: https://github.com/nbQA-dev/nbQA
```
junpenglao, line 36: reverse (maybe your branch is not at HEAD?)

rlouf: It is at HEAD. We forgot to remove those in the PR that converted all notebooks to markdown.

junpenglao: Oh right we dont do notebook any more haha

rlouf:
@junpenglao would appreciate your opinion on `window_adaptation` before I move on to refactor `pathfinder_adaptation` as well

blackjax/kernels.py
```diff
@@ -672,41 +672,41 @@ def window_adaptation(
...
+        target_acceptance_rate,
     )
 
     def one_step(carry, xs):
         _, rng_key, adaptation_stage = xs
         state, adaptation_state = carry
-        state, adaptation_state, info = update(
-            rng_key, state, adaptation_state, adaptation_stage
+
+        new_state, info = kernel(
```
junpenglao, line 688: I am getting confuse here for a moment because we have too many things call kernel - could you add docstring or use `mcmc_one_step` (or `mutation`/`state_transition` etc) here?

junpenglao (commented):
Looks great, especially love it that you change the input signature to pass in only the necessary parameters, really makes clear what the functions are adapting.

High level question: are we going to do away the `kernel_factory` in other adaptation scheme as well? IIUC they suffer the same hard to vmap problem.

rlouf:
> High level question: are we going to do away the `kernel_factory` in other adaptation scheme as well? IIUC they suffer the same hard to vmap problem.

Yes! I'm perfecting `windows_adaptation` and will then refactor `pathfinder_adaptation` (and probably wait for MEADS to be merged). Probably should have started with Pathfinder, it's a much simpler warmup.

blackjax/adaptation/window_adaptation.py
```diff
@@ -12,20 +12,21 @@
...
+from blackjax.types import Array, PRNGKey, PyTree
 
 __all__ = ["base", "schedule"]
 
 
 class WindowAdaptationState(NamedTuple):
-    da_state: DualAveragingAdaptationState
-    mm_state: MassMatrixAdaptationState
+    ss_state: DualAveragingAdaptationState
+    imm_state: MassMatrixAdaptationState
```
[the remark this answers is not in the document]
rlouf: `stepsize_state`

blackjax/kernels.py
rlouf: Yes, kernel is not a very informative name.

blackjax/adaptation/window_adaptation.py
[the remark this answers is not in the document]
junpenglao: I see, make sense. Add comment?

junpenglao:
> Yes! I'm perfecting `windows_adaptation` and will then refactor `pathfinder_adaptation` (and probably wait for MEADS to be merged). Probably should have started with Pathfinder, it's a much simpler warmup.

SG - what about other places with `kernel_factory` signature (`*tempered_smc` in blackjax/kernel.py)

junpenglao:
> > Yes! I'm perfecting `windows_adaptation` and will then refactor `pathfinder_adaptation` (and probably wait for MEADS to be merged). Probably should have started with Pathfinder, it's a much simpler warmup.
> 
> SG - what about other places with `kernel_factory` signature (`*tempered_smc` in blackjax/kernel.py)

rlouf:
> SG - what about other places with `kernel_factory` signature (`*tempered_smc` in blackjax/kernel.py)?

Oh yes these are going to, they've been bothering me for a while now. 

The new warmup paradigm (sample then update parameters) should also help with adaptation within SMC.

rlouf:
The main pain point in terms of API seemed to be adaptation with multiple chains. My original incline was to not return the adapted mcmc kernel but the values of the parameters instead. But I am not a big fan of this solution either, re-initializing the kernel user-side is slightly cumbersome and needs familiarity with the `kernels` API that users otherwise don't need when they use adaptation.

I came up with a compromise. I created a `AdaptationResults` object:

```python
class AdaptationResults(NamedTuple):
    state: NamedTuple
    kernel: Callable
    parameters: Tuple
```

that all adaptation algorithms will return, along with additional information. This vaguely follows the `state, info` convention that is adopted throughout the library. Now there are two ways to run window adaptation with multiple chains.

### Adapt in parallel, then sample in parallel

In the following examples I am using the linear regression model found in `test_sampling.py`. This was not possible to do before:

```python
window_adaptation = blackjax.window_adaptation(
    blackjax.nuts,
    logposterior_fn,
    1000
)

adapt_key, sample_key = jax.random.split(rng_key)

def adapt(key, initial_position):
    adaptation_result, _ = window_adaptation.run(
        key,
        initial_position,
    )
    return adaptation_result.parameters, adaptation_result.state

keys = jax.random.split(adapt_key, 10)
initial_positions = {"scale": jnp.ones(10), "coefs": 2. * jnp.ones(10)}
params, states = jax.vmap(adapt)(keys, initial_positions)

def sample(rng_key, parameters, initial_state):
    kernel = blackjax.nuts(logposterior_fn, *parameters)
    chain = inference_loop(kernel.step, 1000, rng_key, initial_state)
    return chain

keys = jax.random.split(sample_key, 10)
res = jax.vmap(sample)(keys, params, states)
```

### Parallelize adaptation + sampling 

```python
window_adaptation = blackjax.window_adaptation(
    blackjax.nuts,
    logposterior_fn,
    1000
)

def adapt_and_sample_one_chain(rng_key, initial_position):
    
    adapt_key, sample_key = jax.random.split(rng_key)

    adaptation_result, _ = window_adaptation.run(
        adapt_key,
        initial_position,
    )

    chain = inference_loop(
        adaptation_result.kernel,
        1000,
        sample_key,
        adaptation_result.state
    )

    return chain

rng_keys = jax.random.PRNGKey(0)
keys = jax.random.split(rng_key, 10)
initial_positions = {"scale": jnp.ones(10), "coefs": 2. * jnp.ones(10)}

res = jax.vmap(adapt_and_sample_one_chain)(keys, initial_positions)
```

which is by far my favorite out of the two.

I am not sure why someone would use method number one, maybe sampling in an "interactive" way to debug and wanting to see the results of adaptation before moving on to sampling? But then couldn't `window_adaptation` return the `one_step` function as well as `run`, i.e. return `AdaptationAlgorithm(one_step, run)` ? I'd like your opinion on this.

junpenglao:
After thinking about it a bit more, here are some thoughts:

1. For high level usage (e.g., I dont care about what magic is happening, just return me good quality MCMC samples), the option 2`Parallelize adaptation + sampling` makes sense, but even the origin approach that only returns a kernel would do.
2. This leads to a different, lower level usage pattern: I want to parallel across the adaptation. It could be use case as you mentioned (sampling in an "interactive" way to debug and wanting to see the results of adaptation before moving on to sampling), but it could also for additional computation/adaptation across multiple chain (ChEES comes to mind). For that user can of course vmap across the adaptation one step, but to provide end to end `adaptation.run(...)` it does require returning the PyTree.
3. Not sure i understand the `AdaptationAlgorithm(one_step, run)` proposal, where are the parameters goes?

Bottom line: it is good practice to have pure function that have input and output being PyTree (then it should works for all JAX transformation), and simplifying the closure like get rid of the `kernel_factory` is better.

rlouf:
I'm not sure I understand; what are you suggesting we should do?

For lower-lewel usage I would personally recommend the following:

```python
import blackjax

schedule = blackjax.adaptation.window_adaptation.schedule(num_steps)
adapt_init, adapt_update, _ = blackjax.adaptation.window_adaptation.base()  # Find better name, actually

nuts_step = blackjax.nuts.kernel()  # I don't like this, probably should be `new_kernel` or `build_kernel`
nuts_init = blackjax.nuts.init

state = nuts_inits(logprob_fn, init_position)
adapt_state = init(init_position)

# - `AdaptationState.parameters` should be a dictionary and the extra parameters
#    passed to `window_adaptation` should be be added to it at initialization so we
#    can do this sort of things.
#  - We can `vmap` the two steps below over `schedule` and `rng_keys` and do anything else we want
#    with the results between steps
state, info = nuts_step(rng_key, state, **adapt_state.parameters)
adapt_state = adapt_update(adaptation_state, schedule[0], state.position, info.acceptance_probability)
```

If you're digging at this level writing a few extra lines of code shouldn't be a big deal, it also makes it clearer exactly what is happening. If you encourage people to do fancy stuff with the high-level API you will get hacky and unreadable code imho. 

Modularity is and should remain a core principle in blackjax, and we should keep this in mind when we refactor. Otherwise we can easily fall into this trap of _making the high-level API do more things_, like most libraries do and which results in poor code quality *and* clarity.

junpenglao:
Fully agree what you said on not making high-level API do too much and keeping the modularity as core.

IIUC there is not much concern of the the top most high-level API, as it would be just replicating what numpyro and pymc is doing (i.e., just give me back MCMC samples). What we are trying to refactor here is the intermediate use case. 

I think the general pattern of what you propose make sense, what do you think about pattern like:
 
```python
state, sampler_info = sampler_one_step(rng_key, state, **adapt_state.parameters)
adapt_state, adapt_info = adapt_update(adapt_state, adapt_info, state.position, sampler_info.acceptance_probability)
adapt_info = schedule_update(adapt_info)
```

Why having adapt_info update separate? Because we have window adaptation which works like a for loop of different smaller sampling routine. rewriting that into schedule update we can then have 1 single sampling loop. So now schedule update is aware of current iteration
- if current_iter > num_warm_up `adapt_update` is none op
- if current_iter <= num_warm_up
  - update step size or both step size and mass_matrix according to which window it is currently in.

rlouf:
While I fully agree with we should adopt the `adapt_state, adapt_info` like anywhere else in the library (that's a good pattern),  I don't think we should use a `schedule_update` function.

I considered it initially, but the issue is that, at least for window adaptation _what we do at each step depends on the total number of adaptation steps_. We could not, for instance, return the schedule as a generator and then use for as many steps we wish. Your proposal hides this fact from the user, and may lead to errors they do not understand. The best way imo is to make these kinds of mistake **impossible to make**.

We can re-evaluate this for adaptation schemes that don't need to know the number of adaptation steps in advance. For instance, the `pathfinder` adaptation doesn't need any schedule at all since you're essentially running the dual averaging algorithm for as long as you like.

rlouf:
I think I made a mistake and my code is gone :(

rlouf:
I think we should consider the possiblity of using another optimizer than dual averaging for the step size adaptation in the design of the internals.

rlouf:
I realized with `MEADS`that the `kernel_factory` function can actually be very handy. We can define very generally a kernel factory as a function that takes an algorithm step function and returns a kernel:

```python
def kernel_factory(**parameters, logprob_grad) -> KernelType
```

Where `KernelType` designates every function with the following "kernel" signature:

```python
def kernel(rng_key, state) -> Tuple[StateType, Info]
```

I am starting to wonder if the base modules should have a `kernel_factory` function, e.g. `blackjax.mcmc.nuts.kernel_factory`. This needs to be part of a *semantic clarification* effort throughout blackjax. Here we would have:
- Step functions (? find better name) that take the key state, and all parameters to return a new state. `StepType`
- Kernels take key, state and return a new state `KernelType`
- Kernel factories take step functions and turn them into kernels

In passing, the great thing with functions of type `KernelType` is that they can be trivially transformed into universal functions:

```python
jax.vmap(kernel)
jax.pmap(kernel, device=...)
```

which will then operate element-wise along the first dimension of the arrays passed to them.

rlouf:
@junpenglao could you review this? It's almost ready to merge.

blackjax/adaptation/meads.py
```diff
@@ -1,35 +1,38 @@
...
+    delta: float
+
+
 def base(
-    kernel_factory: Callable,
     logprob_grad_fn: Callable,
     batch_fn: Callable = jax.vmap,
 ):
-    """Maximum-Eigenvalue Adaptation of Damping and Step size for the Generalized
+    """Maximum-Eigenvalue Adaptation of damping and dtep size for the generalized
```
junpenglao, line 20:
```suggestion
    """Maximum-Eigenvalue Adaptation of damping and step size for the generalized
```

blackjax/adaptation/meads.py
```diff
@@ -77,49 +81,63 @@ def maximum_eigenvalue(matrix: PyTree):
...
+        """
+        current_iter = adaptation_state.current_iteration
+        step_size, alpha, delta = parameter_gn(positions, current_iter)
 
-    def final(last_state: chain_adaptation.ChainState) -> PyTree:
-        parameters = parameter_gn(
-            last_state.states,
-            last_state.current_iter,
+        return MEADSAdaptationState(
+            adaptation_state.current_iteration + 1, step_size, alpha, delta
```
junpenglao, line 140:
```suggestion
            current_iter + 1, step_size, alpha, delta
```

blackjax/adaptation/pathfinder_adaptation.py
```diff
@@ -18,109 +17,110 @@
...
-        step_size = jnp.exp(adaptation_state.da_state.log_step_size)
-        inverse_mass_matrix = adaptation_state.inverse_mass_matrix
-        kernel = kernel_factory(step_size, inverse_mass_matrix)
-
-        chain_state, chain_info = kernel(rng_key, chain_state)
-        new_da_state = da_update(adaptation_state.da_state, chain_info)
-        new_warmup_state = PathfinderAdaptationState(
-            new_da_state, adaptation_state.inverse_mass_matrix
-        )
+        new_ss_state = da_update(adaptation_state.ss_state, acceptance_rate)
```
junpenglao, line 114: Amazing - this is much better than initializing a kernel and calling it inside adaptation.

blackjax/kernels.py
```diff
@@ -785,47 +880,67 @@ def meads(
...
+
+    batch_init = batch_fn(lambda r, p: ghmc.init(r, p, logprob_fn, logprob_grad_fn))
 
-    def kernel_factory(step_size: PyTree, alpha: float, delta: float):
-        def kernel_fn(rng_key, state):
-            return kernel(
+    def one_step(carry, rng_key):
+        states, adaptation_state = carry
+
+        def kernel(rng_key, state):
```
junpenglao, line 894: How about
```suggestion
        def kernel_closure(rng_key, state):
```

blackjax/kernels.py
```diff
@@ -715,27 +721,116 @@ def run(rng_key: PRNGKey, position: PyTree):
...
+            return step_fn(rng_key, state, logprob_fn, **parameters)
+
+        return AdaptationResults(last_chain_state, kernel, parameters)
+
+    return AdaptationAlgorithm(run)
+
+
+def pathfinder_adaptation(
```
junpenglao, line 747: Could you move the pathfinder_adaptation to the bottom? it's easier to see the diff.

junpenglao (approved):
Really great job! Only minor comments.

junpenglao:
Also, remove https://github.com/blackjax-devs/blackjax/blob/main/blackjax/adaptation/chain_adaptation.py now that is all in MEADS

blackjax/kernels.py
rlouf: This one is truly a kernel in the sense of the original design document. However I generally agree with you that there's a naming issue across the library and we seriously need to work on it see #280 for instance.

blackjax/kernels.py
rlouf: Ok! We'll need another PR to re-arrange things in there. It's getting messy, hence #292

blackjax/adaptation/pathfinder_adaptation.py
rlouf: Glad you like it :)

blackjax/adaptation/meads.py
rlouf: good catch

junpenglao (dismissed):

PR #279, opened 2022-09-16
Refactor the SMC kernels

rlouf:
There are a few things that I find unsatisfactory with the SMC base kernel:

1. The use of `kernel_factory`, which prevents the users from passing different parameters at different iterations. It forces adaptation to happen **inside** the SMC kernel, which is something we don't want. We should stick to the sample, then update the parameters paradigm sketched in #276 
2. The internals of the base kernel could be better organised, following this pseudo code:

```python
new_particles, weights = vectorized_fn(particles)  # scatter computation
sampled_particles = sample_fn(new_particles, weights)  # gatherresults
```

I think this decomposition is what was conceptually missing to properly integrate #117

3. The current code uses `vmap`, but it should be possible to use `pmap` as wellj
4. `num_mcmc_iterations`, `mcmc_iter`, etc. naming is inconsistent.

The implementation of the SMC base kernel should be based on the formalism exposed in [this book](https://nchopin.github.io/books/).


rlouf:
A few thoughts:

- Base SMC should be implemented in the very general form where `smc_step_fn(rng_key, particles, particle_update_fn, particle_subsample_fn, particle_weighing_fn, resampling_fn)` where `particle_update_fn` and `particle_weighing_fn` are batched using `vmap` or `pmap`
- This entails a nice hierarchy for MCMC: `mcmc_step_fn < inference_loop < SMC(inference_loop)`
- Check how robust design is to $SMC^2$
- Make sure Waste-free SMC fits
- Show an example off tuning MCMC with SMC (new adaptation algorithm)
- Check how robust design is to $SMC-ABC$

rlouf:
@AdrienCorenflos could you please take a look? (especially `blackjax.smc.step` and `blackjax.smc.tempered` to see how the base is used)?

Left in this PR:
- [ ] Further simplify the Tempered SMC tests
- [ ] Refactor the ESS tests
- [ ] Use `log` for ESS by default (and `logsumexp` for numerical stability)

rlouf:
I am about done with this. The new interface allows to work with the Waste-Free version of SMC, one just has to design the appropriate "update" function and pass the desired number of samples extracted during resampling. I added a test that demonstrates this.

I am waiting for tonight's meeting to decide whether we move all the content of `smc` to a new `/meta` directory and rename `base.py -> smc.py`.

rlouf:
This is ready for review. We'll do the repo re-organisation in a separate PR so it is easily reversible.

PR #293, opened 2022-09-19
Allow users to pass a float for constant learning rate

rlouf:
The current API requires user to pass a function of the index of the current iteration, even when the learning rate is constant. However, in practice most people use a constant learning rate (even if they shouldn't, but who are we to judge?)  and we should simplify that pattern of use.

This commit thus allows users to pass either a function, either a float as learning rate to the kernel builder. It will raise an exception for any other type.


rlouf:
codevoc is not catching the test cases that check that a `TypeError` is effectively being raised for some reason.

junpenglao (approved):

PR #295, opened 2022-09-19
Truncate `$GITHUB_REF` at the right level when testing for `pip install` success

rlouf:
The workflow that checks that blackjax can be installed from PyPi currently fails because `$GITHUB_REF` is truncated before the leading version number (currently 0). This commit fixes this.


rlouf:
Yet another spurious codecov failure.

Issue #303, opened 2022-09-19
Replace `tensorflow-datasets` with a lightweight libraries for the SGMCMC example


rlouf:
Users should not have to install > 100 Mb just to be able to load the MNIST dataset.

junpenglao:
Maybe replace with scikit-learn? only ~7 mb

PR #304, opened 2022-09-19
Update examples

rlouf:
I updated the SGMCMC example, which now runs in no time:
- Evaluation now happens on test set
- `jax.jit` the step function

I also updated the Aesara example; Aesara has a much simpler JAX transpilation pipeline since `aesara=2.8.5`.

Finally I changed the tensorflow requirement to `tensorflow-cpu`. I leave #303 as a good first issue.

### DIFF
--- a/examples/SGMCMC.md
+++ b/examples/SGMCMC.md
@@ -6,10 +6,11 @@ jupytext:
     format_version: 0.13
     jupytext_version: 1.14.0
 kernelspec:
-  display_name: blackjax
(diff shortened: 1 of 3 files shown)


rlouf:
Notes on the tests that take the longest:
- The elliptical slice sampler is slow because the slice sampler works "too well": NUTS crawls for the same number of steps versus a few seconds.
- The Hierarchical BNN notebook could be much faster, especially the `get_predictions` function. I suspect this is where most of the time is spent, and not the sampling.

PR #338, opened 2022-09-22
s/BlackJAX/Blackjax in README

rlouf:


junpenglao:
We shouldnt make this change!!

Issue #318, opened 2022-09-21
Add "How to use custom gradients?" documentation


rlouf:
It's very simple: https://jax.readthedocs.io/en/latest/notebooks/Custom_derivative_rules_for_Python_code.html#python-debugging

PR #342, opened 2022-09-22
Remove `logprob_grad_fn` kwargs in favor of `jax.custom_vjp`

### LINKED ISSUE #341: Remove the `logprob_grad` keyword arguments
To use custom gradients users currently need to pass the function as a keyword arguments; this makes the API more complicated as we need to carry this argument downstream.

We can bypass this thanks to JAX allowing to register a custom gradient for functions. This is related to #318 and also #340 and probably #319 as well.

rlouf:
Custom gradient currently need to be passed to the algorithms all the way down to the integrators. However JAX has a much nicer way of doing that; you can write a logprob and register a gradient for that logprob; JAX will then use it when `jax.grad` is called on that function.

Closes #341 #318 

 - [x] The doc is up-to-date;

### DIFF
--- a/blackjax/adaptation/meads.py
+++ b/blackjax/adaptation/meads.py
@@ -1,4 +1,4 @@
-from typing import Callable, NamedTuple
+from typing import NamedTuple
 
 import jax
 import jax.numpy as jnp
@@ -26,10 +26,7 @@ class MEADSAdaptationState(NamedTuple):
     delta: float
 
 
-def base(
-    logprob_grad_fn: Callable,
-    batch_fn: Callable = jax.vmap,
-):
+def base():
     """Maximum-Eigenvalue Adaptation of damping and step size for the generalized
     Hamiltonian Monte Carlo kernel [1]_.
 
@@ -43,13 +40,6 @@ def base(
 
     This is an implementation of Algorithm 3 of [1]_ using cross-chain
     adaptation instead of parallel ensample chain adaptation.
-
-    Parameters
-    ----------
-    logprob_grad_fn
-        The gradient of logprob_fn, outputs the gradient PyTree for sample.
-    batch_fn
-        Either jax.vmap or jax.pmap to perform parallel operations.
 
     Returns
     -------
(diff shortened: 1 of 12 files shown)


rlouf:
We still need a small regression test to make sure that the new solution (relying on JAX's `custom_jvp` or `custom_vjp`) works, something with the `velocity_verlet` integrator should do, and to add a good example that illustrates the need for custom gradients (functions that contain `minimize` for instance).

rlouf:
One thing I've noticed: MEADS compute the gradients during the warmup; however those are computed by the GHMC step. We should use that instead.

blackjax/adaptation/meads.py
```diff
@@ -26,10 +26,7 @@ class MEADSAdaptationState(NamedTuple):
     delta: float
 
 
-def base(
-    logprob_grad_fn: Callable,
-    batch_fn: Callable = jax.vmap,
-):
+def base():
```
junpenglao, line 29: Seems unnecessary - should we just remove this and let the function exists in the namespace?

examples/howto_custom_gradients.md
```diff
@@ -0,0 +1,193 @@
...
+  name: python3
+---
+
+# Use custom gradients
+
+
+JAX provides a convenient `jax.grad` function to evaluate the gradient of any function build with JAX primitives. Which is why Blackjax uses `jax.grad` internally whenever it needs to evaluate the gradient. This should be enough for most applications, but sometimes you may need to provide your own gradients to blackjax for several reasons:
+
+- You have a convenient closed-form expression for the gradient that is evaluated faster than the gradient that JAX produces;
+- The forward-mode differentiation is faster than the backward-mode;
```
junpenglao, line 20: This is actually the initial motivation of adding this API - could you add a small example of `custom_jvp` with `jax.jacfwd()`?

rlouf: Do you have an example of a logprob where it makes sense to use `jax.jaxfwd()` instead of `jax.grad`?

junpenglao: [FedericoV](https://github.com/FedericoV) mentioned a stiff ODE: https://github.com/blackjax-devs/blackjax/issues/201#issuecomment-1255592363

rlouf: I didn't see how to do it immediately; I think the best way is to do something like

```python
import jax.numpy as jnp

def grad(x):
    return jnp.dot(x, jax.jacfwd(logprob_fn)(x))

# Override `jax.grad` in the namespace
```

PR #345, opened 2022-09-24
Add How-to reference on using logpdf built with another framework

### LINKED ISSUE #318: Add "How to use custom gradients?" documentation

rlouf:
Use an Aesara model compiled to Numba with Blackjax. The example is currently very slow, even after calling `jax.jit` on the NUTS kernel, even though the Numba logp evaluates faster thant its JAX equivalent.

Closes #318

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -74,6 +74,7 @@ with the relevant hardware acceleration support.
 
    howto_use_ppl
    Sample with multiple chains?<examples/howto_sample_multiple_chains.md>
+   Use non-JAX log-prob functions?<examples/howto_other_frameworks.md>
 
 .. toctree::
    :maxdepth: 1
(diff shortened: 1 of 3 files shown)


junpenglao (approved):

rlouf:
We can do better by using custom XLA calls for the function, which will prevent back-and-forth with Python in the inner loop. I had an inkling this is what was happening, and this was confirmed by Sharad Vikram who also shared an example of such custom calls:

https://github.com/jax-ml/jax-triton/blob/main/jax_triton/triton_call.py

We´d need to add at least a note about this and maybe, better, an example.

PR #373, opened 2022-10-05
Update change_of_variable_hmc.md

### LINKED ISSUE #371: Fix "Change of variable in HMC" example's use of window adaptation
The notebook tries to adapt the parameters for multiple chains, and ends up making the same computation twice. Vectorize the whole adaptation + sampling loop instead.

junpenglao:
Fix #371

I did not combine the warmup and inference in a single compiled function, but opt to showcase how you can vmap warmup and sampling separately.


rlouf:
LGTM

Issue #163, opened 2022-02-01
Track library performance over time


rlouf:
We need to prevent performance regression when changes are made on the libray's core. We should thus select a few purposefully hard logprob densities to sample from and setup a couple of benchmarks, e.g. using `pytest-benchmarks`. 

The results of benchmarks should be kept somwhere for comparisons between branches. Since blackjax can run with models implemented in `aeppl`, `tfp`, `numpyro` DSL we can imagine continuously benchmarking against these libraries and adding the results to the documentation automatically.

junpenglao:
For v1 release, here are some items to close out:

- [ ] continue tracking with https://github.com/marketplace/actions/continuous-benchmark
- [ ] add more models
- [ ] specific benchmarking of different component (e.g., seperating warmup with sampling, and warmup+sampling end to end)
- [ ] add more metrics

rlouf:
In case that's still of interest, [airspeed velocity](https://github.com/airspeed-velocity/asv) was designed for this. It works well out of the box, and is used by libraries like NumPy.

PR #380, opened 2022-10-13
Fix benchmark

### LINKED ISSUE #379: test_benchmark.py currently broken
The benchmark is currently broken:
- very slow as we jit nothing
- the output MCMC chain is stuck (return same value), ~~could be because wrong log_prob #378~~ the warmup API was called incorrectly.

junpenglao:
Close #379.

Further improvement of benchmark will be tracked in #163


rlouf:
Good catch!

PR #388, opened 2022-10-24
Modify linear regressions to parameters in R

### LINKED ISSUE #378: Improve the model we use in test and benchmark
Currently, in test and benchmark we use a simple linear regression, with the logprob function as:
```python
def regression_logprob(scale, coefs, preds, x):
    """Linear regression"""
    scale_prior = stats.expon.logpdf(scale, 1, 1)
    coefs_prior = stats.norm.logpdf(coefs, 0, 5)
    y = jnp.dot(x, coefs)
    logpdf = stats.norm.logpdf(preds, y, scale)
    return sum(x.sum() for x in [scale_prior, coefs_prior, logpdf])
```

However, the free parameter `scale` is positive only, which can create problem for HMC. We should rewrite the logprob func so that the free parameters are in $R$ (e.g., what we are doing in https://blackjax-devs.github.io/blackjax/examples/change_of_variable_hmc.html)

albcab:
Closes #378 

As a side note: If initial parameters for the `log_scale` are set to $1.0$ in `regression_test_cases` only `test_pathfinder_adaptation` fails with samples for `log_scale` diverging to `inf`.

### DIFF
--- a/tests/test_benchmarks.py
+++ b/tests/test_benchmarks.py
@@ -15,9 +15,10 @@ import pytest
 import blackjax
 
 
-def regression_logprob(scale, coefs, preds, x):
+def regression_logprob(log_scale, coefs, preds, x):
     """Linear regression"""
-    scale_prior = stats.expon.logpdf(scale, 1, 1)
+    scale = jnp.exp(log_scale)
+    scale_prior = stats.expon.logpdf(scale, 0, 1) + log_scale
     coefs_prior = stats.norm.logpdf(coefs, 0, 5)
     y = jnp.dot(x, coefs)
     logpdf = stats.norm.logpdf(preds, y, scale)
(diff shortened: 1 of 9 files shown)


rlouf:
Thanks! Why use flat priors?

junpenglao:
I think it is better to keep the prior and add the forward log jacobian. @rlouf wdyt?

rlouf:
> I think it is better to keep the prior and add the forward log jacobian. @rlouf wdyt?

We had the same thought at the same time 😄

junpenglao:
Could you also update the one in:

- [ ] tests/test_optimizers.py
- [ ] tests/test_benchmarks.py
- [ ] tests/test_tempered_smc.py

Optional the ones in examples.

tests/test_benchmarks.py
```diff
@@ -15,12 +15,12 @@
 import blackjax
 
 
-def regression_logprob(scale, coefs, preds, x):
+def regression_logprob(log_scale, coefs, preds, x):
     """Linear regression"""
-    scale_prior = stats.expon.logpdf(scale, 1, 1)
+    scale_prior = stats.expon.logpdf(jnp.exp(log_scale), 0, 1) + log_scale
```
junpenglao, line 20:
```suggestion
    scale = jnp.exp(log_scale)
    scale_prior = stats.expon.logpdf(scale, 0, 1) + log_scale
```
Same below

junpenglao (commented):
Looks good, minor comments.

albcab:
Great. Added a reminder in the introduction notebook to reparametrize before MCMC.

rlouf:
LGTM, thank you for doing the unsexy work!

rlouf (approved):

PR #392, opened 2022-10-27
Update the pathfinder API

### LINKED ISSUE #282: Refactor the Pathfinder API
Pathfinder is a variational inference algorithm, and works in two steps:
- Find the best approximation to the target distribution with optimisation;
- Generate samples from the approximation.

I suggest the following API that should be adopted by every VI algorithm in the future:

```python

pathfinder = blackjax.pathfinder(logprob_fn)
approximation = pathfinder.approximate(init_positions, max_num_steps)
samples = pathfinder.sample(rng_key, approximation, num_samples)
```

rlouf:
In this PR I update the pathfinder API to fit the variational inference semantics more closely. The `init`, `kernel` terminology is not adapted to variational inference: instead of using a markov kernel to update the state, vi algorithms compute an approximation to the target distribution and then sample from this approximation.

I thus propose the following API for pathfinder:

```python
import blackjax

pathfinder = blackjax.pathfinder(logprob_fn)
approximation, info = pathfinder.approximate(rng_key, initial_position)
samples, _ = pathfinder.sample(rng_key, approximation, num_samples)
```

Where `info` contains the full path (which is thus always returned), I created a new `VISamplingAlgorithm` type, and made a few simplifying tweaks to the implementation of pathfinder. Closes #282

### DIFF
--- a/blackjax/base.py
+++ b/blackjax/base.py
@@ -1,4 +1,4 @@
-from typing import NamedTuple, Tuple
+from typing import Callable, NamedTuple, Tuple
 
 from typing_extensions import Protocol
 
@@ -73,8 +73,8 @@ class UpdateFn(Protocol):
         """
 
 
-class SamplingAlgorithm(NamedTuple):
-    """A pair of functions that implement a sampling algorithm.
+class MCMCSamplingAlgorithm(NamedTuple):
+    """A pair of functions that represents a MCMC sampling algorithm.
 
     Blackjax sampling algorithms are implemented as a pair of pure functions: a
     kernel, that takes a new samples starting from the current state, and an
@@ -101,6 +101,30 @@ class SamplingAlgorithm(NamedTuple):
 
     init: InitFn
     step: UpdateFn
+
+
+class VIAlgorithm(NamedTuple):
+    """A pair of functions that represents a Variational Inference algorithm.
+
+    Blackjax variational inference algorithms are implemented as a pair of pure
+    functions: an approximator, which takes a target probability density (and
+    potentially a guide), and a sampling function that uses the approximation to
+    draw samples.
+
+    Attributes
+    ----------
+    approximate
+        A pure function, which when called with an initial position (and
+        potentially a guide function) returns a state that allows to build
+        an approximation to the target probability density function.
+    sample
+        A pure function which returns samples from the approximation computed
+        by `approximate`.
+
+    """
+
+    approximate: Callable
+    sample: Callable
 
 
 class RunFn(Protocol):
--- a/blackjax/__init__.py
+++ b/blackjax/__init__.py
@@ -11,6 +11,7 @@ from .kernels import (
     mgrad_gaussian,
     nuts,
     orbital_hmc,
+    pathfinder,
     pathfinder_adaptation,
     rmh,
     sghmc,
@@ -39,6 +40,7 @@ __all__ = [
     "pathfinder_adaptation",
     "adaptive_tempered_smc",  # smc
     "tempered_smc",
+    "pathfinder",  # variational inference
     "ess",  # diagnostics
     "rhat",
 ]
--- a/blackjax/kernels.py
+++ b/blackjax/kernels.py
@@ -1246,7 +1246,7 @@ class pathfinder:
             **lbfgs_parameters,
         ):
             return cls.approximate(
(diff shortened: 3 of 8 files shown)


rlouf:
@junpenglao ping

junpenglao:
Sorry about the delay, will take a look this week.

blackjax/base.py
```diff
@@ -103,6 +103,30 @@ class SamplingAlgorithm(NamedTuple):
...
+    functions: an approximator, which takes a target probability density (and
+    potentially a guide), and a sampling function that uses the approximation to
+    draw samples.
+
+    Attributes
+    ----------
+    approximate
+        A pure function, which when called with an initial position (and
+        potentially a guide function) returns a state that allows to build
+        an approximation to the target probability density function.
```
junpenglao, line 119: A non-blocking comment but this is something we need to refine further once we have more VI implementation. The part that need clarification is initial position, for MCMC and Pathfinder, `init_position` is the PyTree representing the parameter space where we can do `log_prob_fn(init_position)`. But for general VI, `init_position` is usually the parameterization of an approximation, for example in Meanfield ADVI:
```python
# init_param is a pytree where we can call 
# log_prob_fn(init_position)

def gen_meanfield_logprob(params):
    mu_param, rho_param = params
    sigma_param = jax.tree_map(jnp.exp, rho_param)
    def meanfield_logprob(position):
        logq_pytree = jax.tree_map(
            stats.norm.logpdf, position, mu_param, sigma_param
            )
        logq = jax.tree_map(jnp.sum, logq_pytree)
        return jax.tree_util.tree_reduce(jnp.add, logq)
    return meanfield_logprob

# gen_meanfield_logprob(init_params)(init_position)

def meanfield_sample(
    rng_key, meanfield_param, num_samples: int
    ):
    if not isinstance(num_samples, tuple):
        num_samples = (num_samples,)
    mu_param, rho_param = meanfield_param
    sigma_param = jax.tree_map(jnp.exp, rho_param)
    mu_flatten, unravel_fn = jax.flatten_util.ravel_pytree(mu_param)
    sigma_flatten, _ = jax.flatten_util.ravel_pytree(sigma_param)
    flatten_sample = jax.random.normal(
        rng_key, num_samples + mu_flatten.shape
        ) * sigma_flatten + mu_flatten
    if len(num_samples) == 0:
        return unravel_fn(flatten_sample)
    return jax.vmap(unravel_fn)(flatten_sample)

# meanfield_sample(rng, init_params, ())

def meanfield_approximate(rng, init_params, log_prob_fn, optimizer, sample_size=1, num_steps=200):
    def meanfield_approximate_step(
        state, rng_key_sample
        ):
        params, opt_state = state
        def kl_fn(params):
            sample = meanfield_sample(rng_key_sample, params, sample_size)
            logq = gen_meanfield_logprob(params)(sample)
            logp = log_prob_fn(sample)
            return (logq - logp).mean()
        # compute KL divergence
        elbo, grad = jax.value_and_grad(kl_fn)(params)
        updates, opt_state = optimizer.update(grad, opt_state, params)
        params = jax.tree_map(
            lambda p, u: p + u, params, updates
            )
        return (params, opt_state), elbo
    
    def run_optimization(init_params):
        opt_state = optimizer.init(init_params)
        state = (init_params, opt_state)
        rng_key = jax.random.split(rng, num_steps)
        return jax.lax.scan(
            meanfield_approximate_step, state, rng_key
            )
    
    return run_optimization(init_params)

optimizer = optax.chain(optax.clip(10.), optax.adam(1.))
output = meanfield_approximate(rng, init_params, log_prob_fn, optimizer)

# meanfield_vi = VISamplingAlgorithm(
#     meanfield_approximate,
#     meanfield_sample
# )
```

The high level summary here is that:
1. signature of `VISamplingAlgorithm.sample`. Both in the example above and pathfinder assuming it is `sample(rng_key, vi_param, num_samples: int):` where vi_param is part of the result of `approximate`
2. signature of `VISamplingAlgorithm.approximate`, should the init_param be init_position like mcmc or init_vi_param that parameterized the VI

junpenglao (approved):

blackjax/base.py
rlouf: You are right and I should probably replace `VISamplingAlgorithm` with `ApproximationAlgorithm` for now like you suggested.

PR #396, opened 2022-10-31
Onging project: add Contour sampler to achieve free mode exploration on MNIST dataset.

WayneDW:
# Thank you for opening a PR!

 - [ ] *If I add a new sampler*, there is an issue discussing it already;
 - [ ] We should be able to understand what the PR does from its title only;
 - [ ] There is a high-level description of the changes;
 - [ ] There are links to *all* the relevant issues, discussions and PRs;
 - [ ] The branch is rebased on the latest `main` commit;
 - [ ] `pre-commit` is installed and configured on your machine, and you ran it before opening the PR;
 - [ ] The doc is up-to-date;


blackjax/sgmcmc/diffusion.py
```diff
@@ -35,6 +35,102 @@ def one_step(
...
+
+
+class ContourState(NamedTuple):
+    position: PyTree
+    logprob_grad: PyTree
+    energy_pdf: PyTree
+    energy_idx: int
+
+
+def logposterior_estimator_fn(position: PyTree, data_batch: PyTree) -> float:
```
rlouf, line 53: I will refactor this out of `sgmcmc.gradients` in #299. Can you copy it there (even if that means it's implemented twice in that file for the time being?

blackjax/sgmcmc/diffusion.py
```diff
@@ -35,6 +35,102 @@ def one_step(
...
 
+""" 
+    We are interested in the simulations of exp(-U(x) / T), 
+    where U is energy function and T is the temperature.
+
+    logprob normally means U(x)/ T instead of U(x).
+"""
+
+
+class ContourState(NamedTuple):
```
rlouf, line 46: Would it be possible to add a docstring that explains what `energy_pdf` and `energy_idx` correspond to? The other two are self-explanatory.

blackjax/sgmcmc/diffusion.py
```diff
@@ -35,6 +35,102 @@ def one_step(
...
+        batch_loglikelihood(position, data_batch), axis=0
+    )
+
+
+def csgld(logprob_grad_fn, 
+          zeta: float=1, 
+          temperature: float=0.01, 
+          partition: int=512,
+          energy_gap: float=100
+          total_samples: float=60000):
```
rlouf, line 80: `total_samples` is not used in the function

rlouf: Would it be possible to describe the parameters in the docstring?

blackjax/sgmcmc/csgld.py
```diff
@@ -0,0 +1,59 @@
+"""Public API for the Contour Stochastic gradient Langevin Dynamics kernel.
+.. [0]: Deng, W., Lin, G., Liang, F. (2020). 
+        A Contour Stochastic Gradient Langevin Dynamics Algorithm 
+        for Simulations of Multi-modal Distributions. 
+        In Neural Information Processing Systems (NeurIPS 2020).
+
+   [1]: Deng, W., Liang, S., Hao, B., Lin, G., Liang, F. (2022) 
+        Interacting Contour Stochastic Gradient Langevin Dynamics
+        In International Conference on Learning Representations (ICLR)
```
rlouf, line 9:
```suggestion

References
----------
.. [0]: Deng, W., Lin, G., Liang, F. (2020). 
        A Contour Stochastic Gradient Langevin Dynamics Algorithm 
        for Simulations of Multi-modal Distributions. 
        In Neural Information Processing Systems (NeurIPS 2020).

   [1]: Deng, W., Liang, S., Hao, B., Lin, G., Liang, F. (2022) 
        Interacting Contour Stochastic Gradient Langevin Dynamics
        In International Conference on Learning Representations (ICLR) 
```

blackjax/sgmcmc/csgld.py
```diff
@@ -0,0 +1,59 @@
...
+    return CSGLDState(0, position, logprob_grad, energy_pdf, partition-1)
+
+
+def kernel(grad_estimator_fn: Callable) -> Callable:
+    integrator = csgld(grad_estimator_fn, 
+                       zeta=1,
+                       temperature=0.01,
+                       partition=512,
+                       energy_gap=10,
+                       total_samples=60000)
```
rlouf, line 45: These parameters should be passed as arguments to `kernel` (besides `total_samples` which does not seem to be used).

rlouf:
This looks great, the code is very readable! A few things are missing, but what's there is a very good start:

1. You may want to install and run `pre-commit` in the repo. At the very least, `black` should reformat your files, and `flake8` detect the unused argument;

blackjax/sgmcmc/diffusion.py
WayneDW: Gotcha will do!

blackjax/sgmcmc/diffusion.py
WayneDW: No problem, I will add an explanation very soon.

blackjax/sgmcmc/diffusion.py
WayneDW: No problem. The reason of using total_samples is that in maximum likelihood, the loss is summed instead averaged. If we want to apply a sampler to data-driven problems. We may need to decide if we want to change the whole averaged loss to summed loss (that is why I am using total_samples).

WayneDW:
> This looks great, the code is very readable! A few things are missing, but what's there is a very good start:
> 
> 1. You may want to install and run `pre-commit` in the repo. At the very least, `black` should reformat your files, and `flake8` detect the unused argument;
> 2. We want to export this at the higher level `blackjax.csgld`, to do so you will need to add an implementation in `kernels.py`;
> 3. We need a test for this, along the lines of the ones that are already there for SGMCMC algorithms;
> 4. Finally an example would be really nice, a comparison with the other SGMCMC algorithms even better :)

No problem. After I run `pre-commit` in the repo, I can import this sampler like the one below? 

from blackjax.sgmcmc.gradients import grad_estimator

blackjax/sgmcmc/csgld.py
WayneDW: Not sure what the kernel means, let me check carefully and let you know later.

rlouf:
> No problem. After I run `pre-commit` in the repo, I can import this sampler like the one below? 

`pre-commit` is just a tool that checks that the style of your code adheres to Blackjax's conventions.

Do you mean being able to import it as?

```python 
from blackjax import csgld
```

WayneDW:
> > No problem. After I run `pre-commit` in the repo, I can import this sampler like the one below?
> 
> `pre-commit` is just a tool that checks that the style of your code follows Blackjax's conventions.
> 
> Do you mean being able to import it as?
> 
> ```python
> from blackjax import csgld
> ```

Yes, in this way, I may be able to mimic your MNIST/SGHMC example using csgld. If there are other ways to test this code, please also feel free to suggest them.

rlouf:
You will need to add CSGLD to [this file](https://github.com/blackjax-devs/blackjax/blob/main/blackjax/kernels.py) and then import this in `__init__.py`.

If this doesn't make sense I can do it whenever I have time.

WayneDW:
> You will need to add CSGLD to [this file](https://github.com/blackjax-devs/blackjax/blob/main/blackjax/kernels.py) and then import this in `__init__.py`.
> 
> If this doesn't make sense I can do it whenever I have time.

Let me try. Have a great weekend.

blackjax/sgmcmc/csgld.py
WayneDW: do you mean i should define 
def kernel(grad_estimator_fn, zet, temperature, etc)?

WayneDW: > These parameters should be passed as arguments to `kernel` (besides `total_samples` which does not seem to be used).

do you mean i should define
def kernel(grad_estimator_fn, zet, temperature, etc)?

blackjax/sgmcmc/diffusion.py
WayneDW: > I will refactor this out of `sgmcmc.gradients` in #299. Can you copy it there (even if that means it's implemented twice in that file) for the time being?

Could you elaborate on this part? I don't know what to do next.

blackjax/sgmcmc/csgld.py
WayneDW: Solved it.

WayneDW:
> > No problem. After I run `pre-commit` in the repo, I can import this sampler like the one below?
> 
> `pre-commit` is just a tool that checks that the style of your code follows Blackjax's conventions.
> 
> Do you mean being able to import it as?
> 
> ```python
> from blackjax import csgld
> ```

pre-commit run --all-files works but make test fails
"
JAX_PLATFORM_NAME=cpu pytest -n 4 --cov=blackjax --cov-report term --cov-report html:coverage tests
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: -n --cov=blackjax --cov-report term --cov-report html:coverage tests
"

Do you have similar issues before?

rlouf:
Hey- I'm going to fix the merge conflicts in your branch and fix any other error; you will have to make sure to pull it locally before making any other modification.

WayneDW:
> Hey- I'm going to fix the merge conflicts (due to the changes in#299) in your branch and fix any other error; you will have to make sure to pull it locally before making any other modification.

Got it, thanks a lot. Happy thanksgiving day!

rlouf:
I rebased your branch on the current `main` branch, and made your code match the new API for stochastic gradient MCMC algorithms. I have a few comments on the next steps and a question; please excuse my current ignorance regarding your paper (I will read it before my next review!):

1. Are any of `zeta`, `energy_gap`, `temperature` susceptible to be changed during inference? Is there a world in which it would make sense? In this case we would need to pass them to `one_step` directly in `diffusions.py`
2. `energy_estimator_fn`requires a small refactor of the library (which I will do), I suggest that we have a `blackjax.value_and_grad_estimator` function to mimick `jax.value_and_grad` that would build the energy estimator and the estimator of its gradient.
3. We'll need to add a test
4. And finally add an example.

I'll work on (2) and (3) tomorrow, happy Thanksgiving!

WayneDW:
2. `energy_estimator_fn`requires a small refactor of the library (which I will do), I suggest that we have a `blackjax.value_and_grad_estimator` function to mimick `jax.value_and_grad` that would build the energy estimator and the estimator of its gradient.

Reply to 2: Yeah, we probably need an independent function for energy_estimator_fn, which is required by many interesting advanced samplers.

WayneDW:
> 1. Are any of `zeta`, `energy_gap`, `temperature` susceptible to be changed during inference? Is there a world in which it would make sense? In this case we would need to pass them to `one_step` directly in `diffusions.py`

These three can be just treated as hyperparameters. 

1. **zeta** should be fine-tuned;

2. **energy_gap** should be specified before training (no fine-tune), e.g. if we use the vanilla training and find the range of loss goes from 1e4 to 0, then we may set energy_gap = 150 (some buffer is helpful) with around 100 partitions. 
3. **temperature**: due to the cold posterior phenomenon in DNN, the temperature is often set to 1e-2 or similar values. If we fix the temperature, it is more appropriate for sampling; if we anneal it, it is better for global optimization.

rlouf:
Thanks! Just wanted to say while I'm learning about the algorithm: your [blog post](https://waynedw.github.io/posts/CSGLD/) is very well-written!

rlouf:
> 3. **temperature**: due to the cold posterior phenomenon in DNN, the temperature is often set to 1e-2 or similar values. If we fix the temperature, it is more appropriate for sampling; if we anneal it, it is better for global optimization.

Is that really necessary in practice? I've recently read this paper that claimed the opposite: https://arxiv.org/abs/2104.14421

rlouf:
With the current code on the branch (don't forget to `pull --rebase` locally before working on this code again!), the following code outputs a result:

```python
import jax
import jax.numpy as jnp

import blackjax
import blackjax.sgmcmc.gradients as gradients

rng_key, data_key = jax.random.split(rng_key, 2)

data_size = 1000
X_data = jax.random.normal(data_key, shape=(data_size, 5))
data_batch = X_data[:100, :]

init_position = 1.0

def logprior_fn(position):
    return -0.5 * jnp.dot(position, position) * 0.01

def loglikelihood_fn(position, x):
    w = x - position
    return -0.5 * jnp.dot(w, w)

# Build the CSGDL sampling step and init functions
logprob_fn, grad_fn = gradients.logprob_and_grad_estimator(logprior_fn, loglikelihood_fn, data_size)
csgld = blackjax.csgld(
    logprob_fn,
    grad_fn,
    zeta=1.,  # can be specified at each step in lower-level interface
    temperature=1e-2,  # can be specified at each step
    num_partitions=12, # cannot be specified at each step
    energy_gap=1000, # cannot be specified at each step
)

# Initialize and take one step using the CSGLD algorithm
init_state = csgld.init(init_position)
new_state = csgld.step(rng_key, init_state, data_batch, 1e-3, 1e-3)
print(new_state)
# CSGLDState(position=DeviceArray(-4.366815, dtype=float32), energy_pdf=DeviceArray([0.15382446, 0.14114678, 0.12818706, 0.11536834, 0.10254964,
#              0.08973093, 0.07691223, 0.06409353, 0.05127482, 0.03845612,
#              0.02563741, 0.01281871], dtype=float32), energy_idx=1)
```

... but it doesn't mean it yields correct samples! This is where I'll hand the PR over to you again. We need the following:

1. A test. If there's a target energy function for which you have theoretical convergence results then we should turn this result into a test for `csgld`. Otherwise we'll figure something out.
2. A working example.

I also have a question regarding `step_size_stoch`: is there a situation in which we'd want this step size to vary over time or does it always remain fixed?

WayneDW:
> > 3. **temperature**: due to the cold posterior phenomenon in DNN, the temperature is often set to 1e-2 or similar values. If we fix the temperature, it is more appropriate for sampling; if we anneal it, it is better for global optimization.
> 
> Is that really necessary in practice? I've recently read this paper that claimed the opposite: https://arxiv.org/abs/2104.14421

You can check this paper: https://arxiv.org/pdf/2002.02405.pdf. Data augmentation is a key factor that leads to the cold posterior effect. e.g. cifar has 50K images, but after data augmentation, we have many more images (>>50K). For the sum of maximum likelihood, it might be inappropriate to sum the likelihood from 1 to 50K. More importantly, we don't know what is an exact appropriate number.

WayneDW:
> 



> Thanks! Just wanted to say while I'm learning about the algorithm: your [blog post](https://waynedw.github.io/posts/CSGLD/) is very well-written!
> 
> One thing I am noticing, both in the code and in the paper, is that Contour SgLD is a "meta-algorithm", in the sense that it uses the Langevin dynamics as a base transition kernel, but this transition kernel is called by another process that changes the base measure of the original density (and carries a latent vector) to make the transition kernel's job easier. I am wondering if we could make this more explicit in the code, which would be very in line with the rest of the Blackjax codebase where we try to implement basic elements and combine them to get more complex algorithms.
> 
> Which leads me to a naive question: could this be generalized to be used with other Langevin dynamics?

This is a great question, a short answer is yes, we can use SGLD or SGHMC as the base sampler. Maybe we can first implement the SGLD version of contour and then generalize it in the next version.

WayneDW:
> ```python
> new_state = csgld.step(rng_key, init_state, data_batch, 1e-3, 1e-3)
> ```

That is great, I am working on it now and will try to add a demo. Thanks a lot for the updates.

Great catch on the choice of step_size_stoch. Yes, annealing it is more theoretically grounded to make it converge to a fixed point, especially when we can obtain the exact energy estimator; however, in practical approximation tasks, we may set it to a constant to facilitate fine-tuning.

WayneDW:
> With the current code on the branch (don't forget to `pull --rebase` locally before working on this code again!), the following code outputs a result:
> 
> ```python
> import jax
> import jax.numpy as jnp
> 
> import blackjax
> import blackjax.sgmcmc.gradients as gradients
> 
> rng_key = jax.random.PRNGKey(0)
> rng_key, data_key = jax.random.split(rng_key, 2)
> 
> data_size = 1000
> X_data = jax.random.normal(data_key, shape=(data_size, 5))
> data_batch = X_data[:100, :]
> 
> init_position = 1.0
> 
> def logprior_fn(position):
>     return -0.5 * jnp.dot(position, position) * 0.01
> 
> def loglikelihood_fn(position, x):
>     w = x - position
>     return -0.5 * jnp.dot(w, w)
> 
> # Build the CSGDL sampling step and init functions
> logprob_fn, grad_fn = gradients.logprob_and_grad_estimator(logprior_fn, loglikelihood_fn, data_size)
> csgld = blackjax.csgld(
>     logprob_fn,
>     grad_fn,
>     zeta=1.,  # can be specified at each step in lower-level interface
>     temperature=1e-2,  # can be specified at each step
>     num_partitions=12, # cannot be specified at each step
>     energy_gap=1000, # cannot be specified at each step
> )
> 
> # Initialize and take one step using the CSGLD algorithm
> init_state = csgld.init(init_position)
> new_state = csgld.step(rng_key, init_state, data_batch, 1e-3, 1e-3)
> print(new_state)
> # CSGLDState(position=DeviceArray(-4.366815, dtype=float32), energy_pdf=DeviceArray([0.15382446, 0.14114678, 0.12818706, 0.11536834, 0.10254964,
> #              0.08973093, 0.07691223, 0.06409353, 0.05127482, 0.03845612,
> #              0.02563741, 0.01281871], dtype=float32), energy_idx=1)
> ```
> 
> ... but it doesn't mean it yields correct samples! This is where I'll hand the PR over to you again. We need the following:
> 
> 1. A test. If there's a target energy function for which you have theoretical convergence results then we should turn this result into a test for `csgld`. Otherwise we'll figure something out.
> 2. A working example.
> 
> I also have a question regarding `step_size_stoch`: is there a situation in which we'd want this step size to vary over time or does it always remain fixed?

This is great. One more quick question, do we have a non-data driven example so that I can implement Figure 1 in [1]? For example, the posterior function is simply

def loglikelihood_fn(x):
    return jnp.log(0.4 * jax.scipy.stats.norm.pdf(x, loc=-2, scale=1client_loop: send disconnect: Broken pipe=1, scale=1)).

then we can demonstrate the performance using both SGLD (which get stuck in a local mode) and CSGLD.

[1] A Contour Stochastic Gradient Langevin Dynamics Algorithm for Simulations of Multi-modal Distributions.

rlouf:
You'll have to start over from scratch, but that's as good idea!

WayneDW:
> You'll have to write this test from scratch, but that's a good idea!

Hi, Rémi, I just updated sgmcmc/csgld.py, could you update the algorithm again? I am almost finishing the demo and found extra issues to fix.

rlouf:
Have you tried using `pytest` instead? I'm sorry the contributing instructions are not very good at the moment.

rlouf:
Also I noticed that the temperature is not used to rescale the noise currently; we may need to add a `temperature` parameter to `overdamped_langevin` that defaults to 1.

rlouf:
Strange. Would you mind pushing the test you've been working on? I'll take a look.

WayneDW:
> Strange. Would you mind pushing the test you've been working on? I'll take a look.

Did it. 

examples/how_to_use_csgld.py

rlouf:
I updated the code so the python script runs. I also added support for != 1 temperature, but I am not sure this is correct so I would appreciate if you could take a look (and if you have time add support for temperature != 1 in SgHMC). 

It looks like the results are not correct however, so we'll have to find what's wrong in the implementation. Once this returns correct results we'll add a simple test, I'll convert the example to a notebook and it will be good to merge!

_Don't forget to pull my changes locally before making any other changes!_

WayneDW:
> I updated the code so the python script runs. I also added support for != 1 temperature, but I am not sure this is correct so I would appreciate if you could take a look (and if you have time add support for temperature != 1 in SgHMC).
> 
> It looks like the results are not correct however, so we'll have to find what's wrong in the implementation. Once this returns correct results we'll add a simple test, I'll convert the example to a notebook and it will be good to merge!
> 
> _Don't forget to pull my changes locally before making any other changes!_

Your treatment of temperature for SGLD is correct. I have also added the temperature variable for SGHMC. Here is a link for a reference: https://proceedings.neurips.cc/paper/2020/file/cebd648f9146a6345d604ab093b02c73-Paper.pdf

WayneDW:
> I updated the code so the python script runs. I also added support for != 1 temperature, but I am not sure this is correct so I would appreciate if you could take a look (and if you have time add support for temperature != 1 in SgHMC).
> 
> It looks like the results are not correct however, so we'll have to find what's wrong in the implementation. Once this returns correct results we'll add a simple test, I'll convert the example to a notebook and it will be good to merge!
> 
> _Don't forget to pull my changes locally before making any other changes!_

Another question is that I added min_energy=0 in the file blackjax/sgmcmc/csgld.py. Do you know where to apply it in the demo: examples/how_to_use_csgld.py?

blackjax/kernels.py
```diff
@@ -623,6 +624,48 @@ def step_fn(rng_key: PRNGKey, state, minibatch: PyTree, step_size: float):
...
+    kernel = staticmethod(sgmcmc.csgld.kernel)
+
+    def __new__(  # type: ignore[misc]
+        cls,
+        logprob_estimator_fn: Callable,
+        grad_estimator_fn: Callable,
+        zeta: float = 1,
+        temperature: float = 0.01,
+        num_partitions: int = 512,
+        energy_gap: float = 100,
```
rlouf, line 639: This is where you would add the `min_energy` parameter, I believe

blackjax/kernels.py
```diff
@@ -623,6 +624,48 @@ def step_fn(rng_key: PRNGKey, state, minibatch: PyTree, step_size: float):
...
+    ) -> MCMCSamplingAlgorithm:
+
+        step = cls.kernel(num_partitions, energy_gap)
```
rlouf, line 642: And then pass it here

blackjax/sgmcmc/csgld.py
```diff
@@ -0,0 +1,174 @@
...
+
+    Parameters
+    ----------
+    num_partitions
+        The number of partitions we divide the energy landscape into.
+    energy_gap
+        The difference in energy :math:`\Delta u` between the successive
+        partitions. Can be determined by running e.g. an optimizer to determine
+        the range of energies. `num_partition` * `energy_gap` should match this
+        range.
```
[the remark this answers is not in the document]
WayneDW: Done.

blackjax/kernels.py
WayneDW: Done. Could you update the file so that I can continue my test?

rlouf: Yes you should be able to if you installed the library with `pip -e .`

WayneDW: > Yes you should be able to if you installed the library with `pip install -e .` in the environment where you're executing your example.

Great, probably my last question: do you know how to update new_state.position? CSGLD becomes so explorative, it may explore the extreme tails and lead to nan values. To avoid it, I have to limit the particle in a bounded domain such that  whenever some extreme values occur, we reset new_state.position to some nice values 

Screen Shot 2022-12-12 at 12 13 06 PM

WayneDW: > Yes you should be able to if you installed the library with `pip install -e .` in the environment where you're executing your example.

Screen Shot 2022-12-12 at 12 14 55 PM

rlouf: Try something like

```python 
new_state = CSGLDState(..., state.energy_pdf, state.energy_idx)
```

WayneDW: It is just because CSGLD is sooooo great at exploration and is such an expert in avoiding local optimal. 

Imagine that the standard normal distribution, CSGLD can easily explore tails such as 100. the probability or other relevant tools in this value is nan.

WayneDW: > 

We can consider to solve issue in future updates. Let me show a perfect demo first.

WayneDW: > Try something like
> 
> ```python
> new_state = CSGLDState(4, state.energy_pdf, state.energy_idx)
> ```
> 
> But if you're afraid of getting `NaN` we should probably handle this in a nicer way within the library. What would be the source of those `Nan` values?

Great News! I have implemented the Contour SGLD algorithm and it works perfectly.

image

WayneDW: > Try something like
> 
> ```python
> new_state = CSGLDState(4, state.energy_pdf, state.energy_idx)
> ```
> 
> But if you're afraid of getting `NaN` we should probably handle this in a nicer way within the library. What would be the source of those `Nan` values?

One thing is that I cannot write md in my side because I am using a server that doesn't allow jupyter notebook.

Is it appropriate to ask you to write a md file? I will prepare a python file that is as close to the md file as possible to minimize your cost.

WayneDW: Here is the example python code to reproduce the perfect example: https://github.com/WayneDW/Jax_implementation_of_CSGLD/blob/main/howto_use_csgld_final.py


The generated figures are all presented in the github repo: https://github.com/WayneDW/Jax_implementation_of_CSGLD.git

WayneDW: You can take a look now.

rlouf:
I am editing the example to a notebook format, and I have a couple questions:

1. You recompute `energy_idx` in the sampling loop; couldn't we use `state.energy_idx` instead?
2. Why only considering the top 5% partitions when doing importance sampling?

rlouf:
Also when I plot `scaled_density` I get the following profile:

download

Is this constant decrease expected? How do I get the nice illustration you have on your blog post from this?

WayneDW:
> Also when I plot `scaled_density` I get the following profile:
> 
> download
> 
> Is this constant decrease expected? How do I get the nice illustration you have on your blog post from this?

Yes. 

1. The lowest energy for this dataset is around 3700, which normally has the biggest probability mass. Since energy_gap=0.25, the idx with the highest value is around 15000, which makes sense;
2. A larger value of scaled_density means a larger penalty to stay in that region; that is where a negative learning rate/ bouncy move occurs.

image

WayneDW:
> Also when I plot `scaled_density` I get the following profile:
> 
> download
> 
> Is this constant decrease expected? How do I get the nice illustration you have on your blog post from this?

3. The energy_pdf is learning on the fly. When the algorithm converges, the red curve should be close to 0; but empirically we don't need to wait so long.
image

WayneDW:
4. Since the algorithm is an importance sampling algorithm, the large variance issue is inevitable.

image

WayneDW:
The large negative slope leads to the desired negative learning rate to bounce out of the local trap.

image

WayneDW:
> Thank you for your answers! I pushed the example as a notebook on the repository so you can see the figures. I'm bothered by the behavior of SLGD which stays stuck around 10. instead of quickly converging to one of the modes. Maybe I did something stupid while refactoring, I need to check.
> 
> Once that's figured out we can start adding some explanations of the algorithm on the example; we may want to use all the partitions for important sampling as this may otherwise be too much information for first-time users.

Maybe we can use some nice words, like "using tricks with the cyclical/ cosine learning rate".

image

WayneDW:
> Thank you for your answers! I pushed the example as a notebook on the repository so you can see the figures. ~I'm bothered by the behavior of SLGD which stays stuck around 10. instead of quickly converging to one of the modes. Maybe I did something stupid while refactoring, I need to check.~ Nevermind, I did something really stupid.
> 
> Once that's figured out we can start adding some explanations of the algorithm on the example; we may want to use all the partitions for important sampling as this may otherwise be too much information for first-time users.

No problem, I will try to use fewer partitions so that the re-sampling step is more efficient.

rlouf:
Ok, just let me push the fix for my stupid mistake first!

WayneDW:
> Ok, just let me push the fix for my stupid mistake first!

OK, make sure the temperatures used for SGLD and CSGLD are the same.

rlouf:
That's fixed, your turn! As you can see I decreased the number of iterations with little to no effect on the quality of the results.

WayneDW:
> your turn

What do you suggest to do next?

remove the code on the top 5% of partitions?

rlouf:
Yes and add some explanations. We should also take a step back now that we have an example to see if there is something that needs to be changed in the API. After I'll add a few tests and we should be good to merge!

WayneDW:
> Yes and add some explanations. We should also take a step back now that we have an example to see if there is something that needs to be changed in the API. After I'll add a few tests and we should be good to merge!

No problem.

WayneDW:
One thing that makes things harder to demonstrate is that the function doesn't allow us to explore much in tails

>>> def loglikelihood_fn(position, x):
...     mixture_1 = jax.scipy.stats.norm.pdf(x, loc=position, scale=sigma)
...     mixture_2 = jax.scipy.stats.norm.pdf(x, loc=-position + gamma, scale=sigma)
...     return jnp.log(0.5 * mixture_1 + 0.5 * mixture_2).sum()
... 
>>> 
>>> loglikelihood_fn(80, 5)
DeviceArray(-inf, dtype=float32)

To avoid re-starting when we go beyond the bounded domain, maybe the top X% partition is inevitable and I will use a cleaner way to code it.

WayneDW:
> > Since CSGLD is so good at exploring tails (which leads to high energy), we have to use more partitions; as such maybe the top X% partition is inevitable for efficiency purposes. I will use a cleaner way to code it.
> 
> Wouldn't using `jax.scipy.special.logsumexp` in the `logpdf` be more numerically stable?

Wow, that seems like a perfect function. Let me try!

rlouf:
You can take a look at [my implementation of Cyclical SGLD](https://www.thetypicalset.com/blog/cyclical_sgld.html).

WayneDW:
> 

Gotcha, this is really a wonderful function, I like it. This 25-mode mixture is more challenging and interesting. Let me see if CSGLD can be used at this moment.

WayneDW:
> You can take a look at [my implementation of Cyclical SGLD](https://www.thetypicalset.com/blog/cyclical_sgld.html), where I used it to define the log-density.

Maybe this should be the final version:

https://github.com/WayneDW/Jax_implementation_of_CSGLD/blob/main/howto_use_csgld_final_final.py


Updates: 

1. adopted a stable version of the mixture using log sum exp;
2. much simplified the re-sampling step (still used the 5% selection);
3. added some results to explain why the algorithm works;
4. include more comments on the pros and cons of the algorithm.

rlouf:
Modified the notebook to take your changes into account, this works perfectly. We just need to add some tests and we'll be good to merge. 

Unrelated question, could we apply the flattening/resampling idea to algorithms like HMC (full gradient)?

WayneDW:
> Modified the notebook to take your changes into account, this works perfectly. We just need to add some tests and we'll be good to merge.
> 
> Unrelated question, could we apply the flattening/resampling idea to algorithms like HMC (full gradient)?

I believe so, but I haven't rigorously tried yet. It should work.

WayneDW:
> Modified the notebook to take your changes into account, this works perfectly. We just need to add some tests and we'll be good to merge.
> 
> Unrelated question, could we apply the flattening/resampling idea to algorithms like HMC (full gradient)?

Actually, this is my first time merging an algorithm to a public repo. Very excited.

WayneDW:
One thing to comment: for the contour_sgld.md, I noticed that you included a module to re-start a particle when it goes beyond the domain. If you intentionally did it, that is OK; otherwise, you may check the code below:

https://github.com/WayneDW/Jax_implementation_of_CSGLD/blob/main/howto_use_csgld_final_final.py. In this code, we don't need to be concerned about going beyond the domain because a robust version of the mixture is implemented.

image

docs/examples/contour_sgld.md
```diff
@@ -0,0 +1,303 @@
...
+        csgld_sample_list = jnp.append(csgld_sample_list, state.position)
+        csgld_energy_idx_list = jnp.append(csgld_energy_idx_list, state.energy_idx)
+        pb.comment = (
+            f"| position {state.position: .2f} | re-restarts {re_start_counter}"
+        )
+```
+
+Contour SGLD takes inspiration from the Wang-Landau algorithm to learn the density of states of the model at each energy level, and uses this information to flatten the target density to be able to explore it more easily.
+
+As a result, the samples returned by contour SGLD are not from the target density directly, and we need to resample them using the density of state as importance weights to get samples from the target distribution.
```
zaxtax, line 216: Do we expect to do importance sampling like this more than once? Is this worth factoring into a library function?

zaxtax (commented):
I think this looks pretty good. Some improvements can be made, but I think the underlying algorithm implementation, tests, and example notebooks are top-notch

docs/examples/contour_sgld.md
WayneDW: Currently just once. Maybe we can think about this in the future. My current implementation is only acceptable and may not be fast enough, in the future version, we need to include some additional variables showing the important weight for each simulated sample.

blackjax/sgmcmc/gradients.py
```diff
@@ -11,23 +11,18 @@
...
 from blackjax.types import PyTree
 
 
-class GradientEstimator(NamedTuple):
-    init: Callable
-    estimate: Callable
-
-
-def estimator(
+def logprob_and_grad_estimator(
```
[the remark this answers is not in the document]
WayneDW: You can decide, whichever you think is efficient.

WayneDW:
> I think this looks pretty good. Some improvements can be made, but I think the underlying algorithm implementation, tests, and example notebooks are top-notch

Thanks for the nice words. To my best knowledge, this is the most promising sampler for multi-modal distributions (One of my academic family's (including COPSS Award winners: Wing-Hung Wong and Jun S. Liu) proudest works). It has been proven successful on MNIST level dataset. Although MNIST looks like a baby CV dataset, none of the existing samplers can achieve fluctuation losses like ours.

rlouf:
I didn’t see the change in the example, I will correct this before merging (it’s admittedly much cleaner this way!). I will take @zaxtax’s suggestions into account, maybe tweak a few things here and there in the API and we should be able to merge in the coming week.

WayneDW:
> I didn’t see the change in the example, I will correct this before merging (it’s admittedly much cleaner this way!). I will take @zaxtax’s suggestions into account, maybe tweak a few things here and there in the API and we should be able to merge in the coming week.

Fantastic!

zaxtax:
Yes! I don't want my remarks to be a blocker for this fantastic
contribution!

rlouf (approved):
This is good to merge. Thank you so much for your implementation and being patient through the whole process. Time to celebrate :partying_face:

WayneDW:
BTW, for some minor things:

1) the line of "domain_radius = 50" can be safely deleted;
2) # Build the SGDL sampler -> # Build the SGLD sampler
3) 100K iterations may probably make the result more appealing, but it is fine for now.

I will present more interesting results in the future. Especially to demonstrate it on MNIST dataset. This is the real one.

PR #421, opened 2022-12-11
Add adaptation on HMC trajectory length to increase the expected Change in the Estimator of the Expected Square (ChEES-HMC)

### LINKED ISSUE #382: Implement ChEES-HMC
I have a working draft of ChEES-HMC, except its not JAX-ready yet. The "Select jittered trajectory length..." line of [Algorithm 1](http://proceedings.mlr.press/v130/hoffman21a/hoffman21a.pdf) requires a float from a predetermined Halton sequence of length wamup+sampling number of iterations. When returning the kernel after warm-up, it needs to keep yielding from the sequence at every iteration (but JAX functions need to be pure).

For now my implementation yields the next number using a scipy:
```python
halton_sequence_gn = scipy.stats.qmc.Halton(d=1)
#then inside the kernel...
currrent_halton_number = jnp.array(halton_sequence_gn.random(n=1).squeeze())
```
Any help is appreciated. Will keep updated on progress.

albcab:
Closes #382

### DIFF
--- a/blackjax/adaptation/chees_adaptation.py
+++ b/blackjax/adaptation/chees_adaptation.py
@@ -16,7 +16,7 @@ from blackjax.types import Array, ArrayLikeTree, PRNGKey
 from blackjax.util import pytree_size
 
 # optimal tuning for HMC, see https://arxiv.org/abs/1001.4460
-TARGET_ACCEPTANCE_RATE = 0.651
+OPTIMAL_TARGET_ACCEPTANCE_RATE = 0.651
 
 
 class ChEESAdaptationState(NamedTuple):
@@ -24,35 +24,40 @@ class ChEESAdaptationState(NamedTuple):
 
     step_size
         Value of the step_size parameter of the HMC algorithm.
-    step_size_moving_average
-        Running moving average of the step_size parameter.
+    log_step_size_moving_average
+        Running moving average of the log step_size parameter.
     trajectory_length
         Value of the num_integration_steps / step_size parameter of
         the HMC algorithm.
-    trajectory_length_moving_average
-        Running moving average of the num_integration_steps / step_size
+    log_trajectory_length_moving_average
+        Running moving average of the log num_integration_steps / step_size
         parameter.
     optim_state
         Optax optimizing state for used to maximize the ChEES criterion.
     random_generator_arg
         Utility array for generating a pseudo or quasi-random sequence of
         numbers.
+    step
+        Current iteration number.
 
     """
 
     step_size: float
-    step_size_moving_average: float
+    log_step_size_moving_average: float
     trajectory_length: float
-    trajectory_length_moving_average: float
+    log_trajectory_length_moving_average: float
     da_state: dual_averaging.DualAveragingState
     optim_state: optax.OptState
     random_generator_arg: Array
+    step: int
 
 
 def base(
     jitter_generator: Callable,
     next_random_arg_fn: Callable,
     optim: optax.GradientTransformation,
+    target_acceptance_rate: float,
+    decay_rate: float,
 ) -> Tuple[Callable, Callable]:
     """Maximizing the Change in the Estimator of the Expected Square criterion
     (trajectory length) and dual averaging procedure (step size) for the jittered
@@ -75,6 +80,12 @@ def base(
         Function that generates the next `random_generator_arg` from its previous value.
     optim
         Optax compatible optimizer, which conforms to the `optax.GradientTransformation` protocol.
+    target_acceptance_rate
+        Average acceptance rate to target with dual averaging.
+    decay_rate
+        Float representing how much to favor recent iterations over earlier ones in the optimization
+        of step size and trajectory length.
+
 
     Returns
     -------
@@ -121,26 +132,30 @@ def base(
         """
         (
             step_size,
-            step_size_ma,
+            log_step_size_ma,
             trajectory_length,
-            trajectory_length_ma,
+            log_trajectory_length_ma,
             da_state,
             optim_state,
             random_generator_arg,
+            step,
         ) = initial_adaptation_state
 
         harmonic_mean = 1.0 / jnp.mean(
             1.0 / acceptance_probabilities, where=~is_divergent
         )
-        da_state_ = da_update(da_state, TARGET_ACCEPTANCE_RATE - harmonic_mean)
+        da_state_ = da_update(da_state, target_acceptance_rate - harmonic_mean)
         step_size_ = jnp.exp(da_state_.log_x)
-        new_step_size, new_da_state = jax.lax.cond(
+        new_step_size, new_da_state, new_log_step_size = jax.lax.cond(
             jnp.isfinite(step_size_),
-            lambda _: (step_size_, da_state_),
-            lambda _: (step_size, da_state),
+            lambda _: (step_size_, da_state_, da_state_.log_x),
+            lambda _: (step_size, da_state, da_state.log_x),
             None,
         )
-        new_step_size_ma = 0.9 * step_size_ma + 0.1 * new_step_size
+        update_weight = step ** (-decay_rate)
+        new_log_step_size_ma = (
+            1.0 - update_weight
+        ) * log_step_size_ma + update_weight * new_log_step_size
 
         proposals_mean = jax.tree_util.tree_map(
             lambda p: jnp.nanmean(p, axis=0), proposed_positions
@@ -186,30 +201,32 @@ def base(
             lambda _: (log_trajectory_length, optim_state),
             None,
         )
-        new_trajectory_length = jnp.exp(new_log_trajectory_length)
-        new_trajectory_length_ma = (
-            0.9 * trajectory_length_ma + 0.1 * new_trajectory_length
-        )
+        new_log_trajectory_length_ma = (
+            1.0 - update_weight
+        ) * log_trajectory_length_ma + update_weight * new_log_trajectory_length
+        new_trajectory_length = jnp.exp(new_log_trajectory_length_ma)
 
         return ChEESAdaptationState(
             new_step_size,
-            new_step_size_ma,
+            new_log_step_size_ma,
             new_trajectory_length,
-            new_trajectory_length_ma,
+            new_log_trajectory_length_ma,
             new_da_state,
             new_optim_state,
             next_random_arg_fn(random_generator_arg),
+            step + 1,
         )
 
     def init(random_generator_arg: Array, step_size: float):
         return ChEESAdaptationState(
             step_size=step_size,
-            step_size_moving_average=0.0,
+            log_step_size_moving_average=0.0,
             trajectory_length=step_size,
-            trajectory_length_moving_average=0.0,
+            log_trajectory_length_moving_average=0.0,
             da_state=da_init(step_size),
             optim_state=optim.init(step_size),
             random_generator_arg=random_generator_arg,
+            step=1,
         )
 
     def update(
@@ -260,6 +277,9 @@ def chees_adaptation(
     num_chains: int,
     *,
     jitter_generator: Optional[Callable] = None,
+    jitter_amount: float = 1.0,
+    target_acceptance_rate: float = OPTIMAL_TARGET_ACCEPTANCE_RATE,
+    decay_rate: float = 0.5,
 ) -> AdaptationAlgorithm:
     """Adapt the step size and trajectory length (number of integration steps / step size)
     parameters of the jittered HMC algorthm.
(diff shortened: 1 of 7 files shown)


rlouf:
Isn't it called ChEES? 😁

albcab:
Oh no :facepalm:

junpenglao:
@albcab time to pick this up again (no pressure lol)?

twiecki:
What's missing here? Would be great to get this sampler in!

albcab:
Need to rebase and resolve conflicts for all the remaining, restructuring, etc. that have been done during the past year. Other than that, the algorithm is done.

twiecki:
@albcab Great, let us know if you need any help!

twiecki:
Nice, seems like conflicts are solved, just some mypy errors:
```
blackjax/adaptation/chees_adaptation.py:257: error: Incompatible default for argument "jitter_seed" (default has type "None", argument has type "int")  [assignment]
blackjax/adaptation/chees_adaptation.py:257: note: PEP 484 prohibits implicit Optional. Accordingly, mypy has changed its default to no_implicit_optional=True
blackjax/adaptation/chees_adaptation.py:257: note: Use https://github.com/hauntsaninja/no_implicit_optional to automatically upgrade your codebase
tests/smc/test_tempered_smc.py:69: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
```

blackjax/adaptation/chees_adaptation.py
```diff
@@ -0,0 +1,423 @@
...
+    lengths.
+
+    Examples
+    --------
+
+    An HMC adapted kernel can be learned and used with the following code:
+
+    .. code::
+
+        warmup = blackjax.chees(logprob_fn, num_chains)
```
twiecki, line 286:
```suggestion
        warmup = blackjax.chees_adaptation(logprob_fn, num_chains)
```

tests/mcmc/test_sampling.py
```diff
@@ -223,6 +224,47 @@ def test_meads(self):
...
+
+        logposterior_fn_ = functools.partial(
+            self.regression_logprob, x=x_data, preds=y_data
+        )
+        logposterior_fn = lambda x: logposterior_fn_(**x)
+
+        init_key, warmup_key, inference_key = jax.random.split(rng_key, 3)
+
+        num_chains = 128
+        warmup = blackjax.chees(
```
twiecki, line 241:
```suggestion
        warmup = blackjax.chees_adaptation(
```

blackjax/adaptation/chees_adaptation.py
```diff
@@ -0,0 +1,419 @@
...
+                step_size=parameters["step_size"],
+                inverse_mass_matrix=jnp.ones(num_dim),
+                num_integration_steps=num_integration_steps,
+            )
+            return (
+                ChEESState(new_state, state.current_iteration + 1),
+                info,
+            )
+
+        return AdaptationResults(last_states, parameters), kernel, info
```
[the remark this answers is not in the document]
junpenglao: Hmm this breaks the contract we have for Adaptation, we should instead write a small function that initialized the kernel using the return information

junpenglao (commented):
Sending out first round of review, I will need a bit more time on chees_adaptation.py

blackjax/mcmc/trajectory.py
```diff
@@ -119,10 +119,18 @@ def integrate(
...
         )
 
-        def one_step(_, state):
-            return integrator(state, directed_step_size)
+        def one_step(state_iter):
+            state, iter = state_iter
+            state = integrator(state, directed_step_size)
+            return (state, iter + 1)
+
+        last_state, _ = jax.lax.while_loop(
```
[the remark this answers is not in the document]
junpenglao: Actually, why is this change needed?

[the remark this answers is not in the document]
albcab: Not needed! It was only a problem with the `scan`.

blackjax/mcmc/hmc.py
```diff
@@ -210,17 +211,20 @@ def kernel(
...
+        )
         hmc_state = HMCState(state.position, state.logdensity, state.logdensity_grad)
         hmc_proposal, info = hmc_base(
             rng_key,
             hmc_state,
             logdensity_fn,
             step_size,
             inverse_mass_matrix,
-            num_integration_steps,
+            jax.numpy.asarray(num_integration_steps, dtype=int),
```
junpenglao, line 227: I think it is better to emphasize `integration_steps_fn` should return `int`. Could you reverse this and add in the docstring instead?

junpenglao:
Could you add a test to check that it learns a good step size and trajectory_length_moving_average?
Similar to https://github.com/tensorflow/probability/blob/bc2e0e0d0969d2b5e7cb0a8fbaaf93ba1e0adfb2/tensorflow_probability/python/experimental/mcmc/gradient_based_trajectory_length_adaptation_test.py#L334

blackjax/adaptation/chees_adaptation.py
```diff
@@ -0,0 +1,426 @@
...
+        *,
+        max_sampling_steps: int = 1000,
+    ):
+        num_dim = 0
+        for d in jax.tree_util.tree_leaves(
+            jax.tree_util.tree_map(
+                lambda p: p.reshape(num_chains, -1).shape[1], positions
+            )
+        ):
+            num_dim += d
```
junpenglao, line 337:
```suggestion
        num_dim = blackjax.util.pytree_size(positions) / num_chains
```

blackjax/adaptation/chees_adaptation.py
```diff
@@ -0,0 +1,426 @@
...
+    Returns
+    -------
+    A function that returns the last cross-chain state, a sampling kernel with the
+    tuned parameter values, and all the warm-up states for diagnostics.
+
+    """
+
+    def run(
+        rng_key: PRNGKey,
+        positions: ArrayLikeTree,
```
junpenglao, line 324: given this is top level API, worth adding an assertion to make sure positions have leading axis size = num_chains

blackjax/adaptation/chees_adaptation.py
```diff
@@ -0,0 +1,426 @@
...
+                    state,
+                    logprob_fn,
+                    step_size=adaptation_state.step_size,
+                    inverse_mass_matrix=jnp.ones(num_dim),
+                    trajectory_length_adjusted=adaptation_state.trajectory_length
+                    / adaptation_state.step_size,
+                )
+
+            keys = jax.random.split(rng_key, num_chains)
+            new_states, info = jax.vmap(one_step)(keys, states)
```
junpenglao, line 379: Could you refactor this to:
```python
keys = jax.random.split(rng_key, num_chains)
jax.vmap(step_fn, in_axis=[0, 0, None...])(keys, states, ...)
```

blackjax/adaptation/chees_adaptation.py
```diff
@@ -0,0 +1,426 @@
...
+            trajectory_length_ma,
+            da_state,
+            optim_state,
+            random_generator_arg,
+        ) = initial_adaptation_state
+
+        harmonic_mean = 1.0 / jnp.mean(
+            1.0 / acceptance_probabilities, where=~is_divergent
+        )
+        da_state_ = da_update(da_state, 0.651 - harmonic_mean)
```
junpenglao, line 130: Turn 0.651 into a CONSTANT above with name and comment.

blackjax/adaptation/chees_adaptation.py
```diff
@@ -0,0 +1,426 @@
...
+
+        proposals_matrix = jax.vmap(lambda p: jax.flatten_util.ravel_pytree(p)[0])(
+            proposals_centered
+        )
+        initials_matrix = jax.vmap(lambda p: jax.flatten_util.ravel_pytree(p)[0])(
+            initials_centered
+        )
+        momentums_matrix = jax.vmap(lambda m: jax.flatten_util.ravel_pytree(m)[0])(
+            proposed_momentums
+        )
```
junpenglao, line 161:
```suggestion
        vmap_flatten_op = jax.vmap(lambda p: jax.flatten_util.ravel_pytree(p)[0])
        proposals_matrix = vmap_flatten_op(proposals_centered)
        initials_matrix = vmap_flatten_op(initials_centered)
        momentums_matrix = vmap_flatten_op(proposed_momentums)
```

blackjax/adaptation/chees_adaptation.py
```diff
@@ -0,0 +1,426 @@
...
+        da_state_ = da_update(da_state, 0.651 - harmonic_mean)
+        step_size_ = jnp.exp(da_state_.log_x)
+        new_step_size, new_da_state = jax.lax.cond(
+            jnp.isfinite(step_size_),
+            lambda _: (step_size_, da_state_),
+            lambda _: (step_size, da_state),
+            None,
+        )
+        new_step_size_ma = 0.9 * step_size_ma + 0.1 * new_step_size
```
junpenglao, line 138: The moving average in tfp looks different (same comment applies below): https://github.com/tensorflow/probability/blob/bc2e0e0d0969d2b5e7cb0a8fbaaf93ba1e0adfb2/tensorflow_probability/python/experimental/mcmc/gradient_based_trajectory_length_adaptation.py#L1004-L1010

albcab: They do a couple of things differently, like having an [extra fixed jitter value](https://github.com/tensorflow/probability/blob/bc2e0e0d0969d2b5e7cb0a8fbaaf93ba1e0adfb2/tensorflow_probability/python/experimental/mcmc/gradient_based_trajectory_length_adaptation.py#L786C28-L787C68). 

They update the moving average for the trajectory length using a weighted geometric mean with the weight depending on the iteration number instead of the weighted mean with fixed weights used in [Algorithm 1](http://proceedings.mlr.press/v130/hoffman21a/hoffman21a.pdf).

Also, this script only seems to update the trajectory length; I can't find updates for the step size. In Algorithm 1, authors use the same weighted mean with fixed weights used for the trajectory length. 

The question is if we want to stick to Algo 1 or use more of the methods in tfp. We are already using their unshuffled generation of a [halton sequence](https://github.com/tensorflow/probability/blob/bc2e0e0d0969d2b5e7cb0a8fbaaf93ba1e0adfb2/tensorflow_probability/python/experimental/mcmc/gradient_based_trajectory_length_adaptation.py#L924).

albcab: similar weighted geometric mean for step size: https://github.com/tensorflow/probability/blob/main/tensorflow_probability/python/mcmc/dual_averaging_step_size_adaptation.py#L460C5-L460C27

blackjax/adaptation/chees_adaptation.py
albcab: `vmap` can't handle well kwargs, [see](https://github.com/google/jax/issues/7465). Best I could do was `partial` them out.

albcab:
> Could you add a test to check that it learns a good step size and trajectory_length_moving_average? Similar to https://github.com/tensorflow/probability/blob/bc2e0e0d0969d2b5e7cb0a8fbaaf93ba1e0adfb2/tensorflow_probability/python/experimental/mcmc/gradient_based_trajectory_length_adaptation_test.py#L334

I don't get it; how do we know that an avg step size of 1.5 and an avg number of steps of 15 are the "correct" amounts? Also, they target an acceptance rate of .75 while we do .651, which would change the target step size and number of steps.

blackjax/adaptation/chees_adaptation.py
junpenglao: > They update the moving average for the trajectory length using a weighted geometric mean with the weight depending on the iteration number instead of the weighted mean with fixed weights used in [Algorithm 1](http://proceedings.mlr.press/v130/hoffman21a/hoffman21a.pdf).

I think it make sense to use the weighted geometric mean with the weight depending on the iteration number (TFP current), as it should be better than a fixed weight

The step size adaptation in TFP is also dual averaging, the implementation here using blackjax dual_averaging is equivalent.

junpenglao: We should also add the [extra fixed jitter value](https://github.com/tensorflow/probability/blob/bc2e0e0d0969d2b5e7cb0a8fbaaf93ba1e0adfb2/tensorflow_probability/python/experimental/mcmc/gradient_based_trajectory_length_adaptation.py#L786C28-L787C68). It's pretty straightfoward to do, and the default value is 1 (no jitter)

junpenglao:
> I don't get it; how do we know that an avg step size of 1.5 and an avg number of steps of 15 are the "correct" amounts? Also, they target an acceptance rate of .75 while we do .651, which would change the target step size and number of steps.

I think those values comes from empirical, we should get similar number with the similar log_density set up. Could you turn  .651 into an kwarg?

albcab:
> I think those values comes from empirical, we should get similar number with the similar log_density set up. Could you turn .651 into an kwarg?

I've added the test, although the number of integration steps depends on the optimizer/learning rate used.

junpenglao (approved):
Great job, thanks for working on this!

junpenglao:
> I've added the test, although the number of integration steps depends on the optimizer/learning rate used.

Interesting. Makes sense.

blackjax/adaptation/chees_adaptation.py
```diff
@@ -0,0 +1,455 @@
...
+
+class ChEESAdaptationState(NamedTuple):
+    """State of the ChEES-HMC adaptation scheme.
+
+    step_size
+        Value of the step_size parameter of the HMC algorithm.
+    log_step_size_moving_average
+        Running moving average of the log step_size parameter.
+    trajectory_length
+        Value of the num_integration_steps / step_size parameter of
```
junpenglao, line 30: Should be
```suggestion
        Value of the num_integration_steps * step_size parameter of
```
right?

junpenglao (approved):

twiecki:
Amazing! 🥳

Issue #328, opened 2022-09-21
Add "How to implement Metropolis-within-Gibbs?" document


rlouf:
See the following discussion https://github.com/blackjax-devs/blackjax/discussions/275#discussioncomment-3665608 for a template.

PR #422, opened 2022-12-13
Add documentation for how to implement Metropolis-within-Gibbs

mlysy:
Addresses #328 based on #275.

I encountered an issue rendering some of the math using `\boldsymbol{}`.  A similar issue occurs in the first line of `examples/GP_EllipticalSliceSampler.md`.  In that case you can get rid of it by replacing `$ \mathbf{y}$` by `$\mathbf{y}$`.  In my case unfortunately that didn't work.  So instead I creates macros for the bold symbols using `\newcommand{}` in the first inline math statement (putting it in an otherwise empty math-display adds white space).


rlouf:
Great! Since it's a common use case I think it should be part of the `how_to_X` tutorials.

examples/MetropolisWithinGibbs.md
```diff
@@ -0,0 +1,208 @@
...
+$$
+(\xx, \yy) \sim \operatorname{Normal}(\boldsymbol{0}, \boldsymbol{\Sigma}).
+$$
+
+The corresponding log-probability function is implemented below.
+
+```{code-cell} ipython3
+def logprob_fn(x, y, Sigma):
+    """
+    Log-pdf of ``(x, y) ~ Normal(0, Sigma)``.
```
rlouf, line 43:
```suggestion
    Log-pdf of ``(x, y) ~ MvNormal(0, Sigma)``.
```

examples/MetropolisWithinGibbs.md
```diff
@@ -0,0 +1,208 @@
...
+
+import blackjax
+```
+
+## The Model
+
+Suppose that $(\xx, \yy)$ are drawn from a multivariate normal distribution
+
+$$
+(\xx, \yy) \sim \operatorname{Normal}(\boldsymbol{0}, \boldsymbol{\Sigma}).
```
rlouf, line 35:
```suggestion
(\xx, \yy) \sim \operatorname{MvNormal}(\boldsymbol{0}, \boldsymbol{\Sigma}).
```

examples/MetropolisWithinGibbs.md
```diff
@@ -0,0 +1,208 @@
...
+
+Two possible ways of doing this manual update are:
+
+1. Use the `blackjax.mcmc.algorithm.init()` function of the given component's MCMC algorithm to update the `AlgorithmState`.  This function has arguments `position` and `logprob_fn`.  In the case of the HMC component after obtaining $\xx_t$ but before drawing $\yy_t$, the `position` would be $\yy_{t-1}$ and the `logprob_fn` function would be $\log p(\xx_t, \cdot )$.
+
+2. The method above should work out-of-the-box for most (if not all) MCMC algorithms in BlackJAX.  However, it is not optimally efficient.  For example for the RMH update, after obtaining $\yy_{t-1}$ but before drawing $\xx_t$, the method above would calculate `RMHState.log_probability` to be $\log p(\xx_{t-1}, \yy_{t-1})$.  But we've already calculated this value from the previous HMC update of $\yy_{t-1} \sim p(\yy \mid \xx_{t-1})$.  So, we can save ourselves the cost of calculating the log-probability twice, at the expense of a deeper understanding of the low-level components of the algorithms at hand.
```
rlouf, line 73: I'm wondering if we should just do (1) and mentions this at the end? (1) always works and is simple, so it may be easier for people to grok the example if we only introduce one thing at a time.

examples/MetropolisWithinGibbs.md
```diff
@@ -0,0 +1,208 @@
...
+```{code-cell} ipython3
+from blackjax.mcmc.hmc import HMCState
+from blackjax.mcmc.rmh import RMHState
+
+# MCMC updaters for each algorithm
+rmh_step_fn = blackjax.mcmc.rmh.kernel()
+hmc_step_fn = blackjax.mcmc.hmc.kernel()  # default integrator, etc.
+
+
+def mwg_kernel(rng_key, state_x, state_y, parameters_x, parameters_y):
```
rlouf, line 88: From an interface perspective, I think it would make sense to pass one big `state` and parameters individually, so that this looks like the other lower-level kernels in blackjax:

```python
def mwg_kernel(rng_key, state, sigma, step_size, inverse_mass_matrix, num_integration_steps)
```

where

```python
state = {'rmh': ..., 'hmc':}
```

what do you think?

mlysy: Makes sense for `state`.  For parameters, if we have two `rmh` components we would need to have arguments`sigma_x` and `sigma_y`.  But I guess that's not a great reason to break form with the lower-level BlackJAX kernels.

Perhaps a better reason to have `state` and `parameters` both of the form 
```python
state = {'x': ..., 'y': ...}
parameters = {'x': dict_x, 'y': dict_y}
```
is that with more than two sub-kernels in the Gibbs sampler, this would make it easier to cycle through them in a loop, as opposed to manually copy-pasting as I did in the two-kernel example.  Then again, doing this here would be needlessly complicated for people referring to the "how to" for a simple example to build from.  So maybe this can be flagged for later, in case it is decided to add built-in Gibbs sampling functionality to BlackJAX. 

OK I'll go ahead and make the changes.

rlouf: Makes sense! With that you're just one step away from a general M-w-G implementation 😄

rlouf:
I saw you changed the title, you can also rename the file `howto_metropolis_within_gibbs.md`. Do you need help with anything?

mlysy:
Hi, sorry my day job is taking up most of my waking hours lately.  I'm trying to set aside a few hours for this in the next couple of days.

Just to confirm, you're OK with both `state` and `parameters` being passed as dicts?

rlouf:
No problem at all! Yes, having both `state` and `parameters` being passed as dicts is a great idea.

mlysy:
OK I think I've addressed the review comments.  For some reason though some of the workflows were failing.  As far as I could tell it was due to the build on readthedocs timing out, but if it's something else please let me know.

rlouf:
Looks like the CI is not failing when it should; it was unable to build your example. I’ll take a closer look over the next couple of days.

rlouf:
It was my mistake. Great work, thank you so much for the addition 🙏

mlysy:
So glad to be part of this! :slightly_smiling_face:

PR #429, opened 2022-12-22
Gitlint configuration

### LINKED ISSUE #420: Consider adding `Gitlint` to the pre-commit hooks to check the commit messages
https://github.com/jorisroovers/gitlint should be configured to make sure commit messages follow [these guidelines](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)

valentynbez:
Configuration of a gitlint according to the rules described in https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html.

Closes #420

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,8 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
+      with:
+        python-version: 3.9
     - uses: pre-commit/action@v2.0.0
 
   test:
--- a/.gitlint
+++ b/.gitlint
@@ -1,0 +1,20 @@
+[title-max-length]
+line-length=120
+
+[title-min-length]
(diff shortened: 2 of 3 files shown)


rlouf:
Thank you for contributing! Before merging we should define what rules we want to enforce and simplify the config file accordingly.

valentynbez:
OK, I tried to follow the post, but it was not so clear. So I added the ones I think are appropriate and left a verbose description of the parameters.

rlouf:
I think we can be a bit more flex on the title line length for instance ; it shouldn't get in the way of people wanting to write good and descriptive titles. It should however be at least 10 characters long, and we certainly want to exclude "WIP".

The perfect linter does not exist, but we want to catch all the errors that are easily caught (capitalisation, length, some keywords, etc.)

rlouf:
Thank you for contributing! Merging.

PR #444, opened 2023-01-12
Add link to Blackjax Matrix room

rlouf:
This is to open a discussion on whether we should have a chat room on Matrix on which users can reach out to us more directly. @junpenglao what's your experience with PyMC wrt chat/discussions/etc?


rlouf:
It was decided in https://github.com/blackjax-devs/blackjax/issues/441 to work with GitHub discussions for now.

PR #478, opened 2023-01-26
Use AutoAPI for the API documentation

rlouf:
In this PR I use AutoAPI to build a complete API documentation for Blackjax. We may still need to provide some higher-level information regarding how to use the library, but this is already pretty good out-of-the-box.

My only concern is related to the use of closures in Blackjax, and whether it is possible to write documentation for nested function automatically. We may want to forego the use of closures altogether in the library?


junpenglao (dismissed):
Nice.

rlouf:
What's your take on potentially removing the closures and just have (bigger) step functions?

junpenglao:
> We may want to forego the use of closures altogether in the library?

Closure is pretty huge feature for the functional style of the library, so I would discourage changing this part for having easier doc generation.

blackjax/mcmc/integrators.py
```diff
@@ -49,16 +49,17 @@ def velocity_verlet(
 ) -> EuclideanIntegrator:
     """The velocity Verlet (or Verlet-Störmer) integrator.
 
-    The velocity Verlet is a two-stage palindromic integrator [1]_ of the form
+    The velocity Verlet is a two-stage palindromic integrator [BouRabee2018]_ of the form
```
junpenglao, line 52: From here on it start to differ than above in terms of reference style (`[1]`), we should keep it consistent.

junpenglao: Just saw https://github.com/blackjax-devs/blackjax/issues/493, so we should use explicit ref everywhere right?

rlouf: I think we should, it will make #473 easier. I would even change the citation style to follow what Sphinx recommends.

rlouf: Let's uniformize all this when solving #493.

PR #479, opened 2023-01-29
Implement Fullrank vi

xidulu:
Fullrank VI

TODO 1: Better test cases that verify posterior covariance recovery

TODO 2: Currently the user don't have access to the _real_to_vector function therefore cannot convert the unnormalized real space vector to the covariance matrix. This is not sensible and should be fixed.


blackjax/vi/fullrank_vi.py
```diff
@@ -0,0 +1,148 @@
...
+import jax.scipy as jsp
+from jax.flatten_util import ravel_pytree
+from optax import GradientTransformation, OptState
+
+from blackjax.types import PRNGKey, PyTree, Array
+
+__all__ = ["FullrankVIState", "FullrankVIInfo", "sample", "generate_fullrank_logdensity", "step"]
+
+
+def _real_vector_to_cholesky(X):
```
junpenglao, line 29: Is this jitable? I think we might need to make `m` and `n` as kwarg, and create a closure below when we are setting up the parameters.

blackjax/vi/fullrank_vi.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def init(
+    position: PyTree,
+    optimizer: GradientTransformation,
+    *optimizer_args,
+    **optimizer_kwargs
+) -> FullrankVIState:
+    """Initialize the fullrank VI state."""
+    mu = jax.tree_map(jnp.zeros_like, position) # Is this a good init strategy?
```
junpenglao, line 57: We should also allow random initialization for both mu and L, maybe allowing user to pass a callable that takes random_key and shape as input (for zeros and ones we can just ignore the random_key)

tests/test_fullrank_vi.py
```diff
@@ -0,0 +1,48 @@
...
+        num_samples = 500
+
+        optimizer = optax.sgd(1e-2)
+        fullrank_vi = blackjax.fullrank_vi(logdensity_fn, optimizer, num_samples)
+        state = fullrank_vi.init(initial_position)
+
+        rng_key = self.key
+        for _ in range(num_steps):
+            rng_key, _ = jax.random.split(rng_key)
+            state, _ = jax.jit(fullrank_vi.step)(rng_key, state)
```
junpenglao, line 41: Use chex functionarity instead of jitting by hand here.

blackjax/vi/fullrank_vi.py
xidulu: Sounds good, shall we also make the changes to MFVI?

blackjax/vi/fullrank_vi.py
xidulu: I am not sure if this is jitable but jax seems to be happy with it. I didn't see any warning.

blackjax/vi/fullrank_vi.py
rlouf: I think we need to make a choice and let users initialise manually if they want something different.

xidulu: In fact, the init function here only serves as a "default initialization option", if the users want to specify their own strategy, they could replace the call to "fullrank_vi.init()" https://github.com/blackjax-devs/blackjax/pull/479/files#diff-8923a0a4ea42b4d3c2e1756e182a25f482809bc5b1c23a601423f5908f0f03e2R36 with their own function

tests/test_fullrank_vi.py
xidulu: Could you give an example of how to achieve that? I am not familiar with chex's usage

junpenglao: See https://github.com/deepmind/chex#test-variants-variantspy, you can find some example in our test eg https://github.com/blackjax-devs/blackjax/blob/4fa296fd2334991d4aadb8f15096e530449bfdb3/tests/test_integrators.py#L114

xidulu: @junpenglao Done

xidulu:
@rlouf @junpenglao 
Any suggestions on how to let the user have access to the covariance matrix? Currently, the user can only access the flattened unnormalized version in the state https://github.com/blackjax-devs/blackjax/pull/479/files#diff-d7b1bbb7765064655362418341e99f42d8f9d15d120f946da70f32ee4295abe5R42
And the conversion from `L` to the covariance matrix is non-trivial....

rlouf:
Sorry for my late reaction. I think that the issue that you're facing with the covariance matrix is part of a more general discussion we're having at the library level (we have similar issues with mass matrix for the HMC algorithms). I suggest we leave it as is for now to get the ball rolling on this PR and get it in a mergeable state.

gil2rok:
Is there still an interest in full rank VI?

junpenglao:
> Is there still an interest in full rank VI?

Yes - i think we are nearly there but maybe it is better to start a new PR afresh

junpenglao:
Close in favor of #720

Issue #482, opened 2023-02-06
Implement Textbook Metropolis Hastings


ciguaran:
### Presentation of the new sampler

Our must basic variant of RMH involves mutating the state of the chain by adding some normal noise on each step (this is typically called random walk, being a subcase of RMH). We are not exposing in kernels.py a proper RMH.

(a) Follow the suggestion from #162  and rename RMH to Random Walk Metropolis
(b) expose RMH in kernels.py 

### Where does it fit in blackjax
Although RMH is basic having it explicitly in the API could foster academic adoption.

Issue #162, opened 2022-01-24
s/rmh.py/random_walk.py


rlouf:

rlouf:
The idea behind this issue is to be able to use Radford Neal's non reversible update of the value to which we compare the acceptance probability ([paper](https://arxiv.org/abs/2001.11950)).

This will require some thought to apply everywhere throughout the `mcmc` module, especially with the progressive sampling schemes. But maybe we don't need to do this yet, applying it to vanilla HMC/GHMC would be a good start.

rlouf:
This is not going to be a breaking change so we can drop this for v1.

Issue #280, opened 2022-09-16
 Names of lower-level implementations are inconsistent 


rlouf:
e.g. `step_fn`, `kernel`, etc. We should also be more explicit names like `build_kernel` of `build_step_fn`. Indeed 

```python
blackjax.mcmc.nuts.kernel()
```

is confusing and should be replaced with

```python
blackjax.mcmc.nuts.build_kernel()
```

And the name should be propagated downstream to aliases in the high level kernel:

```python
blackjax.nuts.build_kernel()
```

rlouf:
We first need to better define our vocabulary. What do we call a kernel, a step function, etc.

albcab:
For MCMC/SMC/SGMCMC: I like `kernel` because it references Markov or general transition kernels, `build_kernel` or `kernel_gn` are good options to avoid confusion. Instead of `one_step` what about `iterator`? Also `init_fn` and `step_fn` are good names.

rlouf:
Let's avoid `iterator` as it has a well-defined meaning in Python already. My main concern when it comes to naming is the distinction between the "elementary" step functions that take the log-density as well as algorithm parameters, and the specialized step function that only take the PRNG key and current position.

albcab:
What about the following naming changes?
`kernel` -> `build_kernel`
`one_step` -> `kernel`
`step` -> `kernel`
`init_fn` -> `init_fn`
`step_fn` -> `step_fn`

rlouf:
This all seems reasonable to me. @junpenglao ?

PR #484, opened 2023-02-10
Exposing RMH and Random Walk as two different algorithms, generalizing RW to non-gaussian jumps

ciguaran:
1. Exposes RMH as an isolated kernel, being able to take in a proposal distribution
2. Renames to make explicit the assumption of 4 MH variants.
3. Generalizes additive step RW to non-gaussian steps
4. Adds the gaussian case as a convenience factory method.
5. Makes RMH depend on the proposal.py abstraction.
6. Adds test for several components.

# Thank you for opening a PR!

 - [Y] *If I add a new sampler*, there is an issue discussing it already;
 This is related to #162 and #482 
 - [Y] We should be able to understand what the PR does from its title only;
 - [Y] There is a high-level description of the changes;
 - [Y] There are links to *all* the relevant issues, discussions and PRs;
 - [Y] The branch is rebased on the latest `main` commit;
 - [Y] Commit messages follow these [guidelines](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html);
 - [Y] The code respects the current naming conventions;
 - [Y] Docstrings follow the [numpy style guide](https://numpydoc.readthedocs.io/en/latest/format.html)
 - [Y] `pre-commit` is installed and configured on your machine, and you ran it before opening the PR;
 - [Y] There are tests covering the changes;
 - [Y] The doc is up-to-date;
 - [Y] *If I add a new sampler** I added/updated related [examples](https://github.com/blackjax-devs/blackjax/tree/main/examples)

Q: is testing enough? I am concerned about keeping adding 'sampling' tests. Although that's a broader discussion on testing strategy not specific to this PR.


blackjax/kernels.py
```diff
@@ -899,7 +894,7 @@ def run(rng_key: PRNGKey, positions: PyTree, num_steps: int = 1000):
     return AdaptationAlgorithm(run)  # type: ignore[arg-type]
 
 
-class rmh:
+class random_walk:
```
albcab, line 897: No need to implement this as it is a very basic example of random walk metropolis hastings that the user could easily implement with a more general `random_walk` kernel (see below).

blackjax/mcmc/rmh.py
```diff
@@ -74,35 +73,6 @@ def init(position: PyTree, logdensity_fn: Callable) -> RMHState:
     return RMHState(position, logdensity_fn(position))
 
 
-def kernel():
```
albcab, line 77: This `kernel` function should mimic the `hmc_proposal` function in `hmc.py` (more info below).

blackjax/mcmc/rmh.py
```diff
@@ -74,35 +73,6 @@ def init(position: PyTree, logdensity_fn: Callable) -> RMHState:
...
-
-    """
-
-    def one_step(
-        rng_key: PRNGKey, state: RMHState, logdensity_fn: Callable, sigma: Array
-    ) -> Tuple[RMHState, RMHInfo]:
-
-        move_proposal_generator = normal(sigma)
-
-        def proposal_generator(key_proposal, position):
```
albcab, line 95: `proposal_generator` builds a trajectory (`trajectory.py`) with an integrator (`integrators.py`, in this case the integrator is a proposal conditional on the previous position) to then create a proposal (`proposal.py`).

blackjax/mcmc/rmh.py
```diff
@@ -74,35 +73,6 @@ def init(position: PyTree, logdensity_fn: Callable) -> RMHState:
...
-            move_proposal = move_proposal_generator(key_proposal, position)
-            new_position = jax.tree_util.tree_map(jnp.add, position, move_proposal)
-            return new_position
-
-        kernel = rmh(logdensity_fn, proposal_generator)
```
albcab, line 100: Then the proposal is accepted using `static_binomial_sampling` of `proposal.py`.

blackjax/mcmc/random_walk.py
```diff
@@ -0,0 +1,58 @@
...
+            new_position = jax.tree_util.tree_map(jnp.add, position, move_proposal)
+            return new_position
+
+        kernel = rmh(logdensity_fn, proposal_generator)
+        return kernel(rng_key, state)
+
+    return one_step
+
+
+def normal(sigma: Array) -> Callable:
```
albcab, line 39: No need to do this example of a random walk.

blackjax/mcmc/random_walk.py
```diff
@@ -0,0 +1,58 @@
...
+from jax import numpy as jnp
+
+from blackjax.mcmc.rmh import RMHInfo, RMHState, rmh
+from blackjax.types import Array, PRNGKey, PyTree
+from blackjax.util import generate_gaussian_noise
+
+__all__ = ["kernel", "normal"]
+
+
+def kernel(random_step):
```
albcab, line 13: Everything from `rmh.py` should be moved here (see details below) but in the more general implementation of random walk metropolis hastings (not only gaussian RWMH).

Ideally, we'd use the functions from `proposal.py` and objects from `integrators.IntegratorState` to do what `rmh.py` currently does.

albcab (changes requested):
Thanks for opening this PR!

The idea here is to simplify the API and reuse the functions and objects in `proposals.py` and `integrators.py` to accept or reject a simple proposal using metropolis hastings. It should work for any proposal which is only conditional on the previous position (so not HMC for example, since it depends on a momentum variable as well). Make sure to be clear on what the proposal function needs to look like in the docs. A specific example of this is the gaussian random walk MH, but the API should cover only the general case. 

On the backend, we want something similar to `mcmc.hmc.hmc_proposal` but for random walks. If you need to add to `proposals.py` or `integrators.py` go ahead, but ideally we only use the functions that already exist (and clean up some stuff that does the same thing).

Let me know if anything is unclear or there is something you want to discuss.

You can ignore the tests for now to avoid unnecessary refactoring.

blackjax/mcmc/rmh.py
albcab: Pretty sure the "trajectory" is given always by the user.

blackjax/mcmc/proposal.py
```diff
@@ -44,11 +59,21 @@ class Proposal(NamedTuple):
 def proposal_generator(
     kinetic_energy: Callable, divergence_threshold: float
 ) -> Tuple[Callable, Callable]:
-    def new(state: IntegratorState) -> Proposal:
-        energy = -state.logdensity + kinetic_energy(state.momentum)
-        return Proposal(state, energy, 0.0, -np.inf)
+    return proposal_generator_from_energy(
+        energy=lambda state: -state.logdensity + kinetic_energy(state.momentum),
```
albcab, line 63: I agree with this, we want to generalize `proposal_generator` so that it can do acceptance for any target, not just an (independent) potential and kinetic density pair. However, the function here in `proposal.py` should only be the general one and in `hmc.py` the specific `lambda state: -state.logdensity + kinetic_energy(state.momentum)` target should be defined. 

So, I would keep `proposal_generator` as it was, but with any `energy` instead of a `kinetic_energy`, then we can call the specific `energy` (given by the target and kinetic in the case of HMC) inside each specific algorithm's file.

Then, we would need to somehow introduce also a `proposal_generator` for asymmetric proposals, i.e. the `delta_energy` would depend not only on the energy but also on the proposal density. A good example of how it would work is on the `kernel` function of `mcmc.mala.py`, where the acceptance probability of a proposal given by Langevin dynamics needs to include the transition probability since the proposal is not reversible (so asymmetric). In this case, the user would need to pass a function that calculates this probability (what `transition_probability` in `mala.py` does) which would need to be added to the delta energy. 

It might be possible even to use the same `proposal_generator` with an optional `transition_probability` argument which just outputs 0 when not given, but I'm not sure it will work.

blackjax/mcmc/trajectory.py
```diff
@@ -617,3 +618,18 @@ def update_sum_log_p_accept(inputs):
         return expansion_state, (is_diverging, is_turning)
 
     return expand
+
+
+def stochastic_trajectory(
```
albcab, line 623: I feel like `trajectory.py` can be left as is, it's just a matter of building an `IntegratorState` from `rmh.py` and passing it to the `proposal_generator`.

albcab (commented):
In the end we want to generalize the `proposal_generator` so that it can serve the purpose of a general Metropolis-Hastings accept/reject step (not only a Metropolis accept/reject step as it does now). 

If you prefer not using `IntegratorState` because it has a useless momentum, there is also `DiffusionState` in `mcmc.diffusions.py` without the momentum. As mentioned, the `mala.py` kernel serves as a good example of how algorithms with a general Metropolis-Hastings accept/reject step should work. 

We'll probably need to rethink the naming. Since we've moved from potential energies to log densities, it might be wise to use a better name than `energy` for the `proposal_generator`. But we can leave this for later.

blackjax/mcmc/proposal.py
```diff
@@ -16,8 +16,23 @@
 import jax
 import jax.numpy as jnp
 import numpy as np
+from typing_extensions import Protocol
 
-from blackjax.mcmc.integrators import IntegratorState
+from blackjax.types import PyTree
+
+
+class StateWithPosition(Protocol):
```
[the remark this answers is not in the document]
ciguaran: Check that StateWithPosition is a Protocol. By having it, we can have Interface Segretation Principle on proposal_generator_from_energy and there's no need to create an IntegratorState for RMH since RWSTate satisfies the protocol.

blackjax/mcmc/proposal.py
ciguaran: Hum, recall that the concept of `def hmc_energy(kinetic_energy):
    def energy(state):
        return -state.logdensity + kinetic_energy(state.momentum)
    return energy
` is present not only in hmc, but also in `dynamic_progressive_integration` and `dynamic_recursive_integration` for now I am just extracting that to a function so the fact that is reused is more explicit we can consider further refactoring in the future.

blackjax/mcmc/trajectory.py
ciguaran: Done

blackjax/mcmc/proposal.py
albcab: > proposal_generator before this PR) assummes that `delta_energy = initial_energy - new_energy` and the initial energy is already calculated, aka there's no way to recalculate it based on the landing state.

Ideally, any algorithm that needs to modify the "inital_energy" based on the proposal would do it before passing it to the `update` function of `proposal_generator`.

> Hum, recall that the concept of def hmc_energy(kinetic_energy): def energy(state): return -state.logdensity + kinetic_energy(state.momentum) return energy is present not only in hmc, but also in dynamic_progressive_integration and dynamic_recursive_integration

Good point, *only* HMC and NUTS are currently using `proposal_generator`. IRMH, RMH, MALA and marginal latent gaussian algorithms all use the same MH step, so they should all be using `proposal_generator`.

blackjax/mcmc/trajectory.py
```diff
@@ -44,6 +44,7 @@
 from blackjax.mcmc.integrators import IntegratorState
 from blackjax.mcmc.proposal import (
     Proposal,
+    hmc_energy,
```
albcab, line 47: this should be defined here, not in `proposal.py`.

The rest of changes here seem good to me.

blackjax/mcmc/proposal.py
```diff
@@ -11,13 +11,15 @@
...
+from __future__ import annotations
+
+from typing import Callable, NamedTuple, Protocol
 
 import jax
 import jax.numpy as jnp
 import numpy as np
 
-from blackjax.mcmc.integrators import IntegratorState
+TrajectoryState = NamedTuple
```
albcab, line 22: :+1:

blackjax/mcmc/proposal.py
```diff
@@ -90,6 +91,40 @@ def update(initial_energy: float, state: IntegratorState) -> Tuple[Proposal, boo
...
+    def update(
+        initial_energy: float, state
+    ) -> tuple[Proposal, bool]:
+        """Generate a new proposal from a trajectory state,
+        correcting for asymmetry in the proposal distribution
+        """
+        new_proposal, is_transition_divergent = symmetric_update(initial_energy, state)
+        new_position = new_proposal.state.position
+        previous_position = state.position
+        weight_correction = proposal_logdensity_fn(
```
albcab, line 110: The weight correction can be done directly on the "energy" calculation. Thus, we only need one `proposal_generator` which can do Metropolis, Metropolis Hastings, or any acceptance probability really.

blackjax/mcmc/proposal.py
```diff
@@ -65,26 +66,55 @@ def update(initial_energy: float, state: IntegratorState) -> Tuple[Proposal, boo
...
+
+
+def transition_aware_proposal_generator(
+    initial_energy_fn: Callable,
+    transition_energy_fn: Callable,
+    divergence_threshold: float,
+    proposal_factory=proposal_from_energy_diff,
+) -> tuple[Callable, Callable]:
+    """
+    A proposal generator that takes into account the transition from one state to the
```
albcab, line 106: detailed docs here are also very important.

blackjax/mcmc/proposal.py
```diff
@@ -65,26 +66,55 @@ def update(initial_energy: float, state: IntegratorState) -> Tuple[Proposal, boo
...
+    other in order to calculate energies.
+    """
+
+    def new(state: TrajectoryState) -> Proposal:
+        return Proposal(state, initial_energy_fn(state), 0.0, -np.inf)
+
+    def update(prev_state: float, new_state: TrajectoryState) -> tuple[Proposal, bool]:
```
albcab, line 113: `prev_state` is a `TrajectoryState` not a `float`.

I prefer the naming:
`prev_state` -> `inital_state`
`new_state` -> `state`

blackjax/mcmc/random_walk.py
```diff
@@ -0,0 +1,296 @@
...
+        def proposal_generator(rng_key: PRNGKey, position: PyTree):
+            return proposal_distribution(rng_key)
+
+        kernel = rmh(logdensity_fn, proposal_generator)
+        return kernel(rng_key, state)
+
+    return one_step
+
+
+def metropolis(logdensity_fn: Callable, transition_generator: Callable):
```
albcab, line 196: I know per our conversations we wanted to make available a Metropolis only kernel. But looking at it here it seems unnecessary, good detailed documentation on `rmh` should give the user access to a Metropolis kernel and (unless you make tests for this kernel) it only reduces test coverage.

blackjax/mcmc/trajectory.py
```diff
@@ -36,7 +36,9 @@
 memory by keeping states that will subsequently be discarded.
 
 """
-from typing import Callable, NamedTuple, Tuple
+from __future__ import annotations
```
albcab, line 39: Not sure what is the use for replacing typing's `Tuple` with `tuple` and using `annotations`, but it seems like a change that should be done separately to this PR.

tests/mcmc/test_proposal.py
```diff
@@ -65,3 +74,75 @@ def _check_mean_and_std(expected_mean, expected_std, samples):
             rtol=1e-2,
             atol=1e-1,
         )
+
+
+class TestTransitionAwareProposalGenerator(unittest.TestCase):
```
albcab, line 79: Use `chex.TestCase` as base class for tests. This way we are able to tests important JAX usage like jitting, pmapping and using device memory for GPU usage. 

Same for all new tests here and in `test_random_walk.py`

albcab (changes requested):
Awesome! Almost done, just some minor revisions and tests need to be part of the JAX ecosystem.

albcab (changes requested):
Just some last small changes. Also, please standardize the naming on all algorithms to (#280)
`kernel` -> `build_kernel`
`one_step` -> `kernel`
`step` -> `kernel`
`init_fn` -> `init_fn`
`step_fn` -> `step_fn`
in `random_walk.py` don't use `build_kernel`, keep the kernel builder with the name of the algorithm, standardize for the rest of the names. In `kernels.py` do use `build_kernel` instead of `kernel`.

I feel like the testing without chex are additions for another PR / needs to be discussed separately. If you add another example to `test_sampling.py` that uses an asymmetric proposal the coverage shouldn't change much. Maybe @junpenglao or @rlouf have thoughts on this?

albcab:
@ciguaran all looks great, thanks again! Tests are good, if you have any good ideas on how to restructure the testing more generally it could be useful addition to the library. Please open an issue to discuss if you do.

albcab (approved):

Issue #289, opened 2022-09-17
Port SGMCMCJax to blackjax


rlouf:
As discussed with the author of SGMCMCJax @jeremiecoullon we will be porting SGMCMCJax to blackjax.

## Gradient estimation

- [x] Standard gradient estimator
- [ ] Control variate gradient estimator
- [ ] SVRG gradient estimator

# Integrators

- [x] SgLD
- [ ] pSgLD
- [ ] SgldAdam
- [x] SgHMC
- [ ] baoab
- [ ] sgnht
- [ ] badodab

PR #515, opened 2023-04-07
Add SGNHT

SamDuffield:
Add Stochastic gradient Nosé-Hoover thermostat sampler of [Ding et al](http://people.ee.duke.edu/~lcarin/sgnht-4.pdf), discussed in #289.

I have also rescaled the `alpha`, `beta` parameters for `sghmc` (and the new `sgnht`) to align with the description in [Ma et al](https://proceedings.neurips.cc/paper_files/paper/2015/file/9a4400501febb2a95e79248486a5f6d3-Paper.pdf), which IMO has better interpretability (`beta` is variance of stochastic gradient). But I understand that this could modify existing code so can be reverted if need be. I also refactored `diffusions.sghmc` to read more clearly as an Euler solver.

I also noticed that the `sgld` and `sghmc` kernels return the `step` functions themselves rather than an `MCMCSamplingAlgorithm` (and that this is in disagreement with the docs).

 - [X] We should be able to understand what the PR does from its title only;
 - [X] There is a high-level description of the changes;
 - [X] There are links to *all* the relevant issues, discussions and PRs;
 - [X] The branch is rebased on the latest `main` commit;
 - [X] The code respects the current naming conventions;
 - [X] Docstrings follow the [numpy style guide](https://numpydoc.readthedocs.io/en/latest/format.html)
 - [X] `pre-commit` is installed and configured on your machine, and you ran it before opening the PR;
 - [X] There are tests covering the changes;
 - [X] The doc is up-to-date;


albcab (changes requested):
Thanks for this PR! Sorry for the delay, busy doing some restructuring.

Everything looks good, docs are great, tests are good, and the general structure of the algorithm looks great.

I agree that the parametrization from [Ma et al](https://proceedings.neurips.cc/paper_files/paper/2015/file/9a4400501febb2a95e79248486a5f6d3-Paper.pdf) for SGHMC is best, let's keep the rescaling. 

You'll only need to follow the new structure of the API introduced in #501 and rebase to the latest commit. All you really need to do is move what you put in `kernels.py` to `sgmcmc/snht.py` and use the naming discussed in #280 for your algorithms. Everything else should be rebased to the latest commit history without any conflicts. You probably don't need it, but [here is a basic skeleton](https://github.com/blackjax-devs/blackjax/blob/4b9d9804b8143d7325c72950d94d844430d3f840/docs/developer/sampling_algorithm.py) for the new structure of sampling algorithms. 

Since `sgld` and `sghmc` don't need a state (only the position needs to be carried over to the next iteration) they also don't need an initializer, hence why they don't return a `MCMCSamplingAlgorithm`. For now, we'll just keep it like this. Even though returning the momentum could be useful for debugging (if someone complains we'll change it).

SamDuffield:
I can't believe I was just getting the hang of the BlackJAX structure and then you went and changed it! 😆 Just kidding, big fan of the changes.

Hopefully, I have adopted them correctly, let me know what you think.

As discussed I have included the rescaling of the parameters for SGHMC. I also updated the docs for SGLC and SGHMC to say they return a "step function" rather than an "MCMCSamplingAlgorithm", but not sure if there is a better terminology.

IMO it might be nice for SGLD and SGHMC to return MCMCSamplingAlgorithm (perhaps with a dummy `init` function) so that users can switch between samplers seamlessly.

albcab (changes requested):
Just two easy naming changes and we are ready to merge!

> IMO it might be nice for SGLD and SGHMC to return MCMCSamplingAlgorithm (perhaps with a dummy `init` function) so that users can switch between samplers seamlessly.

This is an excellent point, completely agree. If you can/want to do the changes, go ahead with another PR (leave the doc updates of this PR), else open an issue and we'll take care of it :+1:

albcab (approved):

PR #519, opened 2023-04-13
Drop support of Python 3.7 and install latest developments of optax to avoid failing tests

albcab:
The latest release of JAX raises an error for the use of any deprecated version of `jax.Array`. This creates errors in tests when importing optax (the latest release is from Nov 2022). In the latests developments of the library this is fixed, so we now install these latests developments instead of the latests release.

We could install an earlier version of JAX, instead of the latest developments of optax.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,7 +51,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v1
       with:
-        python-version: 3.7
+        python-version: 3.8
     - name: Give PyPI some time to update the index
       run: sleep 240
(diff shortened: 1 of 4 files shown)


albcab:
Additionally, JAX has [dropped support of Python 3.7 from v0.4.1](https://github.com/google/jax/releases/tag/jax-v0.4.1), optax has done the same in the latests release. So, for the tests to pass in Python 3.7, I'd have to set an upper bound for the version of JAX (and optax) to install or not run tests in 3.7. I've chosen the latter because I imagine we want the latests features published on JAX.

junpenglao (approved):

PR #522, opened 2023-04-13
SMC-MCMC integration test, plus fixes.

ciguaran:
# Thank you for opening a PR!
Some kernel interfaces weren't compatible with SMC. This PR fixes them and adds a test that protects us from regressions that may break the integration between the MCMC kernels and SMC. If we believe a kernel could be used within SMC, then it should be added to this test.

### DIFF
--- a/blackjax/kernels.py
+++ b/blackjax/kernels.py
@@ -1023,7 +1023,7 @@ class additive_step_random_walk:
             return cls.init(position, logdensity_fn)
 
         def step_fn(rng_key: PRNGKey, state):
-            return kernel(rng_key, random_step, state, logdensity_fn)
+            return kernel(rng_key, state, logdensity_fn, random_step)
 
         return MCMCSamplingAlgorithm(init_fn, step_fn)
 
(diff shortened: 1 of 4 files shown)


albcab (commented):
The changes to the random walk algorithms look fine to me. Only `test_kernel_compatibility.py` seems frivolous, we are simply checking that the low level kernel function has a certain order in its inputs: `kernel(key, state, logprob_fn, *kernel_params)`. Feels like more tests simply to check this are unnecessary, @junpenglao @rlouf thoughts?

albcab:
> The point is that before this PR we were able to break the integration between MCMC and SMC, and no test would complain. By adding this test we are including a 'safety net' that alerts the developer that a new change is breaking.

But the developer of a new algorithm would still need to manually add a test that checks for this, its not done automatically. For example NUTS, MALA, MGrad, GHMC, elliptical slice are not checked by this safety net, could still break it.

I see your point though, maybe we can add support (in the test) for NUTS and MALA and use this as a safety net for the popular algorithms.

PR #526, opened 2023-04-20
Proper splitting of PRNG key

howsiyu:


junpenglao (commented):
Thanks! 
Let's follow the official jax style and do:
```python
key, *multiple_subkeys = random.split(key, num=k)
```
where `key` will get pass down to the next `split` while `multiple_subkeys` would be used in random call.

junpenglao (dismissed):

junpenglao (approved):
Thank you!

PR #532, opened 2023-05-08
Consistent naming of low-level function

### LINKED ISSUE #280: Names of lower-level implementations are inconsistent

albcab:
Closes #280

### DIFF
--- a/blackjax/kernels.py
+++ b/blackjax/kernels.py
@@ -89,7 +89,7 @@ class adaptive_tempered_smc:
     """
 
     init = staticmethod(smc.tempered.init)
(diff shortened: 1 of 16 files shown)


junpenglao (approved):
Thanks!

PR #533, opened 2023-05-09
Fix momentum samples for 2D inverse mass matrix

### LINKED ISSUE #531: Incorrect momentum samples generated when inverse mass matrix is 2D ?
### Describe the issue as clearly as possible:

From the docs, it appears that `jax.scipy.linalg.cholesky` returns the upper triangular factor and the call at https://github.com/blackjax-devs/blackjax/blob/6fb0198154570e1a0ecb95431b8040549e42ad21/blackjax/mcmc/metrics.py#L86-L89 doesn't appear to use any extra keywords (meaning `lower=False`). Using this info, the inverse mass matrix can be factored into $M^{-1} = U^TU$, where $U$ is the upper triangular matrix. Next we want to use $U$ in order to generate a sample from $\mathcal{N}(0, M)$ using the formula $U^{-1} z$ where $z$ is a standard normal sample. This is because the actual distribution's covariance ($M$) is factored as $U^{-1}(U^T)^{-1}$. From looking at the implementation, it seems like the call to `cholesky` returns $U$ (since by default the upper triangular matrix is returned). We are then supposed to invert this upper triangular to obtain the actual factor to use for generating a sample from this distribution (using my notation above, this is $U^{-1}$. But the call to `solve_triangular` appears to be solving $U^TX = I$ which returns $(U^T)^{-1}$, instead of $U^{-1}$, because of the lower=True option).

### Steps/code to reproduce the bug:

```python
None
```

### Expected result:

```shell
Correct samples
```

### Error message:

```shell
None
```

### Blackjax/JAX/jaxlib/Python version information:

```python
jaxlib 0.3.22
jax 0.4.1
Python 3.8.5
```

### Context for the issue:

This was inspired by a similar bug in aesara's `aehmc` implementation here: https://github.com/aesara-devs/aehmc/issues/92

It seems the solution is to make sure `solve_triangular` call solves the right equation by removing the `lower=True` option in the call.

Cc @rlo

junpenglao:
# Thank you for opening a PR!

Fix #531

 - [x] The doc is up-to-date;
 - [x] *If I add a new sampler** I added/updated related [examples](https://github.com/blackjax-devs/blackjax/tree/main/examples)

### DIFF
--- a/blackjax/mcmc/metrics.py
+++ b/blackjax/mcmc/metrics.py
@@ -83,11 +83,18 @@ def gaussian_euclidean(
         matmul = jnp.multiply
 
     elif ndim == 2:
-        tril_inv = jscipy.linalg.cholesky(inverse_mass_matrix)
+        # inverse mass matrix can be factored into L*L.T. We want the cholesky
+        # factor (inverse of L.T) of the mass matrix.
+        L = jscipy.linalg.cholesky(inverse_mass_matrix, lower=True)
         identity = jnp.identity(shape[0])
         mass_matrix_sqrt = jscipy.linalg.solve_triangular(
-            tril_inv, identity, lower=True
+            L, identity, lower=True, trans=True
         )
+        # Note that mass_matrix_sqrt is a upper triangular matrix here, with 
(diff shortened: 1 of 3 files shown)


albcab (approved):
Looks good!

PR #538, opened 2023-05-28
Adding Riemannian Manifold HMC

### LINKED ISSUE #283: Add Riemannian HMC
We can port https://github.com/dfm/rmhmc with the author's blessing. It should be relatively straightforward, difficulties may come from the changes needed to support PyTrees instead of flat arrays.

Since the integrator differs between metric systems, we _may_ benefit from having a data structure that encapsulates this knowledge.

```python

class System(NamedTuple):
    metric: Callable
    integrator: Callable
```

dfm:
Close #283.

As suggested by @junpenglao over in https://github.com/dfm/rmhmc/issues/1, I've started the skeleton of a RMHMC implementation.

It doesn't check all the boxes yet, and I'll need to keep working on it, but I wanted to open the draft since I'm about to go on leave and might not get to it soon :D 

There are various open interface question so it's probably not worth reviewing in too much detail as yet!

Refs:
- [Girolami et al.](https://statmodeling.stat.columbia.edu/wp-content/uploads/2010/04/RMHMC_MG_BC_SC_REV_08_04_10.pdf)
- [Brofos & Lederman](https://arxiv.org/abs/2102.07139)

---

 - [x] **ish** *If I add a new sampler*, there is an issue discussing it already;
 - [x] The doc is up-to-date;


blackjax/mcmc/trajectory.py
```diff
@@ -625,6 +625,11 @@ def update_sum_log_p_accept(inputs):
 
 def hmc_energy(kinetic_energy):
     def energy(state):
-        return -state.logdensity + kinetic_energy(state.momentum)
+        # TODO: This is extremely inelegant!
```
junpenglao, line 628: LOL yikes.
Let's change the kinetic_energy implementation to:
```
def kinetic_energy(momentum, position=None):
    ...
```
and for `gaussian_euclidean` we just `del position`

junpenglao:
Thanks @dfm!!! Super excited for this!

blackjax/mcmc/trajectory.py
dfm: This is a very sensible suggestion!! 😄

junpenglao:
> @junpenglao is this something of interest for blackjax devs?

Yes! Especially nice that you already see a path of building it on top of this PR!

blackjax/mcmc/integrators.py
```diff
@@ -246,3 +248,117 @@ def one_step(state: IntegratorState, step_size: float) -> IntegratorState:
...
+    return one_step
+
+
+class FixedPointSolverInfo(NamedTuple):
+    success: bool
+    norm: float
+    iters: int
+
+
+def _solve_fixed_point(
```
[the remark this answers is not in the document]
junpenglao: Cool - feel free to add a TODO here.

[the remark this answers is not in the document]
dfm: I added the option of using a custom solver here and left a comment in the docstring about possible future work.

blackjax/mcmc/metrics.py
```diff
@@ -22,29 +22,80 @@
...
+from jax.scipy import stats as sp_stats
 
 from blackjax.types import Array, PRNGKey, PyTree
 from blackjax.util import generate_gaussian_noise
 
-__all__ = ["gaussian_euclidean"]
+__all__ = ["default_metric", "gaussian_euclidean", "gaussian_riemannian"]
+
+
+class KineticEnergy(Protocol):
```
junpenglao, line 44: Sweeeeeeeeeeeeeeeeeeet ❤️

krzysztofrusek:
> 
> But either way, I'd love to hear any feedback about this implementation at this point, including from @krzysztofrusek to get a sense for if there would be any interface changes that would make your future PR more straightforward!

Great abstraction, it will be super simple for me to use blackjax parts.
I started with a direct port of Matlab code from the paper and now I am splitting it into parts like integrator, and proposal. 

I think I can implement it within your interface.   
All I  need is to pass the Jacobian of the constraint function to a few parts.  Metric abstraction is nice and will serve me as well.

The additional constrained function is needed in 
- the momentum generator, so it will stay in the tangent space
- Logdet to account for projection
- the Hamiltonian - this is tricky because it's neither kinetic nor potential energy. I could add constrain as a part of potential but it would lose its physical interpretation so I need to think about how to handle this case.

dfm:
> the Hamiltonian - this is tricky because it's neither kinetic nor potential energy. I could add constrain as a part of potential but it would lose its physical interpretation so I need to think about how to handle this case.

Good point! In general, I think it might actually be better to abstract all the way to Hamiltonian rather than just the metric because it many cases where you might want to use RMHMC it would actually be more efficient to evaluate the log density and mass matrix together! Let me poke at that a little bit more.

krzysztofrusek:
> > the Hamiltonian - this is tricky because it's neither kinetic nor potential energy. I could add constrain as a part of potential but it would lose its physical interpretation so I need to think about how to handle this case.
> 
> Good point! In general, I think it might actually be better to abstract all the way to Hamiltonian rather than just the metric because it many cases where you might want to use RMHMC it would actually be more efficient to evaluate the log density and mass matrix together! Let me poke at that a little bit more.

While waiting for this PR to be reviewed, I drafter [CHMC](https://github.com/krzysztofrusek/jax_chmc) in a separate repo.
The api is similar to the one from blackjax but slightly simplified, and closer to the original MATLAB [implementation](https://www.cs.toronto.edu/~mbrubake/projects/cmcmc/) by @mbrubake .

dfm:
@junpenglao — I've been on parental leave and I'm just catching up, so I don't have much to add here at the moment. I'm happy to revisit this implementation in the coming months or happy to hand it off to someone who is keen to take it over!

junpenglao:
Thanks Dan! Let me take a look and get back to you.

junpenglao:
@dfm I updated the PR to the main branch, and it is good to go now. Do you have other things you would like to add?

cc @krzysztofrusek

junpenglao (approved):

junpenglao:
Thank you for the contribution @dfm !!!

dfm:
Thanks @junpenglao! Sorry for disappearing!!

PR #539, opened 2023-05-30
Converting scan to fori_loop in static_integration in order to support a variable number of integration steps in HMC sampler

dfm:
It is common practice to jitter the integration lengths HMC sampling for each proposal, but as currently written, BlackJAX's implementation of HMC doesn't support this. This tiny PR adds support for non-static integration lengths by converting the `scan` that currently backs the HMC proposal integration to a `fori_loop` which supports both static and variable numbers of steps. In the case where a static number is passed, `fori_loop` automatically rewrites itself as the `scan` that was previously used so I don't expect this to introduce any changes for the currently supported use case.

I don't believe this has been previously discussed, but I'm happy to chat if anyone has suggestions!

---

 - [ ] *If I add a new sampler*, there is an issue discussing it already;
 - [ ] The doc is up-to-date;

### DIFF
--- a/blackjax/mcmc/trajectory.py
+++ b/blackjax/mcmc/trajectory.py
@@ -115,19 +115,14 @@ def static_integration(
     def integrate(
         initial_state: IntegratorState, step_size, num_integration_steps
     ) -> IntegratorState:
-        directed_step_size = jax.tree_map(
+        directed_step_size = jax.tree_util.tree_map(
             lambda step_size: direction * step_size, step_size
         )
 
-        def one_step(state, _):
-            state = integrator(state, directed_step_size)
-            return state, state
+        def one_step(_, state):
+            return integrator(state, directed_step_size)
 
-        last_state, _ = jax.lax.scan(
-            one_step, initial_state, jnp.arange(num_integration_steps)
-        )
-
-        return last_state
+        return jax.lax.fori_loop(0, num_integration_steps, one_step, initial_state)
 
     return integrate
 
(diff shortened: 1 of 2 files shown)


junpenglao (approved):

junpenglao:
Thanks @dfm!

junpenglao:
BTW, somewhat related: I personally always prefer to write loop as `scan` because we can backprog through it, and there are a few places `blackjax` use `while_loop`. Should we write them all into `fori_loop` so that it gets dispatch depending on static-ness of num_step? Also, would it dispatch to `scan` when num_step is static _and_ we are taking gradient?

dfm:
Thanks!

> there are a few places `blackjax` use `while_loop`. Should we write them all into `fori_loop` so that it gets dispatch depending on static-ness of num_step?

I don't think this will work in general since `while_loop` is significantly more flexible than `scan` and `fori_loop`. In particular, any `while_loop` that has a non-trivial loop condition will be difficult or impossible to port to a `fori_loop`. In general, I don't expect this to be a huge problem since I don't expect that it is very common to need to diff you MCMC proposals, but I'm sure someone wants to!

> Also, would it dispatch to `scan` when num_step is static and we are taking gradient?

Yes it does!

junpenglao:
I see your point - there is a workaround by enforcing the max_num_iter and roll that into `scan` (that's what it is done in `pathfinder`), but I would not put it as a priority for blackjax right now.

Issue #537, opened 2023-05-26
Unify SGMCMC API


SamDuffield:
Currently we have `sgld` and `sghmc` return `step` functions with no `init`, whilst `csgld` and `sgnht` return a `MCMCSamplingAlgorithm` with an `init` function.

It may be desirable to unify this API so that all SGMCMC algorithms have the same `MCMCSamplingAlgorithm` API with an `init` function that may just be a dummy function e.g. `lambda position: position`. This would be desirable so that the user can seamlessly switch between algorithms.

One question is how to format the `init`, i.e. should we enforce a `rng_key` input even though some algorithms (e.g. `csgld`) do not require a random key on `init`

To me, there seems to be four options

1. Leave it as is, each sampler has its own structure and we do not enforce an `init`
2. Enforce an `init` but no structured signature
3. Enforce an API like `init(rng_key: PRNGKey, position: PyTree, alpha: float = 0.01)` where `rng_key` is enforced but not necessarily used. Any additional arguments come after `position`.
4. Swap `position` and `rng_key` so that we can give the default `rng_key=None` for algorithms that do not need it.
E.g. `init(position: PyTree, rng_key: PRNGKey = None, alpha: float = 0.01)`. Problem here is if the algorithm `init` requires additional arguments that are not optional, they would have to come between `position` and `rng_key` which would disturb the _swappable_ API. This would not be a problem for the current `sgmcmc` suite but it could feasibly be for new algorithms.

We should also keep in mind the API for `blackjax.mcmc` and other parts of the package when deciding on this design.

WDYT @albcab ?

junpenglao:
+1 to standardized it with `init` so the top layer API return a `MCMCSamplingAlgorithm` as well (option 4).

albcab:
`MCMCSamplingAlgorithm` currently enforces an `init(position: PyTree)` signature and it should be kept this way. Exceptions like `ghmc` or `sgnht` should add `rng_key` after `position` as an additional argument and use `# type: ignore[arg-type]` to ignore the signature when testing. Any other additional argument can come before or after `rng_key`, but after `position`. 

In the specific case of `sghmc` and `sgnht` with the `alpha` and `beta` parameters: these are parameters of the kernel, so they should be passed in `__new__`. They are ignored now and passed as default, they should be passed as default in `__new__` (possibly after a `*,` because you'd hope that the end user that is modifying these parameters actually knows what they are doing). 

[...]

PR #512, opened 2023-03-24
SVGD implementation

antotocar34:
As discussed in #385.

Done:
- Implementation of the algorithm
- Added very simple (bad) test.

TODO:
- Add an example with GPU where gradients is calculated with pmap.
- How to take mini batch gradients? 
- Come up with a better test.

**Appreciate any comments & feedback!**

 A few important guidelines and requirements before we can merge your PR:

 - [ X ] *If I add a new sampler*, there is an issue discussing it already;
 - [ X ] We should be able to understand what the PR does from its title only;
 - [ X ] There is a high-level description of the changes;
 - [ X ] There are links to *all* the relevant issues, discussions and PRs;
 - [ X ] The branch is rebased on the latest `main` commit;
 - [ ] Commit messages follow these [guidelines](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html);
 - [ ] The code respects the current naming conventions;
 - [ ] Docstrings follow the [numpy style guide](https://numpydoc.readthedocs.io/en/latest/format.html)
 - [ X ] `pre-commit` is installed and configured on your machine, and you ran it before opening the PR;
 - [ ~ ] There are tests covering the changes;
 - [ ] The doc is up-to-date;
 - [ ] *If I add a new sampler** I added/updated related [examples](https://github.com/blackjax-devs/blackjax/tree/main/examples)

PR #540, opened 2023-05-31
Unify SGMCMC init structure

SamDuffield:
As discussed in #537 , this PR unifies the `sgmcmc` module such that all algorithms have an `init` function and return a `MCMCSamplingAlgorithm`, in the case of `sgld` and `sghmc` the init function is the trivial `lambda position: position`.

This does not result in totally seamless switching since some algorithms (`sgnht` and `csgld`) require a `rng_key` whereas others do not. However, the block to seamless switching is not just down to the `init` since some algorithms (`csgld`) also require custom `step_fun` signatures.

I have reordered the `init` signatures such that `position` is first, `rng_key` is second (if required) followed by any other 
algorithm-specific arguments. This included a refactor to `ghmc` and as a results `meads_adaptation`.

I suggest that possibly we should consider a global reordering of `step_fn` with `state` before `rng_key` since it is possible for an algorithm to not require a `rng_key` in their `step_fn`, e.g. SVGD #512

Finally, I unified the `temperature` parameter such that it is given as an input to the `step_fn` with a default value of 1, previously it had a default value of 0.01 for `csgld` which I am not sure was intended.

I also left the `init` out of the `docs` for `sgld` and `sghmc` since they still work without it, i.e. only using `.step`

 - [x] The doc is up-to-date;

### DIFF
--- a/blackjax/sgmcmc/csgld.py
+++ b/blackjax/sgmcmc/csgld.py
@@ -1,3 +1,16 @@
+# Copyright 2020- The Blackjax Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 """Public API for the Contour Stochastic gradient Langevin Dynamics kernel :cite:p:`deng2020contour,deng2022interacting`.
 
 """
@@ -33,7 +46,7 @@ class ContourSGLDState(NamedTuple):
     energy_idx: int
 
 
-def init(position: PyTree, num_partitions=512):
+def init(position: PyTree, num_partitions=512) -> ContourSGLDState:
     energy_pdf = (
         jnp.arange(num_partitions, 0, -1) / jnp.arange(num_partitions, 0, -1).sum()
     )
@@ -204,7 +217,6 @@ class csgld:
         logdensity_estimator: Callable,
         gradient_estimator: Callable,
         zeta: float = 1,
-        temperature: float = 0.01,
         num_partitions: int = 512,
         energy_gap: float = 100,
         min_energy: float = 0,
@@ -220,6 +232,7 @@ class csgld:
             minibatch: PyTree,
             step_size_diff: float,
             step_size_stoch: float,
+            temperature: float = 1.0,
         ):
             return kernel(
                 rng_key,
--- a/blackjax/sgmcmc/csgld.py
+++ b/blackjax/sgmcmc/csgld.py
@@ -228,12 +228,12 @@ class csgld:
 
         def step_fn(
             rng_key: PRNGKey,
-            state,
+            state: ContourSGLDState,
             minibatch: PyTree,
             step_size_diff: float,
             step_size_stoch: float,
             temperature: float = 1.0,
-        ):
+        ) -> ContourSGLDState:
             return kernel(
                 rng_key,
                 state,
--- a/blackjax/sgmcmc/sgnht.py
+++ b/blackjax/sgmcmc/sgnht.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Public API for the Stochastic gradient Nosé-Hoover Thermostat kernel."""
-from typing import Callable, NamedTuple
+from typing import Callable, NamedTuple, Union
 
 import blackjax.sgmcmc.diffusions as diffusions
 from blackjax.base import MCMCSamplingAlgorithm
@@ -40,9 +40,9 @@ class SGNHTState(NamedTuple):
     xi: float
 
 
-def init(rng_key: PRNGKey, position: PyTree, alpha: float = 0.01):
+def init(position: PyTree, rng_key: PRNGKey, xi: float) -> SGNHTState:
     momentum = generate_gaussian_noise(rng_key, position)
-    return SGNHTState(position, momentum, alpha)
+    return SGNHTState(position, momentum, xi)
 
 
 def build_kernel(alpha: float = 0.01, beta: float = 0) -> Callable:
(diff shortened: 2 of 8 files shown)


junpenglao:
Is it possible to not use `rng_key` in `init`? For example, could we replace it with a placeholder value/PyTree? Would replacing the  `rng_key` with a fix key works?

SamDuffield:
> Is it possible to not use `rng_key` in `init`? For example, could we replace it with a placeholder value/PyTree? Would replacing the `rng_key` with a fix key works?

Yep! We could get close to seamless switching by either
1. Enforce a `rng_key` placeholder even for algorithms that do not require an `init`
```
alg = blackjax.sgld
# or e.g. alg = blackjax.sgnht

state = alg.init(position, rng_key1)
state = alg.step(rng_key2, state, minibatch)
```
2. Set a default fixed rng_key 
```
alg = blackjax.sgld
# or e.g. alg = blackjax.sgnht

state = alg.init(position)
state = alg.step(rng_key, state, minibatch)
```

Although both options have drawbacks (1 clutters the API of those algorithms that do not require `rng_key` on `init`, 2 uses a default fixed key which could hide the pseudo-randomness of the algorithm initiation).

Note that this would allow seamless switching for `sgld`, `sghmc` and `sgnht` but not for `csgld` since `csgld` has a different `step` signature with `step_size_diff` and `step_size_stoch`

blackjax/sgmcmc/csgld.py
```diff
@@ -216,11 +228,12 @@ def init_fn(position: PyTree):
 
         def step_fn(
             rng_key: PRNGKey,
-            state,
+            state: ContourSGLDState,
             minibatch: PyTree,
             step_size_diff: float,
             step_size_stoch: float,
-        ):
+            temperature: float = 1.0,
```
albcab, line 235: In Appendix J of https://arxiv.org/pdf/1902.03932.pdf they do some analysis on the robustness of the temperature, concluding that "We find that the optimal temperature is often less than 1. We hypothesize that this result is due to the model misspecification common to neural networks." For CIFAR-10 the optimal temperature is 0.1 and for CIFAR-100 it is 0.01. So I guess this is why they chose it, but not sure what would be the best default value in this case.

blackjax/sgmcmc/sgnht.py
```diff
@@ -127,19 +127,25 @@ class sgnht:
...
+            position: PyTree, rng_key: PRNGKey, init_xi: Union[None, float] = None
+        ):
+            return cls.init(position, rng_key, init_xi or alpha)
+
+        def step_fn(
+            rng_key: PRNGKey,
+            state: SGNHTState,
+            minibatch: PyTree,
+            step_size: float,
+            temperature: float = 1,
```
albcab, line 145: As I understand it, SGNHT controls the temperature with `xi`, so we shouldn't need a temperature other than 1. If this is the case, we should at least add a `*,` before temperature so that the user that changes it knows what she is doing.

albcab (approved):
> I suggest that possibly we should consider a global reordering of step_fn with state before rng_key since it is possible for an algorithm to not require a rng_key in their step_fn, e.g. SVGD https://github.com/blackjax-devs/blackjax/pull/512

Considering #541 this makes sense, some sampling algorithms have deterministic steps. Most sampling algorithms have random steps though, so maybe we keep the default as is for now.

> Although both options have drawbacks (1 clutters the API of those algorithms that do not require rng_key on init, 2 uses a default fixed key which could hide the pseudo-randomness of the algorithm initiation).

Samplers that require a `rng_key` at `init` have some persistent value, like a momentum. The initialization of this value could impact the resulting samples so I think it's important to make this explicit. Because sampling algorithms don't usually have persistent values, I think the default should remain as is.

blackjax/sgmcmc/csgld.py
SamDuffield: I see ok, yes I think for our use case a default value of 1. makes more sense since we do not assume the model context never mind any cold posterior effect, and this is in line with the other `sgmcmc` algorithms

tests/mcmc/test_sampling.py
```diff
@@ -270,7 +270,7 @@ def test_linear_regression_sgld(self):
 
         data_batch = X_data[:100, :]
         init_position = 1.0
-        _ = sgld(rng_key, init_position, data_batch, 1e-3)
+        _ = sgld.step(rng_key, init_position, data_batch, 1e-3)
```
albcab, line 273: Also what about adding `init_position = sgld.init(init_position)` to keep the coverage up?

blackjax/sgmcmc/sgnht.py
SamDuffield: I think this might be one case where the terminology is a bit confusing 😅 , my understanding from [Ding et al](https://proceedings.neurips.cc/paper/2014/file/21fe5b8ba755eeaece7a450849876228-Paper.pdf) is that `xi` automatically modifies the friction coefficient to force the "system temperature" to be near a "target temperature". This target temperature is the one defined above with default value 1, and behaves in the same way as in the other `sgmcmc` algorithms (just scales the logpdf/gradient).

SamDuffield: Actually looking at it now I think the temperature scaling might be incorrect in `diffusions.sghmc` and `diffusions.sgnht`

It should be either simply rescaling the gradient `g / temperature` or equivalently all `stepsize * temperature` right? Atm for `diffusions.sghmc` and `diffusions.sgnht` I think its doing the latter but incorrectly

albcab: I see what you mean. Change was made in #396. It feels like `g / temperature` is the best option here.

SamDuffield: I think I actually prefer the `step_size * temperature` parameterization as it results in a nice transition to optimisation algorithms with `temperature` -> 0
SGLD -> SGD
SGHMC -> SGD with momentum
SGNHT -> something funky/cool/new 👀 

For SGHMC and SGNHT the temperature is applied to the augmented target distribution $\pi(x, v) \propto \exp(- (U(x) + \frac12 v^Tv) / T)$ rather than just the target in $x$

Probably this should be explained in the docs, and the transition to gradient descent(s) might make a nice little example.

All that needs to be changed is `step_size * beta` -> `step_size * temperature * beta` in `sghmc` and `sgnht` - I'll update the PR shortly.

SamDuffield:
Added `sgld.init` and `sghmc.init` to the tests and amended the temperature scaling for `sghmc` and `sgnht`

PR #554, opened 2023-06-20
Bump to Py3.9

junpenglao:


blackjax/adaptation/mass_matrix.py
```diff
@@ -68,7 +68,7 @@ class MassMatrixAdaptationState(NamedTuple):
 
 def mass_matrix_adaptation(
     is_diagonal_matrix: bool = True,
-) -> Tuple[Callable, Callable, Callable]:
+) -> tuple[Callable, Callable, Callable]:
```
rlouf, line 71: Is `Tuple` deprecated?

junpenglao: Nope, but from py3.9 onward we can "use built-in collection types such as `list` and `dict` as generic types instead of importing the corresponding capitalized types (e.g. `List` or `Dict`) from `typing`."

https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections

albcab (approved):

PR #569, opened 2023-09-20
Update dependencies and use new jax random key

junpenglao:
See https://github.com/google/jax/pull/17297/files for background.

- [x] Update typing of PRNG keys type alias to `jax.Array`
- [x] Use new-style RNG keys `jax.random.key(...)` instead of `jax.random.PRNGKey(...)`


junpenglao:
The usage in the documentation is not yet updated.

PR #576, opened 2023-10-23
Remove unused input in scan

junpenglao:
Minor clean up


junpenglao:
Ah so this is not unused, we need it for the progressive bar

PR #579, opened 2023-10-29
Refactor divergence check to each sampler

### LINKED ISSUE #391: Move `divergence_threshold` to postprocessing in the higher-level kernel
The divergence threshold has nothing fundamental to it and is just an arbitrary way to identify problematic transitions. This parameter is propagated all the way down to `blackjax.mcmc.proposals`, and the information `is_divergent` has to be bubbled up from there through `blackjax.mcmc.trajectory` to `blackjax.mcmc.<algorithm>` when we could simply compute it in a post-processing step in `blackjax.kernels` as long as we return the `delta` energy.

This would simplify the code greatly, provide a small performance improvement for those who use the lower-level directly, and make the internals generally easier to play with.

I have considered this before, but stopped short because NUTS uses this information to return early when building trajectories. It could still be computed in `do_keep_integrating` to stop integration early, and we could check whether the integration returned early in `do_keep_expanding` instead. I think that containing the complexity that NUTS entails in NUTS-specific functions is a good thing.

junpenglao:
Fix #391.

The change end up quite small that it do away divergence check in `proposal.py`, and compute divergence only when needed (in `hmc.py` and `trajectory.py`). The functions in `proposal.py` is more general than just HMC that we see pattern like:
```python
new_proposal, _ = generate_proposal(state, end_state)
```
So removing the check would keeps the functions in `proposal.py` simpler and more focus.

However, we cannot push the check the top most level (ie., it need to remain in `trajectory.py`), as the information is used during the trajactory building and it is not easy to refactor it away.

### DIFF
--- a/blackjax/mcmc/hmc.py
+++ b/blackjax/mcmc/hmc.py
@@ -276,7 +276,7 @@ def hmc_proposal(
     """
     build_trajectory = trajectory.static_integration(integrator)
     init_proposal, generate_proposal = proposal.proposal_generator(
-        hmc_energy(kinetic_energy), divergence_threshold
+        hmc_energy(kinetic_energy)
     )
 
     def generate(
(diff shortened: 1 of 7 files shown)


albcab (approved):
Looks good, makes `proposal.py` cleaner!

PR #580, opened 2023-10-30
Dynamic HMC

albcab:
Dynamic HMC kernel builder and state. It is useful when the number of integration steps in HMC is generated pseudo or quasi-randomly.

### DIFF
--- a/blackjax/mcmc/hmc.py
+++ b/blackjax/mcmc/hmc.py
@@ -24,7 +24,7 @@ from blackjax.base import SamplingAlgorithm
 from blackjax.mcmc.trajectory import hmc_energy
 from blackjax.types import Array, ArrayLikeTree, ArrayTree, PRNGKey
 
-__all__ = ["HMCState", "HMCInfo", "init", "build_kernel", "hmc"]
+__all__ = ["HMCState", "DynamicHMCState", "HMCInfo", "init", "build_kernel", "hmc"]
 
 
 class HMCState(NamedTuple):
@@ -39,6 +39,11 @@ class HMCState(NamedTuple):
     position: ArrayTree
     logdensity: float
     logdensity_grad: ArrayTree
+
+
+class DynamicHMCState(NamedTuple):
+    hmc_state: HMCState
+    random_generator_arg: Array
 
 
 class HMCInfo(NamedTuple):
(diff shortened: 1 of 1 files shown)


blackjax/mcmc/hmc.py
```diff
@@ -145,6 +150,58 @@ def kernel(
...
+):
+    """Build a Dynamic HMC kernel where the number of integration steps is chosen randomly.
+
+    Parameters
+    ----------
+    integrator
+        The symplectic integrator to use to integrate the Hamiltonian dynamics.
+    divergence_threshold
+        Value of the difference in energy above which we consider that the transition is divergent.
+
```
junpenglao, line 167: missing docstring for kwarg

blackjax/mcmc/hmc.py
```diff
@@ -41,6 +41,11 @@ class HMCState(NamedTuple):
     logdensity_grad: ArrayTree
 
 
+class DynamicHMCState(NamedTuple):
+    hmc_state: HMCState
+    random_generator_arg: Array
```
junpenglao, line 46: some comment here (mostly to highlight why we type it more general than a PRNGKey)

blackjax/mcmc/hmc.py
```diff
@@ -41,6 +41,18 @@ class HMCState(NamedTuple):
...
 
+class DynamicHMCState(NamedTuple):
+    """State of the dynamic HMC algorithm.
+
+    Adds a utility array for generating a pseudo or quasi-random sequence of
+    number of integration steps.
+
+    """
+
+    hmc_state: HMCState
```
junpenglao, line 52: On second thought, let's unpack this and do
```python
class DynamicHMCState(NamedTuple):
    """State of the HMC algorithm.
    The HMC algorithm takes one position of the chain and returns another
    position. In order to make computations more efficient, we also store
    the current logdensity as well as the current gradient of the logdensity.
    """
    position: ArrayTree
    logdensity: float
    logdensity_grad: ArrayTree
    random_generator_arg: Array
```
The reason being that we dont need to do additional unpack and the output of static and dynamic hmc are mostly the same.

junpenglao (approved):

PR #581, opened 2023-10-31
Adjusts for the fact that IRMH proposal might not be symmetric

ciguaran:
# Thank you for opening a PR!
Adds proposal_logdensity_fn as parameter, so as to modify MH acceptance rule to accommodate for non-symmetry of the proposal distribution.

 - [N] *If I add a new sampler*, there is an issue discussing it already;
 - [Y] We should be able to understand what the PR does from its title only;
 - [Y] There is a high-level description of the changes;
 - [N/A] There are links to *all* the relevant issues, discussions and PRs;
 - [Y] The branch is rebased on the latest `main` commit;
 - [Y] Commit messages follow these [guidelines](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html);
 - [Y] The code respects the current naming conventions;
 - [Y] Docstrings follow the [numpy style guide](https://numpydoc.readthedocs.io/en/latest/format.html)
 - [Y] `pre-commit` is installed and configured on your machine, and you ran it before opening the PR;
 - [N] There are tests covering the changes;
 I haven't added new tests since this is just calling build_rmh from within a factory (IRMH is indeed a factory) and the functionality is already well tested.
 - [Y] The doc is up-to-date;
 - [N] *If I add a new sampler** I added/updated related [examples](https://github.com/blackjax-devs/blackjax/tree/main/examples)

### DIFF
--- a/blackjax/mcmc/random_walk.py
+++ b/blackjax/mcmc/random_walk.py
@@ -278,6 +278,10 @@ def build_irmh() -> Callable:
         proposal_distribution
             A function that, given a PRNGKey, is able to produce a sample in the same
             domain of the target distribution.
+        proposal_logdensity_fn:
+            For non-symmetric proposals, a function that returns the log-density
+            to obtain a given proposal knowing the current state. If it is not
+            provided we assume the proposal is symmetric.
         """
 
         def proposal_generator(rng_key: PRNGKey, position: ArrayTree):
@@ -316,7 +320,10 @@ class irmh:
     proposal_distribution
         A Callable that takes a random number generator and produces a new proposal. The
         proposal is independent of the sampler's current state.
-
+    proposal_logdensity_fn:
+        For non-symmetric proposals, a function that returns the log-density
+        to obtain a given proposal knowing the current state. If it is not
+        provided we assume the proposal is symmetric.
     Returns
     -------
     A ``SamplingAlgorithm``.
@@ -330,6 +337,7 @@ class irmh:
         cls,
         logdensity_fn: Callable,
         proposal_distribution: Callable,
+        proposal_logdensity_fn: Optional[Callable] = None
     ) -> SamplingAlgorithm:
         kernel = cls.build_kernel()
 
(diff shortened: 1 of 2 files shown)


blackjax/mcmc/random_walk.py
```diff
@@ -60,7 +60,7 @@
         new_state, info = step(rng_key, state)
 
 """
-from typing import Callable, NamedTuple, Optional
+from typing import Callable, NamedTuple, Optional, Tuple
```
junpenglao, line 63: reverse as we can type with `tuple` after py3.9

blackjax/mcmc/random_walk.py
```diff
@@ -270,22 +270,27 @@ def kernel(
         state: RWState,
         logdensity_fn: Callable,
         proposal_distribution: Callable,
-    ) -> tuple[RWState, RWInfo]:
+        proposal_logdensity_fn: Optional[Callable] = None,
+    ) -> Tuple[RWState, RWInfo]:
```
junpenglao, line 274:
```suggestion
    ) -> tuple[RWState, RWInfo]:
```

blackjax/mcmc/random_walk.py
```diff
@@ -270,22 +270,27 @@ def kernel(
...
             A function that, given a PRNGKey, is able to produce a sample in the same
             domain of the target distribution.
+        proposal_logdensity_fn:
+            For non-symmetric proposals, a function that returns the log-density
+            to obtain a given proposal knowing the current state. If it is not
+            provided we assume the proposal is symmetric.
         """
 
         def proposal_generator(rng_key: PRNGKey, position: ArrayTree):
-            del position
```
junpenglao, line 284: reverse.

junpenglao:
Thanks, could you add a small test?

blackjax/mcmc/random_walk.py
ciguaran: done, just out of curiosity why is that del needed?

ciguaran:
> Thanks, could you add a small test?

Added the smallest test I could think of, so as to avoid retesting MH acceptance rule, which is already tested in another place.

blackjax/mcmc/random_walk.py
junpenglao: It's not needed, mostly it is styling to signify the input arg is not needed.

junpenglao (approved):

PR #589, opened 2023-11-19
Refactor integrators.py to make it more general.

### LINKED ISSUE #587: Generalizing integrators
### Current behavior

Currently, integrators like `velocity_verlet` use the (euclidean) Hamiltonian form of update for the momentum and position updates. 

### Desired behavior

For some algorithms, it is useful to be able to specify the form of the momentum and position update, but to have the order and size of the steps determined by the integrator. 

For instance, for *any* differential equation of the form:

$$
\frac{d}{dt}f = (\mathcal{O}_1+\mathcal{O}_2)f
$$

the leapfrog operator can be seen as approximating $e^{\epsilon(\mathcal{O}_1 + \mathcal{O}_2)}$ by $e^{\epsilon \mathcal{O}_1/2}e^{\epsilon \mathcal{O}_2}e^{\epsilon \mathcal{O}_1/2}$

(See more here: https://main--zesty-daffodil-572c7d.netlify.app/integrators/)

In a standard Hamiltonian, the forms of $e^{\epsilon \mathcal{O}_2}$ and $e^{\epsilon \mathcal{O}_1}$ are simple, but for other differential equations, they may be more complex. For MCLMC, their forms are given here: https://main--zesty-daffodil-572c7d.netlify.app/choice_of_integrator/#choosing-the-integrator.

For this reason, I propose that we explicitly parametrize all the integrators by the form of the $\mathcal{O}_1$ and $\mathcal{O}_2$ updates. For example:

```python
def leapfrog(O_2, O_1):

  # optionally, something like: O_1 = partial(tree_map(O_1))

  def step(x, u, step_size):

    uu, xx = O_1(step_size * 0.5, u, x)
    uu, xx = O_2(step_size, uu, xx)
    uu, xx = O_1(step_size * 0.5, uu, xx)

    return xx, uu
  
  return step
```

Then by choosing concrete functions for `O_1` and `O_2`, one can instantiate leapfrog for different problems as desired.

(Note: the MCLMC leapfrog roughly resembles the above, but I've made some simplifications, and also made `O_1` and `O_2` return both u (momentum) and x (position), for the sake of uniformity. Also, the logdensity_grad should be passed through, to avoid recalculation. )

junpenglao:
Also add momentum update based on Esh dynamics

Close #587

### DIFF
--- a/blackjax/mcmc/integrators.py
+++ b/blackjax/mcmc/integrators.py
@@ -74,8 +74,8 @@ def generalized_symplectic_integrator(
 
     def one_step(state: IntegratorState, step_size: float):
         position, momentum, _, logdensity_grad = state
-        # auxiliary infomation generated during integration for diagnostics. It is updated
-        # by the momentum_update_fn and position_update_fn at each call
+        # auxiliary infomation generated during integration for diagnostics. It is
+        # updated by the momentum_update_fn and position_update_fn at each call.
         momentum_update_info = None
         position_update_info = None
         for i, coef in enumerate(coefficients[:-1]):
@@ -101,7 +101,7 @@ def generalized_symplectic_integrator(
                     coef,
                     position_update_info,
                 )
-        # Separate the last steps to short circuit the computation of the kinetic_grad
+        # Separate the last steps to short circuit the computation of the kinetic_grad.
         momentum, kinetic_grad, momentum_update_info = momentum_update_fn(
             momentum,
             logdensity_grad,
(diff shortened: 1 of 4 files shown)


blackjax/mcmc/integrators.py
```diff
@@ -184,65 +271,108 @@ def yoshida(
...
 
-        _, logdensity_grad = logdensity_and_grad_fn(position)
-        momentum = jax.tree_util.tree_map(
-            lambda momentum, logdensity_grad: momentum
-            + b2 * step_size * logdensity_grad,
-            momentum,
-            logdensity_grad,
-        )
+# Intergrators with non Euclidean updates
+def esh_dynamics_momentum_update_one_step(
```
junpenglao, line 288: @reubenharry could you check the implementation to see if it is correct? I ended up using `ravel_pytree(logdensity_grad)` because otherwise using the `tree_map` and `tree_reduce` gets complicated (also complicated for dot product).

tests/mcmc/test_integrators.py
```diff
@@ -101,7 +106,13 @@ class IntegratorTest(chex.TestCase):
     @parameterized.parameters(
         itertools.product(
             ["free_fall", "harmonic_oscillator", "planetary_motion"],
-            ["velocity_verlet", "mclachlan", "yoshida"],
+            [
+                "velocity_verlet",
+                "mclachlan",
+                "yoshida",
+                "non_euclidean_leapfrog",
+                "minimal_norm",
```
junpenglao, line 114: They are not passing yet

blackjax/mcmc/integrators.py
```diff
@@ -38,11 +40,154 @@ class IntegratorState(NamedTuple):
...
+    return update
+
+
+def euclidean_momentum_update_fn(kinetic_energy_fn: EuclideanKineticEnergy):
+    kinetic_energy_grad_fn = jax.grad(kinetic_energy_fn)
+
+    def update(
+        momentum: ArrayTree,
+        logdensity_grad: ArrayTree,
+        step_size: float,
```
reubenharry, line 159: Since step_size and coeff are always multiplied together, why not just treat them as a single step_size argument?

blackjax/mcmc/integrators.py
```diff
@@ -68,43 +213,21 @@ def velocity_verlet(
...
-        momentum = jax.tree_util.tree_map(
-            lambda momentum, logdensity_grad: momentum
-            + b1 * step_size * logdensity_grad,
-            momentum,
-            logdensity_grad,
-        )
-
-        return IntegratorState(position, momentum, logdensity, logdensity_grad)
-
+    cofficients = [b1, a2, b1]
```
reubenharry, line 216: I feel that it's (very slightly) cleaner for the coefficients list to see [b1,a2], and for `generalized_symplectic_integrator` to transform it by `coeff = coeff[:-1] + coeff[::,-1]`.

blackjax/mcmc/integrators.py
```diff
@@ -38,11 +40,154 @@ class IntegratorState(NamedTuple):
 Integrator = Callable[[IntegratorState, float], IntegratorState]
 
 
+def generalized_symplectic_integrator(
```
reubenharry, line 43: I believe that the integrator for MCLMC ends up not being symplectic, so perhaps this could just be called `generalized_two_step_integrator` or similar?

blackjax/mcmc/integrators.py
```diff
@@ -115,61 +238,25 @@ def mclachlan(
...
+        momentum_update_fn,
+        position_update_fn,
+        cofficients,
+        format_output_fn=format_euclidean_state_output,
+    )
 
     return one_step
 
 
 def yoshida(
```
reubenharry, line 257: Another minor point, but it seems like a bit of duplicated code between velocity_verlet, mclachlan and yoshida could be removed, since they are all identical except for the choice of coefficients. Why not have just one function `make_euclidean_integrator`, and then set e.g. `velocity_verlet = make_euclidean_integrator([0.5,1,05])`

tests/mcmc/test_integrators.py
reubenharry, line 114: For uniformity, we could rename "minimal_norm" to "non_euclidean_mclachlan", and also add "non_euclidean_yoshida". Jakob (MCLMC author) said that using Yoshida hadn't yet given better results for MCLMC, but that he was in favour of experimenting with it in future.

blackjax/mcmc/integrators.py
junpenglao: Since the coefficients are pretty small (i.e., still fit in one line), I would say prefer to be explicit here.

blackjax/mcmc/integrators.py
junpenglao: Are the integrator for MCLMC volume-preserve and preserves the symplectic structure of phase space?

blackjax/mcmc/integrators.py
junpenglao: Yes good point.

tests/mcmc/test_integrators.py
junpenglao: yep sounds good.

blackjax/mcmc/integrators.py
junpenglao: I am thinking of if we are going to add `sigma` (inverse_mass_matrix) to MCLMC intergrators, we need some way to make sure the `step_size * coef` is general. Doing that operation in the function is more general as you can do e.g., tree_map(fn, step_size, ...) with coef inside the function.

blackjax/mcmc/integrators.py
reubenharry: I believe the integrator for MCLMC is not volume-preserving.

reubenharry: My understanding is that the normal proof of symplecticity relies on the Euclidean form of the update (i.e. $e^{O_1}$ and $e^{O_2}$ are orthogonal matrices), but that this isn't the case for the MCLMC non-Euclidean updates. In some sense this is not unexpected, since the ODE is no longer derived from a Hamiltonian in (1) here: https://arxiv.org/pdf/2303.18221.pdf .

junpenglao: Got you. Will update the name and docstring.

reubenharry:
I'm getting changed results (on a sampling run) when I substitute in the non euclidean mclachlan instead of the previous `minimal_norm` integrator. Have you confirmed that the two implementations yield the same results on the same inputs?

junpenglao:
Yes the two implementation returns identical result: https://colab.research.google.com/drive/1HMnkMM8PNIaaaSAmZyKrZByCd-GKErsN

But to properly test it we still need to add some test:
- an isotropic Gaussian log_density so we could compare the analytic solution with the integrator.
- testing energy preservation (and for the Gaussian case, angular momentum preservation) as invariants under the integrator

PR #592, opened 2023-11-28
Allow passing line search algorithm to pathfinder.

ColCarroll:
# Thank you for opening a PR!

 - [x] The doc is up-to-date;


junpenglao:
Thanks! Could you do `**lbfgs_kwargs` instead?

junpenglao:
Unless those kwargs have different default compare to what currently in jaxopt (which IIRC that's why we have them explicit in the first place)

blackjax/optimizers/lbfgs.py
```diff
@@ -200,12 +205,17 @@ def scan_body(tup, it):
         next_tup = lax.cond(not_converged, lbfgs_one_step, non_op, carry, it)
         return next_tup, next_tup[0][-1]
 
-    solver = jaxopt.LBFGS(fun=fun, maxiter=maxiter, maxls=maxls, history_size=maxcor)
-    value0, grad0 = jax.value_and_grad(fun)(x0)
```
ColCarroll, line 204: this appeared to be a no-op so I got rid of it -- lmk if I should add it back

junpenglao (approved):

PR #598, opened 2023-12-01
Resolve 'Functions to run kernels'

### LINKED ISSUE #591: Functions to run kernels
### Current behavior

To run a kernel, users write their own loop or scan manually. This can make it inconvenient for a user to quickly run an algorithm.

### Desired behavior

A function which runs a kernel, like:

```python
def run_sampling_algorithm(
    sampling_algorithm: SamplingAlgorithm, num_steps: int, initial_val, rng_key
):
    keys = jax.random.split(rng_key, num_steps)
    state = sampling_algorithm.init(initial_val)
    _, info = jax.lax.scan(
        lambda s, k: (sampling_algorithm.step(k, s)), state, keys
    )
    return info
```

could be made available. In addition, functions which both do tuning and subsequently run the kernel could be useful.

On the other hand, I understand that BlackJax is intended as an un-opinionated backend for other PPLs, so perhaps this is overkill.

PaulScemama:
# Thank you for opening a PR!

Closes #591. Implements a `run_inference_algorithm` wrapper function in `blackjax/util.py` for convenience. See discussion in #591 for more details. 

 - [ ] The doc is up-to-date;

### DIFF
--- a/blackjax/util.py
+++ b/blackjax/util.py
@@ -1,13 +1,14 @@
 """Utility functions for BlackJax."""
 from functools import partial
-from typing import Union
+from typing import Tuple, Union
 
 import jax.numpy as jnp
 from jax import jit, lax
 from jax.flatten_util import ravel_pytree
-from jax.random import normal
+from jax.random import normal, split
 from jax.tree_util import tree_leaves
 
+from blackjax.base import Info, SamplingAlgorithm, State, VIAlgorithm
 from blackjax.types import Array, ArrayLikeTree, ArrayTree, PRNGKey
 
 
@@ -136,3 +137,47 @@ def index_pytree(input_pytree: ArrayLikeTree) -> ArrayTree:
     (dim_input,) = flat_input.shape
     array = jnp.arange(dim_input, dtype=flat_input.dtype)
     return unravel_fn(array)
+
+
+def run_inference_algorithm(
+    rng_key: PRNGKey,
+    initial_state_or_position: ArrayLikeTree,
+    inference_algorithm: Union[SamplingAlgorithm, VIAlgorithm],
+    num_steps: int,
+) -> Tuple[State, State, Info]:
+    """Wrapper to run an inference algorithm.
+
+    Parameters
+    ----------
+    rng_key : PRNGKey
+        The random state used by JAX's random numbers generator.
+    initial_state_or_position: ArrayLikeTree
+        The initial state OR the initial position of the inference algorithm. If an initial position
+        is passed in, the function will automatically convert it into an initial state.
+    inference_algorithm : Union[SamplingAlgorithm, VIAlgorithm]
+        One of blackjax's sampling algorithms or variational inference algorithms.
+    num_steps : int
+        Number of learning steps.
+
+    Returns
+    -------
+    Tuple[State, State, Info]
+        1. The final state of the inference algorithm.
+        2. The history of states of the inference algorithm.
+        3. The history of the info of the inference algorithm.
+    """
+    try:
+        initial_state = inference_algorithm.init(initial_state_or_position)
+    except TypeError:
+        # initial_state is already in the right format.
+        initial_state = initial_state_or_position
+
+    keys = split(rng_key, num_steps)
+
+    @jit
+    def one_step(state, rng_key):
+        state, info = inference_algorithm.step(rng_key, state)
+        return state, (state, info)
+
+    final_state, (state_history, info_history) = lax.scan(one_step, initial_state, keys)
+    return final_state, state_history, info_history
--- a/blackjax/util.py
+++ b/blackjax/util.py
@@ -1,6 +1,6 @@
 """Utility functions for BlackJax."""
 from functools import partial
-from typing import Tuple, Union
+from typing import Union
 
 import jax.numpy as jnp
 from jax import jit, lax
@@ -140,11 +140,11 @@ def index_pytree(input_pytree: ArrayLikeTree) -> ArrayTree:
 
 
 def run_inference_algorithm(
-    rng_key: PRNGKey,
-    initial_state_or_position: ArrayLikeTree,
-    inference_algorithm: Union[SamplingAlgorithm, VIAlgorithm],
-    num_steps: int,
-) -> Tuple[State, State, Info]:
+    rng_key,
+    initial_state_or_position,
+    inference_algorithm,
+    num_steps,
+) -> tuple[State, State, Info]:
     """Wrapper to run an inference algorithm.
 
     Parameters
@@ -169,8 +169,9 @@ def run_inference_algorithm(
     try:
         initial_state = inference_algorithm.init(initial_state_or_position)
     except TypeError:
-        # initial_state is already in the right format.
+        # We assume initial_state is already in the right format.
         initial_state = initial_state_or_position
+    initial_state = initial_state_or_position
 
     keys = split(rng_key, num_steps)
 
--- a/blackjax/util.py
+++ b/blackjax/util.py
@@ -8,7 +8,7 @@ from jax.flatten_util import ravel_pytree
 from jax.random import normal, split
 from jax.tree_util import tree_leaves
 
-from blackjax.base import Info, SamplingAlgorithm, State, VIAlgorithm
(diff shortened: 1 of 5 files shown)


junpenglao (changes requested):
Could you revert the change in the quick start example (it is good to demonstrate how the basic implementation works), and apply the function in the tests where similar helper inference loop function are used?

blackjax/util.py
```diff
@@ -1,13 +1,14 @@
 """Utility functions for BlackJax."""
 from functools import partial
-from typing import Union
+from typing import Tuple, Union
```
junpenglao, line 3: Use python `tuple` to typing.

PaulScemama:
@junpenglao @albcab ran into an issue while going through the tests converting `inference_loops` to new [`run_inference_algorithm`](https://github.com/PaulScemama/blackjax/blob/4d20b91b9a31cce60147c2eb17bdd5da294ec56a/blackjax/util.py#L142). This pertains to the discussion in the issue [here](https://github.com/blackjax-devs/blackjax/issues/591#issuecomment-1833839600). 

This is what is happening: [here](https://github.com/blackjax-devs/blackjax/blob/ece686136ec15d4785fec1dad3e54a9070173947/tests/mcmc/test_sampling.py#L587) is a univariate normal test for samplers. This particular case uses `elliptical_slice` and passes in a dummy `logdensity_fn`: `lambda _: 1.0`.  Now consider this self-contained code

```python
import jax
import jax.numpy as jnp

import blackjax

algo = blackjax.elliptical_slice(lambda _: 1.0, **{"cov": jnp.array([2.0**2]), "mean": 1.0})
rng_key = jax.random.PRNGKey(123)

initial_position = 5.0
initial_state = blackjax.elliptical_slice.init(initial_position, lambda _: 1.0)
print(f"Initial State before try|except block {initial_state}")

# THIS IS BASICALLY WHAT `run_inference_algorithm` is doing. 
try:
    initial_state = blackjax.elliptical_slice.init(initial_state, lambda _: 1.0)
except TypeError:
    pass

print(f"Initial State after try|except block: {initial_state}")

'Initial State before try|except block EllipSliceState(position=5.0, logdensity=1.0)'
'Initial State after try|except block: EllipSliceState(position=EllipSliceState(position=5.0, logdensity=1.0), logdensity=1.0)'
```

The comment indicates that the [`run_inference_algorithm`](https://github.com/PaulScemama/blackjax/blob/4d20b91b9a31cce60147c2eb17bdd5da294ec56a/blackjax/util.py#L142) uses a try|except block to decipher if the `initial_position_or_state` argument is an initial position or an initial state. If it is an initial position then the try block will work. If it is an initial state then the try block will fail due TypeError because of the assumption that the initial state object is different from the initial position object (which is a good assumption, indeed every `init` algorithm returns a custom NamedTuple `State` class while position is a PyTree). 

The problem is that the [`init` for `elliptical_slice`](https://github.com/blackjax-devs/blackjax/blob/ece686136ec15d4785fec1dad3e54a9070173947/blackjax/mcmc/elliptical_slice.py#L71C1-L73C49) applies a `logdensity_fn` to the `position` argument. But in the [test](https://github.com/blackjax-devs/blackjax/blob/ece686136ec15d4785fec1dad3e54a9070173947/tests/mcmc/test_sampling.py#L587), a dummy `logdensity_fn` does not take any inputs and returns 1.0 regardless: it is `lambda _: 1.0`. This results in the unexpected and unwanted behavior where we can infinitely call `init` on any argument and it will not error out. 

Sorry if this is a bit disorganized 😅 but basically I want to know what you both @junpenglao @albcab would like to do about this. Should the test not include the dummy logdensity fn? This would be the quickest fix I think. Or do we need to rethink the `run_inference_algorithm`? 

If the latter, I still think this could be a useful option:

```python
def run_inference_algorithm(
    rng_key: PRNGKey,
    inference_algorithm: Union[SamplingAlgorithm, VIAlgorithm],
    num_steps: int,
    initial_position: ArrayLikeTree = None,
    initial_state: ArrayLikeTree = None,
) -> Tuple[State, State, Info]:
    """
    Wrapper to run an inference algorithm.

    Parameters
    ----------
    rng_key : PRNGKey
        The random state used by JAX's random numbers generator.
    inference_algorithm : Union[SamplingAlgorithm, VIAlgorithm]
        One of blackjax's sampling algorithms or variational inference algorithms.
    num_steps : int
        Number of learning steps.
    initial_position : ArrayLikeTree, optional
        The initial position to initialize the state of an inference algorithm,
        by default None. Note that either `initial_position` or `initial_state`
        must be passed in, but not both.
    initial_state: ArrayLikeTree, optional
        The initial state of the inference algorithm, by default None.
        Note that either `initial_position` or `initial_state` must be passed in,
        but not both.

    Returns
    -------
    Tuple[State, State, Info]
        1. The final state of the inference algorithm.
        2. The history of states of the inference algorithm.
        3. The history of the info of the inference algorithm.
    """
    if (initial_position is None) == (initial_state is None):
        raise ValueError(
            "Either `initial_position` or `initial_state` must be specified, but not both."
        )
    if initial_state is None:
        initial_state =  inference_algorithm.init(initial_position)

    keys = jax.random.split(rng_key, num_steps)

    @jax.jit
    def one_step(state, rng_key):
        state, info = inference_algorithm.step(rng_key, state)
        return state, (state, info)

    final_state, (state_history, info_history) = jax.lax.scan(
        one_step, initial_state, keys
    )
    return final_state, state_history, info_history
```

@junpenglao you mentioned that having `initial_position` in the second arg makes vmapping easier, but if we wanted to vmap over initial positions couldn't we just:

```python
jax.vmap(lambda rng_key, initial_position: run_inference_algorithm(rng_key, inference_algorithm, num_steps, initial_position=initial_position)(rng_keys, initial_positions)
```

I am also a beginner so I could be wrong; I am trying to learn so apologies in advance! Thanks for all the help :)

blackjax/util.py
```diff
@@ -136,3 +137,47 @@ def index_pytree(input_pytree: ArrayLikeTree) -> ArrayTree:
...
+    rng_key: PRNGKey,
+    initial_state_or_position: ArrayLikeTree,
+    inference_algorithm: Union[SamplingAlgorithm, VIAlgorithm],
+    num_steps: int,
+) -> Tuple[State, State, Info]:
+    """Wrapper to run an inference algorithm.
+
+    Parameters
+    ----------
+    rng_key : PRNGKey
```
junpenglao, line 152:
```suggestion
    rng_key:
```

Since we already type the input, no need to repeat the type here. 
Same for the rest of the args.

blackjax/util.py
```diff
@@ -136,3 +137,47 @@ def index_pytree(input_pytree: ArrayLikeTree) -> ArrayTree:
...
+    -------
+    Tuple[State, State, Info]
+        1. The final state of the inference algorithm.
+        2. The history of states of the inference algorithm.
+        3. The history of the info of the inference algorithm.
+    """
+    try:
+        initial_state = inference_algorithm.init(initial_state_or_position)
+    except TypeError:
+        # initial_state is already in the right format.
```
junpenglao, line 172:
```suggestion
        # We assume initial_state is already in the right format.
```

blackjax/util.py
PaulScemama: @junpenglao any thoughts on my above comment about the "bug" in the tests due to using a dummy `logdensity_fn`?

junpenglao:
We should modify the test instead, try: `logdensity_fn: lambda x: jnp.ones_like(x)`

PaulScemama:
@junpenglao I think it is ready for review

junpenglao:
Could you also update this one: https://github.com/blackjax-devs/blackjax/blob/3845635fc059a74f5742b96898692580fe750172/tests/adaptation/test_adaptation.py#L68

junpenglao (approved):

junpenglao:
Nice work, thank you!

junpenglao:
Thank you for your contribution @PaulScemama! And congrats on your first PR to Blackjax :-)

Could you also open a PR to change the inference loop usage in [samplingboo](https://github.com/blackjax-devs/sampling-book)?

PR #603, opened 2023-12-03
Refactor proposal.py

### LINKED ISSUE #599: Refactor proposal.py
proposal.py was initially designed for HMC/NUTS, thus the API and documentation (e.g., the `Proposal` object) are using HMC terminology. When other MCMC algorithm like random_walk is calling the function in `proposal.py`, it needs to overload some of the arguments and thus becoming hard to read.
We should refactor proposal so that it provides a general sampling scheme, and additional wrapper on top so HMC could use it easier.

junpenglao:
Close #599.

Reduce the overloading usage of `Proposal`, which initially designed for subclass of HMC algorithms with trajectory building (i.e., NUTS in this case).

cc @AdrienCorenflos per discussion
cc @ciguaran regarding design pattern change in `prospoal.py`


albcab (approved):
NOTE: doing a Metropolis-Hastings accept/reject step used `proposal_generator`/`asymmetric_proposal_generator` + `static_binomial_sampling`, now it uses `safe_energy_diff`/`compute_asymmetric_acceptance_ratio` + `static_binomial_sampling`.

PR #611, opened 2023-12-07
Minor clean up for SMC

junpenglao:
cc @AdrienCorenflos @ciguaran


junpenglao:
> The sphinx doc raises for SMC btw, this is due to the discussion on the structure of the particles: there seem to be blank lines missing for the list and block. I was planning on having a look but if you are at it, maybe worth doing the change at the same time as this.

Where is it btw? https://blackjax-devs.github.io/blackjax/autoapi/blackjax/smc/base/index.html looks fine?

AdrienCorenflos:
This only appears in the stdout when you run the doc.

ciguaran (approved):

Issue #615, opened 2023-12-09
Remove `transform` from MCLMC, and place it in `run_inference_loop`


reubenharry:
### Current behavior

The MCLMC algorithm has a `transform` parameter, which is nothing specific to MCLMC, but just allows the user to specify a projection of the possibly high dimensional position space to a lower dimensional one. This is place inside the kernel, so that at each step, the kernel returns an info object which contains:

```python
MCLMCInfo(
            transformed_position=transform(position),
            logdensity=logdensity,
            energy_change=kinetic_change - logdensity + state.logdensity,
            kinetic_change=kinetic_change * (dim - 1),
        )
```

### Desired behavior

This was done out of memory considerations. But I don't think there's any need for MCLMC to be special in this way. Instead, the `run_inference_loop` function should have some similar functionality, and it should be removed from MCLMC.

PR #623, opened 2023-12-10
Remove transform from MCLMC

reubenharry:
MCLMC had previously carried around `transformed_position` in its info, which is both unnecessary and at odds with the design of blackjax.

Now removed, and tests and tuning appropriately changed.

Close https://github.com/blackjax-devs/blackjax/issues/615


junpenglao (approved):
Nice, thank you!

PR #624, opened 2023-12-10
Clean up of `run_inference_algorithm` docstring.

junpenglao:


reubenharry:
I'm not totally convinced the ` * (dim - 1)` is a bug. Let me think about this for a bit before this gets merged.

blackjax/mcmc/mclmc.py
```diff
@@ -95,7 +93,7 @@ def kernel(
         ), MCLMCInfo(
             logdensity=logdensity,
             energy_change=kinetic_change - logdensity + state.logdensity,
-            kinetic_change=kinetic_change * (dim - 1),
+            kinetic_change=kinetic_change,
```
junpenglao, line 96: @reubenharry I notice a small bug here: [generalized_two_stage_integrator](https://github.com/blackjax-devs/blackjax/blob/e05b5fe30586326dc2f6dcc1f57fcb318d569e4c/blackjax/mcmc/integrators.py#L53) contains a [short circuit for the 2nd step of `O1` when it is last call](https://github.com/blackjax-devs/blackjax/blob/e05b5fe30586326dc2f6dcc1f57fcb318d569e4c/blackjax/mcmc/integrators.py#L131), so it can avoid some computation. In the case of `noneuclidean_*` integrators, it would [scale the `kinetic_change` with `(dim - 1)`](https://github.com/blackjax-devs/blackjax/blob/e05b5fe30586326dc2f6dcc1f57fcb318d569e4c/blackjax/mcmc/integrators.py#L330-L331)

reubenharry: Just to confirm, does this change (testably) make no numerical difference to the output? (Btw, the (d-1) is related, I think, to this section: https://microcanonical-monte-carlo.netlify.app/tutorial/#time-rescaling)

junpenglao: We dont have test currently for this.

junpenglao: Oh I guess because our tests are in 2d, so this change also makes no numerical differences

reubenharry: OK, I'm reluctant to merge this (and especially not part of another PR), before checking carefully what the effect is.

reubenharry: Perhaps we could file it as a bug for now in an issue, and then I can make a test

junpenglao: Sure, that works as well.

PR #627, opened 2023-12-11
Extend function signature for InitFn

### LINKED ISSUE #619: Improve SamplingAlgorithm design for init_fn and step_fn
High level user API is defined as:

https://github.com/blackjax-devs/blackjax/blob/08e0d7521b2c06ba29b94a1c186bdaf8c08e6310/blackjax/base.py#L88

Where initialization and update has a function signature as follow:
https://github.com/blackjax-devs/blackjax/blob/08e0d7521b2c06ba29b94a1c186bdaf8c08e6310/blackjax/base.py#L37
https://github.com/blackjax-devs/blackjax/blob/08e0d7521b2c06ba29b94a1c186bdaf8c08e6310/blackjax/base.py#L67

However, there are a few sampler does not follow these pattern, we should either extend the function signature, or make sure the function signature is applied.

cc @reubenharry

junpenglao:
Close #619.

This PR introduce `rng_key` as optional input to `initFn` protocal.

For sampler like `dynamic_hmc` and `ghmc`, the `init_fn` of the top level API does not follow the old patter, as it needs an rng_key to generate part of the `state`. While it is possible to set a default rng_key in the class `__new__`, we actually wants to keep the rng_key as input to `init_fn` as we usually want to vmap init so that it takes a vector of PRNG_key to initialized parallel chains (see eg: in meads_adaptation https://github.com/blackjax-devs/blackjax/blob/08e0d7521b2c06ba29b94a1c186bdaf8c08e6310/blackjax/adaptation/meads_adaptation.py#L209, https://github.com/blackjax-devs/blackjax/blob/08e0d7521b2c06ba29b94a1c186bdaf8c08e6310/blackjax/adaptation/meads_adaptation.py#L239-L240).

Thus, in this PR we introduce rng_key to the `InitFn`, with some minor refactoring to other top level API follows the same contract and easier to plug into utility sampling function `run_inference_algorithm`

### DIFF
--- a/blackjax/adaptation/mclmc_adaptation.py
+++ b/blackjax/adaptation/mclmc_adaptation.py
@@ -92,7 +92,7 @@ def mclmc_find_L_and_step_size(
         initial_state = MCMCState(position=0, momentum=1)
 
         # Generate a random number generator key
-        rng_key = jax.random.PRNGKey(0)
+        rng_key = jax.random.key(0)
 
         # Find the optimal parameters for the MCLMC algorithm
         final_state, final_params = mclmc_find_L_and_step_size(
--- a/blackjax/base.py
+++ b/blackjax/base.py
@@ -10,7 +10,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import Callable, NamedTuple
+from typing import Callable, NamedTuple, Optional
 
 from typing_extensions import Protocol
 
@@ -34,7 +34,7 @@ class InitFn(Protocol):
 
     """
 
-    def __call__(self, position: Position) -> State:
+    def __call__(self, position: Position, rng_key: Optional[PRNGKey]) -> State:
         """Initialize the algorithm's state.
 
         Parameters
--- a/blackjax/mcmc/barker.py
+++ b/blackjax/mcmc/barker.py
@@ -189,7 +189,8 @@ class barker_proposal:
     ) -> SamplingAlgorithm:
         kernel = cls.build_kernel()
 
-        def init_fn(position: ArrayLikeTree):
+        def init_fn(position: ArrayLikeTree, rng_key=None):
+            del rng_key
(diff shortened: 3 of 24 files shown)


junpenglao:
@AdrienCorenflos for changes in blackjax/mcmc/marginal_latent_gaussian.py
@reubenharry for changes in blackjax/mcmc/mclmc.py

AdrienCorenflos:
Why did the delta parameter get out of the step function? This seems like an unrelated change.

AdrienCorenflos:
The step is mostly used for calibration purposes, but creating a new kernel
is very expensive so you don't want the user to do that every time they
want to change it.

junpenglao:
Could user does this instead (which are in general how Blackjax sees top level API):
```python
from blackjax.mcmc import marginal_latent_gaussian as mgrad

init, kernel = mgrad.init_and_kernel(...)

# does tuning for delta
...

# Actual sampling
algorithm = mgrad.mgrad_gaussian(logdensity_fn, cov, delta)
state = algorithm.init(...)
for ...
    state, info = algorithm.step(rnd_key, state)
```

AdrienCorenflos:
> Could user does this instead (which are in general how Blackjax sees top level API):
> 
> ```python
> from blackjax.mcmc import marginal_latent_gaussian as mgrad
> 
> init, kernel = mgrad.init_and_kernel(...)
> 
> # does tuning for delta
> ...
> 
> # Actual sampling, delta is now fixed.
> algorithm = mgrad.mgrad_gaussian(logdensity_fn, cov, delta)
> state = algorithm.init(...)
> for ...
>     state, info = algorithm.step(rnd_key, state)
> ```

I suppose, but this would be a waste of computational resources. The init_and_kernel does some handling of the covariance matrix (SVD) which is easily the most expensive part. Why would you want to do it again post calibration?

junpenglao:
To maintain the same API for what we consider "top-level". This is done to be easier to compare different sampler. I suppose your concern here is that the 2nd call to `init_and_kernel` is wasteful? Sounds like it is better to refactor out the class altogether so that the result of svd is the input to the top level API `mgrad_gaussian`.

PR #628, opened 2023-12-11
Refactor mgrad_gaussian

junpenglao:


junpenglao:
@AdrienCorenflos What do you think about something like this?

Context: https://github.com/blackjax-devs/blackjax/pull/627#issuecomment-1850116277

AdrienCorenflos:
That should be fine. Maybe add a construction method "from_covariance" on the SVD class?

junpenglao:
Ah I see what you meant, added.

PR #629, opened 2023-12-13
Alignment test

reubenharry:
This PR ensures that the original and blackjax versions of MCLMC align. It was becoming tiresome to make sure both were aligned, since both repos were changing, so I have simply copied the core of the original repo's code into a test file, and written tests for:

- comparison of the momentum update
- comparsion of the integrator (mclachlan, aka minimal_norm)
- comparison of the sampler
- comparison of the adaptation

It all works.

**Note**: I have not made any effort to make the code I transported over compliant with blackjax' software practices. In fact I wanted to change it as little as possible. So flake8 and mypy are ignored for this test file. 

 - [x] The doc is up-to-date;

### DIFF
--- a/blackjax/adaptation/mclmc_adaptation.py
+++ b/blackjax/adaptation/mclmc_adaptation.py
@@ -47,6 +47,7 @@ def mclmc_find_L_and_step_size(
     desired_energy_var=5e-4,
     trust_in_estimate=1.5,
     num_effective_samples=150,
+    split_fn=jax.random.split,
 ):
     """
     Finds the optimal value of the parameters for the MCLMC algorithm.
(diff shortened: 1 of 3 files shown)


reubenharry:
It seems that using jax.config to set to float64 is causing a test issue. Float64 is needed however. Suggestions on what to do?

junpenglao:
Hmmm thanks for the PR, but I dont think this is necessary... I suggest to convert it into a notebook, in case you want to run it in the future. We can put it in a misc/ folder so that it does not ship with other code when we release a new version.

reubenharry:
Yes that sounds sensible. It's less for the sake of testing, and more for my own use, because once the code gets unaligned, it's very hard to sync.

junpenglao:
> Yes that sounds sensible. It's less for the sake of testing, and more for my own use, because once the code gets unaligned, it's very hard to sync.

I dont think we should aim to have always the 2 implementation in sync, there are trade off between different implementation re speed, numerical precision, clarity etc. As long as we have test on the key places to make sure it is the same (e.g., the integrators) it is sufficient.

reubenharry:
I agree in general. The reason I've been so focused on the alignment is mainly for the tuning, which is quite complex, and parts of which I don't yet understand. So I'm not confident I implemented it right unless it exactly reproduces the current results. For the main algorithm, I agree that the it is simple and modular, so just testing the integrator, noise and (maybe) initialization is fine.

junpenglao:
I see, let's work on test specifically for tuning then in this case.

PR #630, opened 2023-12-15
Alignment test

reubenharry:
Reopening this PR, but now the tests are in a `misc`. Note that changes were made to `run_inference_algorithm`, in order to allow manually specification of seeds.

### DIFF
--- a/blackjax/adaptation/mclmc_adaptation.py
+++ b/blackjax/adaptation/mclmc_adaptation.py
@@ -47,6 +47,7 @@ def mclmc_find_L_and_step_size(
     desired_energy_var=5e-4,
     trust_in_estimate=1.5,
     num_effective_samples=150,
+    split_fn=jax.random.split,
 ):
     """
     Finds the optimal value of the parameters for the MCLMC algorithm.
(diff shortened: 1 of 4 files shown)


junpenglao (changes requested):
This is not JAX best practice to use the same seed, so I dont agree with this change. For alignment test purpose, you should make sure the seed splitting process in the old MCLMC implementation is the same as blackjax.

Also, I would rather you to just submit a PR with a single ipynb.

reubenharry:
OK, I can work on that. I agree that the same seed shouldn't be used, but would it be reasonable to let the user manually choose the keys? (instead of providing a fixed seed)?

And why an ipynb, not a py file?

junpenglao:
> OK, I can work on that. I agree that the same seed shouldn't be used, but would it be reasonable to let the user manually choose the keys? (instead of providing a fixed seed)?

I dont see a valid use case for that, user are manually choosing the seed since Jax seeding is stateless, and the seed splitting is completely deterministic.

We do have one similar alignment test in blackjax, in case you want to take a lookt: https://github.com/blackjax-devs/blackjax/blob/7499bfd1a5bb4276ad84390598092c5d1fef8e51/tests/mcmc/test_trajectory.py#L77

it is comparing the iterative trajectory building with the recursive trajectory building. Note how to make sure both algorithm is using the same RNG_key, it is called in a for loop using the inner function.

> And why an ipynb, not a py file?

I think it is easier for the float64 setting

PR #637, opened 2024-02-17
Example showing how to use BlackJAX to sample from the word "BlackJAX"

AdrienCorenflos:
As per the description. You can play around with the actual look of the resulting animation if you wish.
scatter


junpenglao (approved):
haha nice

PR #653, opened 2024-03-30
Fix MALA transition energy

ksnxr:
The current MALA transition energy is implemented incorrectly. In fact, the correct one should be the reverse of the current one.

To see that, note that MALA's log acceptance probability is calculated as `prev_energy - new_energy`, with `new_energy` being the transition energy from `initial_state` to `state`. As such, the correct `new_energy` should be `-log p(state) - log q(initial_state|state)`, but the current implementation implements `-log p(initial_state) - log q(state|initial_state)`; the implementation of RWMH seems correct.

Reference: https://en.wikipedia.org/wiki/Metropolis-adjusted_Langevin_algorithm

### DIFF
--- a/blackjax/mcmc/mala.py
+++ b/blackjax/mcmc/mala.py
@@ -87,7 +87,7 @@ def build_kernel():
         theta_dot = jax.tree_util.tree_reduce(
             operator.add, jax.tree_util.tree_map(lambda x: jnp.sum(x * x), theta)
         )
-        return -state.logdensity + 0.25 * (1.0 / step_size) * theta_dot
+        return state.logdensity - 0.25 * (1.0 / step_size) * theta_dot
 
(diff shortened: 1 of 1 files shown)


junpenglao (approved):
Thank you! And congrats on the first PR to blackjax!

blackjax/mcmc/mala.py
```diff
@@ -80,14 +80,14 @@ def transition_energy(state, new_state, step_size):
         """Transition energy to go from `state` to `new_state`"""
         theta = jax.tree_util.tree_map(
             lambda new_x, x, g: new_x - x - step_size * g,
```
ColCarroll, line 82: Should probably update the variable names here.

ksnxr: Indeed, should probably exchange `new_x` and `x` here I guess

PR #657, opened 2024-04-18
A draft PR on the removal of Algorithm classes.

ciguaran:
This is a super draft PR showcasing how can we remove the algorithm classes. These are static classes we don't use directly, but via calling `new` and getting a `SamplingAlgorithm` result. This PR:

- replaces these classes with instances of a single class so now each module contains: an init, build_kernel and a as_sampling_algorithm functions. The idea is that the latter fixes dependencies/parameters, in particular when these are not differentiable. The init and build_kernel still exist and are exposed since we need that lower level API, specially when composing algorithms (for example, when tuning SMC inner kernel we need to be able to change parameter on every call).

By doing this we still can call algorithms directly, like blackjax.hmc(). What we do loose is the (light) type annotations we are doing, for example in window_adaptation. I have been thinking about this type of annotations, and I think we should remove them. 

The reason for doing it is the following: python fosters duck and structural typing, in contrast to nominal typing like you can find in say Java. In the case of window adaptation, we want the type to mean "hmc family" as algorithms that have an inverse_mass_matrix and a step size. But the way is implemented right now, it actually means whatever the class hmc or the class nuts does! So the classes kind of exist just to be able to name them (aka to use nominal typing). Since most of our codebase is functional, from a typing perspective most samplers are Callables that take in matrixes, doubles, pytrees and return something of the same flavours. There's no way to say: this is the type of a callable that takes in an inverse_mass_matrix and a step size and uses it in some consistent way, because that is not duck typing nor structural! aka we are trying to statically type using tools that are not pythonic. I'd suggest we replace this kind of "algorithm level" type annotations with docs and tests. Check for example the smc_compatibility_test 

PD: I haven't repeated the pattern throughout the codebase so as to illustrate the point with a simple example.

### DIFF
--- a/blackjax/__init__.py
+++ b/blackjax/__init__.py
@@ -60,7 +60,6 @@ __all__ = [
     "rmhmc",
     "mala",
     "mgrad_gaussian",
-    "nuts",
     "orbital_hmc",
     "additive_step_random_walk",
     "rmh",
--- a/blackjax/__init__.py
+++ b/blackjax/__init__.py
@@ -1,3 +1,6 @@
+import dataclasses
+from typing import Callable, Union
+
 from blackjax._version import __version__
 
 from .adaptation.chees_adaptation import chees_adaptation
@@ -5,17 +8,16 @@ from .adaptation.mclmc_adaptation import mclmc_find_L_and_step_size
 from .adaptation.meads_adaptation import meads_adaptation
 from .adaptation.pathfinder_adaptation import pathfinder_adaptation
 from .adaptation.window_adaptation import window_adaptation
+from .base import SamplingAlgorithm
 from .diagnostics import effective_sample_size as ess
 from .diagnostics import potential_scale_reduction as rhat
 from .mcmc.barker import barker_proposal
 from .mcmc.dynamic_hmc import dynamic_hmc
 from .mcmc.elliptical_slice import elliptical_slice
 from .mcmc.ghmc import ghmc
-from .mcmc.hmc import hmc
 from .mcmc.mala import mala
 from .mcmc.marginal_latent_gaussian import mgrad_gaussian
 from .mcmc.mclmc import mclmc
-from .mcmc.nuts import nuts
 from .mcmc.periodic_orbital import orbital_hmc
 from .mcmc.random_walk import additive_step_random_walk, irmh, rmh
 from .mcmc.rmhmc import rmhmc
@@ -31,12 +33,29 @@ from .vi.meanfield_vi import meanfield_vi
 from .vi.pathfinder import pathfinder
 from .vi.schrodinger_follmer import schrodinger_follmer
 from .vi.svgd import svgd
+from .mcmc import hmc as _hmc
+from .mcmc import nuts as _nuts
+
+
+@dataclasses.dataclass
+class SamplingAlgorithmFactory:
+    differentiable_callable: Callable
+    init: Callable
+    build_kernel: Callable
+
+    def __call__(self, *args, **kwargs) -> SamplingAlgorithm:
+        return self.differentiable_callable(*args, **kwargs)
+
+
+hmc = SamplingAlgorithmFactory(_hmc.as_sampling_algorithm, _hmc.init, _hmc.build_kernel)
+nuts = SamplingAlgorithmFactory(_nuts.as_sampling_algorithm, _nuts.init, _nuts.build_kernel)
+
+hmc_family = [hmc, nuts]
 
 __all__ = [
     "__version__",
     "dual_averaging",  # optimizers
     "lbfgs",
-    "hmc",  # mcmc
     "dynamic_hmc",
     "rmhmc",
     "mala",
--- a/blackjax/adaptation/window_adaptation.py
+++ b/blackjax/adaptation/window_adaptation.py
@@ -17,7 +17,6 @@ from typing import Callable, NamedTuple, Union
 import jax
 import jax.numpy as jnp
 
-import blackjax.mcmc as mcmc
 from blackjax.adaptation.base import AdaptationInfo, AdaptationResults
 from blackjax.adaptation.mass_matrix import (
     MassMatrixAdaptationState,
@@ -243,7 +242,7 @@ def base(
 
 
 def window_adaptation(
-    algorithm: Union[mcmc.hmc.hmc, mcmc.nuts.nuts],
+    algorithm,
     logdensity_fn: Callable,
     is_mass_matrix_diagonal: bool = True,
     initial_step_size: float = 1.0,
@@ -252,7 +251,7 @@ def window_adaptation(
     **extra_parameters,
 ) -> AdaptationAlgorithm:
     """Adapt the value of the inverse mass matrix and step size parameters of
-    algorithms in the HMC fmaily.
+    algorithms in the HMC family.  See Blackjax.hmc_family
 
     Algorithms in the HMC family on a euclidean manifold depend on the value of
     at least two parameters: the step size, related to the trajectory
--- a/blackjax/adaptation/pathfinder_adaptation.py
+++ b/blackjax/adaptation/pathfinder_adaptation.py
@@ -138,7 +138,7 @@ def base(
 
 
 def pathfinder_adaptation(
-    algorithm: Union[mcmc.hmc.hmc, mcmc.nuts.nuts],
+    algorithm,
     logdensity_fn: Callable,
     initial_step_size: float = 1.0,
     target_acceptance_rate: float = 0.80,
--- a/blackjax/mcmc/hmc.py
+++ b/blackjax/mcmc/hmc.py
@@ -29,7 +29,7 @@ __all__ = [
     "HMCInfo",
     "init",
     "build_kernel",
-    "hmc",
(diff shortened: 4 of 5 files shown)


blackjax/adaptation/window_adaptation.py
```diff
@@ -243,7 +242,7 @@ def final(warmup_state: WindowAdaptationState) -> tuple[float, Array]:
 
 
 def window_adaptation(
-    algorithm: Union[mcmc.hmc.hmc, mcmc.nuts.nuts],
+    algorithm,
```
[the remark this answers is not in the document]
ciguaran: Yes, but adding that type annotation would tell the user that any subtype of SamplingAlgorithm can be used as parameter to that function, which would not be true.

blackjax/__init__.py
```diff
@@ -1,71 +1,158 @@
...
+"""
+The above three classes exist as a backwards compatible way of exposing both the high level, differentiable
+factory and the low level components, which may not be differentiable. Moreover, this design allows for the lower
+level to be mostly functional programming in nature and reducing boilerplate code.
+"""
+
+
+@dataclasses.dataclass
+class SamplingAlgorithmFactories:
+    differentiable_callable: Callable
```
junpenglao, line 47:
```suggestion
    differentiable: Callable
```
or `differentiable_fn`

blackjax/__init__.py
```diff
@@ -1,71 +1,158 @@
...
+
+"""
+The above three classes exist as a backwards compatible way of exposing both the high level, differentiable
+factory and the low level components, which may not be differentiable. Moreover, this design allows for the lower
+level to be mostly functional programming in nature and reducing boilerplate code.
+"""
+
+
+@dataclasses.dataclass
+class SamplingAlgorithmFactories:
```
junpenglao, line 46:
```suggestion
class GenerateSamplingAPI:
```

blackjax/__init__.py
```diff
@@ -1,71 +1,158 @@
...
+    def __call__(self, *args, **kwargs) -> SamplingAlgorithm:
+        return self.differentiable_callable(*args, **kwargs)
+
+    def register_factory(self, name, callable):
+        setattr(self, name, callable)
+
+
+@dataclasses.dataclass
+class VIAlgorithmFactories:
+    differentiable_callable: Callable
```
junpenglao, line 60:
```suggestion
class GenerateVariationalAPI:
    differentiable: Callable
```

blackjax/__init__.py
```diff
@@ -1,71 +1,158 @@
...
+    step: Callable
+    sample: Callable
+
+    def __call__(self, *args, **kwargs) -> VIAlgorithm:
+        return self.differentiable_callable(*args, **kwargs)
+
+
+@dataclasses.dataclass
+class PathfinderAlgorithmFactories:
+    differentiable_callable: Callable
```
junpenglao, line 71:
```suggestion
class GeneratePathfinderAPI:
    differentiable: Callable
```

blackjax/__init__.py
```diff
@@ -1,71 +1,158 @@
...
+    approximate: Callable
+    sample: Callable
+
+    def __call__(self, *args, **kwargs) -> PathFinderAlgorithm:
+        return self.differentiable_callable(*args, **kwargs)
+
+
+def sampling_factory_from_module(module):
```
junpenglao, line 79:
```suggestion
def generate_top_level_api_from(module):
```

blackjax/mcmc/barker.py
```diff
@@ -128,7 +128,10 @@ def kernel(
     return kernel
 
 
-class barker_proposal:
+def as_sampling_algorithm(
```
junpenglao, line 131:
```suggestion
def as_top_level_api(
```

same for the other files

blackjax/mcmc/random_walk.py
```diff
@@ -297,7 +293,11 @@ def proposal_generator(rng_key: PRNGKey, position: ArrayTree):
     return kernel
 
 
-class irmh:
+def irmh_as_sampling_algorithm(
```
junpenglao, line 296:
```suggestion
def irmh_as_top_level_api(
```

blackjax/mcmc/random_walk.py
```diff
@@ -420,7 +410,11 @@ def kernel(
     return kernel
 
 
-class rmh:
+def rmh_as_sampling_algorithm(
```
junpenglao, line 413:
```suggestion
def rmh_as_top_level_api(
```

blackjax/sgmcmc/csgld.py
```diff
@@ -174,7 +174,14 @@ def kernel(
     return kernel
 
 
-class csgld:
+def as_sampling_algorithm(
```
junpenglao, line 177:
```suggestion
def as_top_level_api(
```

blackjax/vi/meanfield_vi.py
```diff
@@ -109,7 +109,11 @@ def sample(rng_key: PRNGKey, state: MFVIState, num_samples: int = 1):
     return _sample(rng_key, state.mu, state.rho, num_samples)
 
 
-class meanfield_vi:
+def as_vi_algorithm(
```
junpenglao, line 112:
```suggestion
def as_top_level_api(
```

blackjax/vi/pathfinder.py
```diff
@@ -242,7 +242,7 @@ def sample(
         return jax.vmap(unravel_fn)(phi), logq
 
 
-class pathfinder:
+def as_pathfinder_algorithm(logdensity_fn: Callable) -> PathFinderAlgorithm:
```
junpenglao, line 245:
```suggestion
def as_top_level_api(logdensity_fn: Callable) -> PathFinderAlgorithm:
```

tests/smc/test_inner_kernel_tuning.py
```diff
@@ -14,7 +14,7 @@
 from blackjax import adaptive_tempered_smc, tempered_smc
 from blackjax.mcmc.random_walk import build_irmh
 from blackjax.smc import extend_params
-from blackjax.smc.inner_kernel_tuning import inner_kernel_tuning
+from blackjax.smc.inner_kernel_tuning import as_sampling_algorithm
```
junpenglao, line 17:
```suggestion
from blackjax.smc.inner_kernel_tuning import as_sampling_algorithm as inner_kernel_tuning
```

same for the rest of the test

junpenglao (changes requested):
I have some suggestion around naming.
@albcab thoughts?

blackjax/__init__.py
albcab: `GenerateSamplingAlgorithm`

blackjax/__init__.py
albcab: `GenerateVariationalAlgorithm`

blackjax/__init__.py
albcab: `GeneratePathfinderAlgorithm`

blackjax/__init__.py
albcab: `generate_algorithm_from`

blackjax/mcmc/barker.py
albcab: `as_algorithm`

junpenglao: I prefer `as_top_level_api` and use it across all inference method (MCMC, VI, Pathfinder)

albcab:
> I have some suggestion around naming. @albcab thoughts?

Suggesting algorithm rather than API, this way we could use `as_algorithm` for all.

blackjax/mcmc/barker.py
albcab: Lets go with `as_top_level_api` :+1:

blackjax/mcmc/dynamic_hmc.py
```diff
@@ -27,7 +27,6 @@
     "DynamicHMCState",
     "init",
     "build_kernel",
-    "dynamic_hmc",
```
junpenglao, line 30: "as_top_level_api" symbol should be exported here.

junpenglao (approved):

junpenglao (approved):
Great job! Thank you for working on this!

blackjax/__init__.py
```diff
@@ -1,71 +1,156 @@
...
+
+# MCMC
+hmc = generate_top_level_api_from(_hmc)
+nuts = generate_top_level_api_from(_nuts)
+rmh = GenerateSamplingAPI(rmh_as_top_level_api, random_walk.init, random_walk.build_rmh)
+irmh = GenerateSamplingAPI(
+    irmh_as_top_level_api, random_walk.init, random_walk.build_irmh
+)
+dynamic_hmc = generate_top_level_api_from(_dynamic_hmc)
+rmhmc = generate_top_level_api_from(rmhmc)
```
junpenglao, line 93: Last bit of nit: wherever there is a pattern like this, add `_` in the import, eg:
```
from .mcmc import rmhmc as _rmhmc
rmhmc = generate_top_level_api_from(_rmhmc)
...

```

junpenglao (approved):

PR #658, opened 2024-04-18
A draft PR on the removal of Algorithm classes. variant 2

ciguaran:
Same as https://github.com/blackjax-devs/blackjax/pull/657 but avoids creating an intermediate object, at the cost of not being able to call directly blackjax.hmc() and having
to do blackjax.hmc.as_sampling_algorithm(). I don't think this is terrible since we are making the SamplingAlgorithm abstraction more explicit. Since the hmc gets exposed by blackjax.py, it can be passed around (for example in tests).


ciguaran:
Closing this one (see discussion here https://github.com/blackjax-devs/blackjax/issues/659)

Issue #668, opened 2024-05-11
Numerical test `test_chees_adaptation` fails on `aarch64-linux`


GaetanLepage:
### Describe the issue as clearly as possible:

Using `jax`/`jaxlib` 0.4.28 and `jaxopt` 0.8.3, the following test fails on `aarch64-linux`:
```
FAILED tests/adaptation/test_adaptation.py::test_chees_adaptation - AssertionError:
```

### Steps/code to reproduce the bug:

```python
pytest
```

### Expected result:

```shell
Tests pass.
```

### Error message:

```shell
=================================== FAILURES ===================================
____________________________ test_chees_adaptation _____________________________
[gw1] linux -- Python 3.11.9 /nix/store/33752yykc8r75jxvpcvpcynm22il4ch7-python3-3.11.9/bin/python3.11

    def test_chees_adaptation():
        logprob_fn = lambda x: jax.scipy.stats.norm.logpdf(
            x, loc=0.0, scale=jnp.array([1.0, 10.0])
        ).sum()
    
        num_burnin_steps = 1000
        num_results = 500
        num_chains = 16
        step_size = 0.1
    
        init_key, warmup_key, inference_key = jax.random.split(jax.random.key(346), 3)
    
        warmup = blackjax.chees_adaptation(
            logprob_fn, num_chains=num_chains, target_acceptance_rate=0.75
        )
    
        initial_positions = jax.random.normal(init_key, (num_chains, 2))
        (last_states, parameters), warmup_info = warmup.run(
            warmup_key,
            initial_positions,
            step_size=step_size,
            optim=optax.adamw(learning_rate=0.5),
            num_steps=num_burnin_steps,
        )
        algorithm = blackjax.dynamic_hmc(logprob_fn, **parameters)
    
        chain_keys = jax.random.split(inference_key, num_chains)
        _, _, infos = jax.vmap(
            lambda key, state: run_inference_algorithm(key, state, algorithm, num_results)
        )(chain_keys, last_states)
    
        harmonic_mean = 1.0 / jnp.mean(1.0 / infos.acceptance_rate)
>       np.testing.assert_allclose(harmonic_mean, 0.75, rtol=1e-1)

tests/adaptation/test_adaptation.py:69: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

[...]

PR #671, opened 2024-05-13
Make tests pass on `aarch64-linux`

albcab:
Make tests pass on `aarch64-linux` (#668)


junpenglao (approved):

GaetanLepage (approved):
Hey !
Thank you for the quick fix.
I can confirm you that without the patch, the test fails on `aarch64-linux`, and including it fixes the issue.

PR #672, opened 2024-05-13
Update `run_inference_algorithm` to split `initial_position` and `initial_state`

reubenharry:
Improvement to `run_inference_algorithm`

**Description**:

`run_inference_algorithm` currently uses a try-except clause to allow the user to either provide an initial position or an initial state. This has led to some problems, when the `except` clause fails to trigger.

`run_inference_algorithm` also produces an array of `[num_steps, n_dims]`, which for long chains on high-dim problems can be prohibitively large. 

**Solution**:

Make `run_inference_algorithm` take `initial_state` or `initial_position` explicitly, so that the caller of `run_inference_algorithm` is responsible for providing `initial_state` or `initial_position`, rather than deferring this task.

Add a `streaming` mode for `run_inference_algorithm`.

 - [x] The doc is up-to-date;


blackjax/util.py
```diff
@@ -142,12 +142,13 @@ def index_pytree(input_pytree: ArrayLikeTree) -> ArrayTree:
 
 def run_inference_algorithm(
     rng_key: PRNGKey,
-    initial_state_or_position: ArrayLikeTree,
+    initial_state: ArrayLikeTree,
```
junpenglao, line 145: While I agree the initial `try ... except ...` block is not rigor to capture both `initial_state` and `initial_position`, we should not remove it here because an important use case for this top level API is for user to just supply the `initial_position`. Let's do it this way:

```python
def run_inference_algorithm(
    rng_key: PRNGKey,
    inference_algorithm: Union[SamplingAlgorithm, VIAlgorithm],
    num_steps: int,
    initial_state: ArrayLikeTree = None,
    initial_position: ArrayLikeTree = None,
    progress_bar: bool = False,
    transform: Callable = lambda x: x,
):
    ...
    if initial_state is None and initial_position is None:
        raise ...
    if initial_state is not None and initial_position is not None:
        raise ...
    if initial_position is not None:
        rng_key, init_key = split(rng_key, 2)
        inital_state = inference_algorithm.init(initial_state_or_position, init_key)
```

blackjax/util.py
```diff
@@ -142,12 +142,13 @@ def index_pytree(input_pytree: ArrayLikeTree) -> ArrayTree:
...
     inference_algorithm: Union[SamplingAlgorithm, VIAlgorithm],
     num_steps: int,
     progress_bar: bool = False,
     transform: Callable = lambda x: x,
-) -> tuple[State, State, Info]:
+    streaming=False,
```
junpenglao, line 150:
```suggestion
    return_states_avg=False,
```

blackjax/util.py
```diff
@@ -171,6 +171,8 @@ def run_inference_algorithm(
         A transformation of the trace of states to be returned. This is useful for
         computing determinstic variables, or returning a subset of the states.
         By default, the states are returned as is.
+    streaming
+        if True, `run_inference_algorithm` will take a streaming average of the value of transform, and return that average instead of the full set of samples. This is useful when memory is a bottleneck.
```
junpenglao, line 175: also no diagnostic is return when set to True

blackjax/util.py
```diff
@@ -179,26 +181,64 @@ def run_inference_algorithm(
...
-    keys = split(sample_key, num_steps)
+    keys = split(rng_key, num_steps)
 
     @jit
     def _one_step(state, xs):
         _, rng_key = xs
         state, info = inference_algorithm.step(rng_key, state)
         return state, (transform(state), info)
 
+    def _online_one_step(average_and_state, xs):
```
junpenglao, line 193: There is too much code repetition here, I suggest always computing the average as additional state for now.

junpenglao (changes requested):

blackjax/util.py
reubenharry: OK, I'm open to that. But I'd argue that this isn't ideal. The user has access to `inference_algorithm.init`, so can easily produce `initial_state` from `initial_position` in just one line. Hiding this process inside `run_inference_algorithm` doesn't really make it easier for the user, and creates an awkward `initial_state_or_position` variable along with various errors that have to be raised when both or neither are present. 

tldr: benefit of this feature is very low, added code complexity is high

blackjax/util.py
reubenharry: I did `return_expectation`, is that OK?

blackjax/util.py
reubenharry: This is true. Fixing this was slightly tricky, and I think should be deferred to an upcoming change which will change transform to also take info. I didn't want to do this here because it breaks other parts of the codebase.

blackjax/util.py
reubenharry: Addressed.

blackjax/util.py
junpenglao: It is a valid point, but then maybe we should change to require the input of `inference_algorithm.step` in the API?

reubenharry: That seems reasonable. If you like, I can do that now.

reubenharry: OK, I put it in, as you outlined. This required changing the order of arguments, so I went through the code and made sure `run_inference_algorithm` is called with default arguments everywhere.

junpenglao (approved):

blackjax/util.py
```diff
@@ -142,12 +143,15 @@ def index_pytree(input_pytree: ArrayLikeTree) -> ArrayTree:
...
-    initial_state_or_position: ArrayLikeTree,
     inference_algorithm: Union[SamplingAlgorithm, VIAlgorithm],
     num_steps: int,
+    initial_state: ArrayLikeTree = None,
+    initial_position: ArrayLikeTree = None,
     progress_bar: bool = False,
     transform: Callable = lambda x: x,
-) -> tuple[State, State, Info]:
+    return_state_history=True,
+    expectation: Callable = lambda x: x,
```
junpenglao, line 153: Should we combine the kwargs of `transform` and `expectation` for now? Otherwise we need a better name for `expectation`

blackjax/util.py
```diff
@@ -171,34 +176,88 @@ def run_inference_algorithm(
...
+    )
+
+    if not return_state_history:
+        return average, transform(final_state)
+    else:
+        state_history, info_history = history
+        return transform(final_state), state_history, info_history
+
+
+def streaming_average(O, x, streaming_avg, weight=1.0, zero_prevention=0.0):
```
junpenglao, line 235: we need a more descriptive name here instead of `O`. Moreover, I think it is better to have only `x` as input and have O(x) being called in `one_step` above.

junpenglao (changes requested):
Actually, a few more nit

blackjax/util.py
reubenharry: `expectation` seems to me like an appropriate name, since indeed the value calculated is the expectation. For example, if `expectation=lambda x: x**2`, then you get back $E[x^2]$.

By contrast, `transform` operates on the full history of samples. In the future, I think it would make sense for it to also take Info, so that the user can choose to dispense with (part of) the diagnostic info. That is why I thought it was better to keep them separate.

I think it would in theory be possible to make them the same, if necessary.

blackjax/util.py
reubenharry: Done. I used `expectation` as the new name.

blackjax/util.py
```diff
@@ -171,34 +176,87 @@ def run_inference_algorithm(
...
+
+
+def streaming_average(expectation, streaming_avg, weight=1.0, zero_prevention=0.0):
+    """Compute the streaming average of a function O(x) using a weight.
+    Parameters:
+    ----------
+        O
+            function to be averaged
+        x
+            current state
```
junpenglao, line 242: Update docstring

blackjax/util.py
junpenglao: That makes sense, but looking at the implementation, the return state is always `transform(state)`, which means for a expectation transformation, we should probably do `expectation(transform(x))`?

reubenharry:
The other argument returned, `average`, is the expectation. I'm not sure `expectation(transform(x))` would work, because they are both functions of `x` (the state).

junpenglao:
> The other argument returned, average, is the expectation. I'm not sure expectation(transform(x)) would work, because they are both functions of x (the state).

What i meant here is that:
1. The function intially returns `transform(x)`, which is a function of x (the state). For examples with transform is `lambda x: x.position`. If user wants to compute expectation eg $E[x^2]$, they need to do `np.mean(mcmc_samples ** 2)`
2. Now with the new functionarity that computes the expectation and not storing the full trace, my arguement here is that we should have `trasnform` being a function of the state, but function `expectation` being a function of the output of `transform`.

Does this makes sense?

reubenharry:
Yeah, that makes sense. Updated.

junpenglao (approved):
Thank you!

Issue #667, opened 2024-05-11
window_adaptation excessive memory usage


andrewdipper:
### Describe the issue as clearly as possible:

The scan in window_adaptation by default saves the `AdaptationInfo` for every sample along the way.  This results in memory usage many times in excess of  (num_samples)*(num_variables) and leads to out of memory issues.  However, it looks like the last states are the only information necessary to performing the window adaptation.  

As such it'd be helpful to disable / select what info to store along the way such that the auxiliary info doesn't cause out of memory issues.  Removing it altogether also doesn't seem ideal.   I'd be happy to give a PR if you have an idea of what/how to best include/exclude the extra info.

I believe https://github.com/blackjax-devs/blackjax/issues/529 is the result of the same thing:   The extra buffers are likely for storing the sample by sample info - I get similar outputs.

Thanks

### Steps/code to reproduce the bug:

```python
#a roughed alternative one_step function (in window_adaptation.py) that completely removes the 
#extra info - None is not the ideal solution here

    def one_step(carry, xs):
        _, rng_key, adaptation_stage = xs
        state, adaptation_state = carry

        new_state, info = mcmc_kernel(
            rng_key,
            state,
            logdensity_fn,
            adaptation_state.step_size,
            adaptation_state.inverse_mass_matrix,
            **extra_parameters,
        )
        new_adaptation_state = adapt_step(
            adaptation_state,
            adaptation_stage,
            new_state.position,
            info.acceptance_rate,
        )

        return (
            (new_state, new_adaptation_state),
            None, #-removed-# AdaptationInfo(new_state, info, new_adaptation_state),
        )
```

### Expected result:

```shell
...
```

### Error message:

```shell
...
```

### Blackjax/JAX/jaxlib/Python version information:

```python
BlackJAX 0.1.dev494+g40efb6c.d20240511  # this is 1.2.1 with the jnp.clip PR removed
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]
Jax 0.4.25
Jaxlib 0.4.25
```

### Context for the issue:

I found this trying to reduce memory consumption for a pymc model sampled with blackjax - there's a similar issue there with storing extra info during the actual sampling process.  With both fixes memory consumption and performance are initially looking more than comparable with pymc's numpyro sampler.

junpenglao:
[...]

PR #674, opened 2024-05-15
Enable fitlering of AdaptationInfo

andrewdipper:
This PR allows for AdaptationInfo to be filtered during the window_adaptation process.  Always storing all the auxiliary information can lead to substantial excess memory usage. 

In response to: https://github.com/blackjax-devs/blackjax/issues/667

I put some initial changes here but it seems like this change applies to the other adaptation algorithms.  Perhaps it's worth moving return_all_adapt_info and get_filter_adapt_info_fn to adaptation/base.py and enabling filtering for the other adaptation algorithms.  But I thought I'd get some perspectives before going down that path. 

Let me know any thoughts / changes

### DIFF
--- a/blackjax/adaptation/base.py
+++ b/blackjax/adaptation/base.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import NamedTuple
+from typing import NamedTuple, Set
 
 from blackjax.types import ArrayTree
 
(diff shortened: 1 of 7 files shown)


junpenglao (approved):
Excellent. Thanks!

junpenglao:
> I put some initial changes here but it seems like this change applies to the other adaptation algorithms. Perhaps it's worth moving return_all_adapt_info and get_filter_adapt_info_fn to adaptation/base.py and enabling filtering for the other adaptation algorithms. But I thought I'd get some perspectives before going down that path.

+1, we should apply this to other adaptation algorithms.

andrewdipper:
I added the AdaptationInfo filtering for the other adaptation algorithms.  The second test that is pretty similar to the first is there to make sure the filtering is general to the other adaptation algorithms.

blackjax/adaptation/base.py
```diff
@@ -25,3 +25,33 @@ class AdaptationInfo(NamedTuple):
...
+    state_keys: Set[str] = set(),
+    info_keys: Set[str] = set(),
+    adapt_state_keys: Set[str] = set(),
+):
+    """Generate a function to filter what is saved in AdaptationInfo.  Used
+    for adptation_info_fn parameters of the adaptation algorithms.
+    adaptation_info_fn=get_filter_adapt_info_fn() saves no auxiliary information
+    """
+
+    def filter_tuple(tup, key_set):
```
junpenglao, line 47: BTW, is there a jax tree function that does this kind of filtering?

andrewdipper: I changed it to use tree_map - that seems like the most I can offload to the jax tree functions

junpenglao (approved):
Awesome! thanks!

Issue #679, opened 2024-05-15
Add some metadata to integrators and export coefficients


reubenharry:
### Current behavior

There's no way to currently obtain the number of gradient calls for a given integrator. 

Also, it is sometimes useful to have the integrator coefficients, and these aren't exported from `integrators.py`.

### Desired behavior

Functions like:

```python
def calls_per_integrator_step(c):
    if c==velocity_verlet_coefficients: return 1
    if c==mclachlan_coefficients: return 2
    if c==yoshida_coefficients: return 3

    else: raise Exception

def name_integrator(c):
    if c==velocity_verlet_coefficients: return "velocity_verlet"
    if c==mclachlan_coefficients: return "mclachlan"
    if c==yoshida_coefficients: return "yoshida"

    else: raise Exception
```

Also, to export the coefficients.

junpenglao:
Per #681, we need some other ways to expose this information. 

How about assigning the property to the return integrator object? We can type integrators as Protocal.

reubenharry:
Yes, I think that would be good. The info we want attached is:

1. Order of the integrator
2. Number of grad calls 
3. Name
4. Type (isokinetic or mclachlan)
5. The integrator itself (this could be a function of "isokinetic" or "euclidean", if you think that's a reasonable design decision

junpenglao:
I think the decision ultimately tie to how these information is used in the library beyond benchmarking. Let's circle back when you have some examples.

PR #681, opened 2024-05-18
New integrator, and add some metadata to integrators.py

reubenharry:
Addresses #679 

 - [x] The doc is up-to-date;


reubenharry:
I see that code coverage is complaining about `integrator_order`, `name_integrator` and `calls_per_integrator_step`. These seem useful, but depending on your view, I could remove them

junpenglao:
> I see that code coverage is complaining about `integrator_order`, `name_integrator` and `calls_per_integrator_step`. These seem useful, but depending on your view, I could remove them

Yeah please remove them, depending on what you want to check there should be better ways to return these properties.

blackjax/mcmc/integrators.py
```diff
@@ -23,13 +23,23 @@
 from blackjax.types import ArrayTree
 
 __all__ = [
+    "velocity_verlet_coefficients",
+    "mclachlan_coefficients",
+    "yoshida_coefficients",
+    "omelyan_coefficients",
```
junpenglao, line 29: Remove? Are you planning to use it outside of integrators?

junpenglao (changes requested):

blackjax/mcmc/integrators.py
reubenharry: Also, it seems like code duplication to have `isokinetic_velocity_verlet`, `velocity_verlet`, etc, when we could just use the coefficients with `generate_isokinetic_integrator` and `generate_euclidean_integrator`

junpenglao: > Also, it seems like code duplication to have isokinetic_velocity_verlet, velocity_verlet, etc, when we could just use the coefficients with generate_isokinetic_integrator and generate_euclidean_integrator

The design choice is motivated by how we are envisioning the usage of integrators in the library. Currently the design is to have integrator being a static set.

> As another example, I often want to do benchmarks against different integrators, but want to just iterate over X_coefficients, and then use generate_isokinetic_integrator for MCLMC and generate_euclidean_integrator for HMC

You should iterate through the integrator objects instead

> For example, if I want to run scripts that try different integrators, and I want access to their number of gradient calls. I suppose the other option would be to have a dictionary like {"velocity_verlet": {"num_grads": ..., "name": ..., "order": ...} }. Is that what you'd recommend?

Yes given that these are static, you should put them in your script as static parameters. I dont yet see those are useful in the library outside of benchmarking.

reubenharry: OK, that makes sense. I think my one point of disagreement is that nothing in the code "knows" that `velocity_verlet` and `isokinetic_velocity_verlet` are related. So I will have to have a dictionary of `{"euclidean": velocity_verlet, "isokinetic": isokinetic_velocity_verlet, ...}` when I want to compare each integrator on hmc vs mclmc, which I'm currently doing. This is a little painful, but not the end of the world

reubenharry: Updated, as per request

junpenglao (approved):

PR #692, opened 2024-06-05
Apply function on pytree directly.

junpenglao:
Avoiding unnecessary unpacking.


reubenharry (changes requested):
`streaming_average_update` is called in `mclmc_adaptation` with kw args, so argument name changes would require changes there too.

junpenglao:
@reubenharry Done, thanks!

reubenharry (approved):

reubenharry:
Tests failed!

Issue #690, opened 2024-06-04
smc inner_kernel_tuning forces stateless mcmc_parameters / vmap over shared parameters


andrewdipper:
### Describe the issue as clearly as possible:

When computing the `new_parameter_override` there's only the new_state and info as input.  This means there can't be a dependency on the previous `parameter_override`.  The caveat is that `mcmc_step_fn` can be hacked to transfer the mcmc parameters stored in `StateWithParameterOverride` to the corresponding info ouput.  However this results in duplications if num_mcmc_steps > 1 and it's theading additional data through functions that don't care about it.  In addition it'd be inconsistent with how tempered handles its `lmbda` parameter.  I think there's also interference with the second issue below.
https://github.com/blackjax-devs/blackjax/blob/83bc3a04a64ac3379cd220932e88b44578f2e8e5/blackjax/smc/inner_kernel_tuning.py#L73

Secondly `state.parameter_override` is threaded through as `mcmc_parameters` until it gets below where the `mcmc_kernel` is vmapped in tempered.py.  

https://github.com/blackjax-devs/blackjax/blob/83bc3a04a64ac3379cd220932e88b44578f2e8e5/blackjax/smc/tempered.py#L143-L149

Given that the vmap is done within tempered.py this is necessary for parameters that are different for each particle.  However, there appears to be no way to avoid this for parameters that are shared among the particles.  See https://blackjax-devs.github.io/sampling-book/algorithms/TemperedSMCWithOptimizedInnerKernel.html  Here we are explicitly duplicating a potentially full covariance matrix for each particle.  Given we're quadratic in the number of variables and linear in the number of particles this explodes quite quickly.  Allowing the shared parameters to be closed over prior to the vmap would prevent this. 

I might be missing something but I couldn't find a way to work around the issues without modifying the library.  Essentially I wanted to be able to have a RMH kernel where I could scale the proposal distribution based on the acceptance rate.  For the moment I'm splitting the shared / unshared mcmc parameters and passing `state.parameter_override` to `mcmc_parameter_update_fn`

https://github.com/blackjax-devs/blackjax/blob/83bc3a04a64ac3379cd220932e88b44578f2e8e5/blackjax/smc/inner_kernel_tuning.py#L59-L74

### Steps/code to reproduce the bug:

```python
#not this kind of bug
```

### Expected result:

```shell
#not this kind of bug
```

### Error message:

```shell
#not this kind of bug
```

[...]

PR #694, opened 2024-06-13
Enable shared mcmc parameters with tempered smc

andrewdipper:
Allows for `mcmc_parameters` to be passed to the mcmc kernel as shared parameters prior to applying vmap.  Thus shared parameters will not need to be duplicated for each individual particle.

This change filters `mcmc_parameters` by the length of the first dimension.  Any parameters with length 1 are considered shared (note this is also acceptable in the case of just a single particle) and the rest are unshared.  Shared parameters are then closed over before applying vmap so they don't need duplication.  The behavior remains the same for shared parameters that are duplicated as they are just treated as unshared as before.  This seems like the most reasonable way to handle shared parameters but let me know.

Related to https://github.com/blackjax-devs/blackjax/issues/690
cc @ciguaran

Some notes:
- the some of tests in test_kernel_compatibility.py can use 1 instead of self.n_particles though the correct choice arguably depends on the meaning of the parameter - I changed the ones in test_inner_kernel_tuning.py that were apparently shared
- https://blackjax-devs.github.io/sampling-book/algorithms/TemperedSMC.html and https://blackjax-devs.github.io/sampling-book/algorithms/TemperedSMCWithOptimizedInnerKernel.html should probably change to avoid duplication of the shared parameters
- I couldn't find a way to filter shared/unshared parameters with jax.tree_util functions - None always remains in the structure but maybe I missed something

### DIFF
--- a/blackjax/smc/tempered.py
+++ b/blackjax/smc/tempered.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from functools import partial
 from typing import Callable, NamedTuple
 
 import jax
@@ -119,6 +120,13 @@ def build_kernel(
         """
         delta = lmbda - state.lmbda
 
+        shared_mcmc_parameters = {
+            k: v[0, ...] for k, v in mcmc_parameters.items() if v.shape[0] == 1
+        }
+        unshared_mcmc_parameters = {
+            k: v for k, v in mcmc_parameters.items() if v.shape[0] != 1
+        }
+
         def log_weights_fn(position: ArrayLikeTree) -> float:
             return delta * loglikelihood_fn(position)
 
@@ -127,11 +135,13 @@ def build_kernel(
             tempered_loglikelihood = state.lmbda * loglikelihood_fn(position)
             return logprior + tempered_loglikelihood
 
+        shared_mcmc_step_fn = partial(mcmc_step_fn, **shared_mcmc_parameters)
+
         def mcmc_kernel(rng_key, position, step_parameters):
             state = mcmc_init_fn(position, tempered_logposterior_fn)
 
             def body_fn(state, rng_key):
-                new_state, info = mcmc_step_fn(
+                new_state, info = shared_mcmc_step_fn(
                     rng_key, state, tempered_logposterior_fn, **step_parameters
                 )
                 return new_state, info
@@ -142,7 +152,7 @@ def build_kernel(
 
         smc_state, info = smc.base.step(
             rng_key,
-            SMCState(state.particles, state.weights, mcmc_parameters),
+            SMCState(state.particles, state.weights, unshared_mcmc_parameters),
             jax.vmap(mcmc_kernel),
             jax.vmap(log_weights_fn),
             resampling_fn,
--- a/blackjax/smc/tempered.py
+++ b/blackjax/smc/tempered.py
@@ -109,6 +109,9 @@ def build_kernel(
             Current state of the tempered SMC algorithm
         lmbda
             Current value of the tempering parameter
+        mcmc_parameters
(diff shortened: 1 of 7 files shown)


blackjax/smc/tempered.py
```diff
@@ -119,6 +120,13 @@ def kernel(
         """
         delta = lmbda - state.lmbda
 
+        shared_mcmc_parameters = {
+            k: v[0, ...] for k, v in mcmc_parameters.items() if v.shape[0] == 1
+        }
+        unshared_mcmc_parameters = {
+            k: v for k, v in mcmc_parameters.items() if v.shape[0] != 1
+        }
+
```
junpenglao, line 129: Please put them in a single for loop.

junpenglao (commented):
LGTM overall - is it still draft?

blackjax/smc/tempered.py
andrewdipper: Done

andrewdipper:
The draft was mostly for if the tests should be modified - removed it

tests/smc/test_inner_kernel_tuning.py
```diff
@@ -281,7 +281,7 @@ def test_with_adaptive_tempered(self):
 
         def parameter_update(state, info):
             return extend_params(
```
ciguaran, line 283: calling extend params shouldn't be needed anymore

andrewdipper: I'm still using extend_params since it needs to convert to an array and add a leading dimension of length one.  I could have added a separate helper but it'd do the same thing as extend_params but just without the jnp.repeat part

junpenglao: I think it is worth to modify `extend_params` into something simple without the `jnp.repeat`.

Would this work?
```python
def extend_params(params: ArrayLikeTree):
    return jax.tree.map(lambda a: jnp.expand_dims(a, 0), params)
```

andrewdipper: That would work, perhaps with potentially a `jnp.asarray()` included.  The concern I have with modifying `extend_params` is I'd assume there's still a use case for it with initializing unshared parameters?  

A current use case is for the remaining tests that are using duplicated parameters for testing the unshared case.

junpenglao: @ciguaran Is there any other use case besides dimension matching initially?

ciguaran: There's none.

ciguaran: I'd assume there's still a use case for it with initializing unshared parameters -> that is a good concern although I think that in most cases we wouldn't initialize by "extending" with copies. We would do it by sampling from some probability distribution. So you can modify it as Junpeng suggested.

tests/smc/test_tempered_smc.py
```diff
@@ -65,16 +65,25 @@ def logprior_fn(x):
...
-        hmc_parameters = extend_params(
-            num_particles,
-            {
-                "step_size": 10e-2,
-                "inverse_mass_matrix": jnp.eye(2),
-                "num_integration_steps": 50,
-            },
+        hmc_parameters_list = [
+            extend_params(
+                num_particles if extend else 1,
```
ciguaran, line 70: I'd add a comment highlighting that here you are testing that extending with a copy is the same as having a 1 dimension parameter. Otherwise it may be tricky to get why this is even hapening.

tests/smc/test_inner_kernel_tuning.py
andrewdipper: ah, that makes sense.  thanks

tests/smc/test_tempered_smc.py
andrewdipper: done

junpenglao (approved):
Thank you!

PR #711, opened 2024-07-31
Fixing typo in example on index page - leading to 30x faster runtime.

johannahaffner:
The jitted step remained unused, leading to the example running with an uncompiled nuts.step. 

Changing this reduces the execution time by a factor of 30 on my system and showcases blackjax' speed.


AdrienCorenflos (approved):
No comment

PR #712, opened 2024-08-03
Enable progress bar under pmap

andrewdipper:
Enables proper progress bar behavior of `progress_bar_scan` under pmap.  However the progress bars are not tied to physical devices since `io_callback` won't expose the device id.  The bars behave as if they are physically tied but are always sorted by most to least complete (they're actually randomly shuffled but one bar will always be first and another last).

I haven't visually tested this under a multi-gpu setup as I don't have an identical pair of gpus - just using `JAX_PLATFORMS='cpu'`

https://github.com/blackjax-devs/blackjax/pull/655 - Related: changes progress bar to tqdm and requires the user to label devices - a different approach

 - [ ] We should be able to understand what the PR does from its title only;
 - [ ] There is a high-level description of the changes;
 - [ ] There are links to *all* the relevant issues, discussions and PRs;
 - [ ] The branch is rebased on the latest `main` commit;
 - [ ] `pre-commit` is installed and configured on your machine, and you ran it before opening the PR;
 - [ ] The doc is up-to-date;

### DIFF
--- a/blackjax/adaptation/window_adaptation.py
+++ b/blackjax/adaptation/window_adaptation.py
@@ -334,16 +334,22 @@ def window_adaptation(
         if progress_bar:
             print("Running window adaptation")
             one_step_ = jax.jit(progress_bar_scan(num_steps)(one_step))
+            start_state = ((init_state, init_adaptation_state), -1)
         else:
             one_step_ = jax.jit(one_step)
+            start_state = (init_state, init_adaptation_state)
 
         keys = jax.random.split(rng_key, num_steps)
         schedule = build_schedule(num_steps)
         last_state, info = jax.lax.scan(
             one_step_,
-            (init_state, init_adaptation_state),
+            start_state,
             (jnp.arange(num_steps), keys, schedule),
         )
+
+        if progress_bar:
+            last_state, _ = last_state
+
         last_chain_state, last_warmup_state, *_ = last_state
 
         step_size, inverse_mass_matrix = adapt_final(last_warmup_state)
--- a/blackjax/progress_bar.py
+++ b/blackjax/progress_bar.py
@@ -22,6 +22,7 @@ from jax.experimental import io_callback
 def progress_bar_scan(num_samples, print_rate=None):
     "Progress bar for a JAX scan"
     progress_bars = {}
+    idx_map = {}
 
     if print_rate is None:
         if num_samples > 20:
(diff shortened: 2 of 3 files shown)


junpenglao:
Thanks! I like this better than https://github.com/blackjax-devs/blackjax/pull/655 because it does not require explicit `num_chain`. @zaxtax thoughts?

zaxtax:
I don't like how the bars get shuffled. I worry that it would be hard to
follow the progress. Maybe the approaches can be combined somehow

andrewdipper:
The initial random shuffle is a bit wonky.  I believe it comes from fastprogress - the bars should always be in sorted order if we used the change from fastprogress to tqdm.  That's based on similar tests with numpyro's progressbar and tqdm - but I'll verify it.  Would that be better?

junpenglao:
> The initial random shuffle is a bit wonky. I believe it comes from fastprogress - the bars should always be in sorted order if we used the change from fastprogress to tqdm. That's based on similar tests with numpyro's progressbar and tqdm - but I'll verify it. Would that be better?

Yes - if that works, how about I merge this PR and then @zaxtax you can update https://github.com/blackjax-devs/blackjax/pull/655?

andrewdipper:
I had it wrong - the randomization was a race condition.  Added a lock and now the bars are always ordered by most to least complete

zaxtax:
But do the bars move around if one chain overtakes another?

andrewdipper:
No, the order will always be in decreasing order of chain progress.  As such the chain to progress bar mapping will change if one overtakes another.  We'd need the equivalent of a chain/device id to avoid this - but from what I've seen so far the new callbacks don't support identifying the underlying device.

junpenglao:
> No, the order will always be in decreasing order of chain progress. As such the chain to progress bar mapping will change if one overtakes another. We'd need the equivalent of a chain/device id to avoid this - but from what I've seen so far the new callbacks don't support identifying the underlying device.

So IIUC, we cannot have identifier like `Chain 0`, `Chain 1` ...? I am not too bothered by the sorting of the chains, but would be great if there is some identifier, so users are not surprised by the behavior.

zaxtax:
In my PR, I handle this by making a chain id that's carried along with the
sample iteration. It's a bit inelegant but straightforward and doesn't
require special machinery from the io callback code

zaxtax:
We could make a helper that hides that plumbing?

andrewdipper:
> So IIUC, we cannot have identifier like Chain 0, Chain 1 ...? I am not too bothered by the sorting of the chains, but would be great if there is some identifier, so users are not surprised by the behavior.

Correct.  Progress is being used as a surrogate id.

> We could make a helper that hides that plumbing?

If I'm not mistaken I think this helper has to reside outside the pmap and wouldn't be completely transparent.  But https://github.com/blackjax-devs/blackjax/pull/655 is more straightforward from a handling perspective.

It might be possible to use the `io_callback` return value to set a chain id to be used in future io_callbacks.  And as a fallback hashing the initial (or whole sequence) of `(input, carry)` could resolve into a unique id.  I'm not sure how friendly jax is to those solutions but I'll take a look.

andrewdipper:
The latest commit uses the return value from the `io_callback` to uniquely identify the chains.  The map from chains to progress bars won't change.

However, this changes the carry used in progress_bar_scan which is a problem if it is used externally.  That'd have to be a slow change.  But it does avoid forcing the user to do the device identification with part of the input.  A wrapper over scan itself might clean things up a bit.

zaxtax:
I'm ok with the PR as is @junpenglao and definitely think we should explore a wrapper over scan in the future.

junpenglao:
Great, thank you @andrewdipper for the contribution!

junpenglao (approved):

zaxtax:
Oh should we update any examples to show how the new functionality works @andrewdipper ? Or does this not change the API?

andrewdipper:
I didn't see any examples using `progress_bar_scan` and only the two uses within blackjax so we should be good.

zaxtax:
No it looks like the examples in the sampling book don't

https://blackjax-devs.github.io/sampling-book/algorithms/cyclical_sgld.html

PR #714, opened 2024-08-06
Simplify `run_inference_algorithm`

reubenharry:
The `run_inference_algorithm` code currently has a mode to incrementally compute a value (e.g. an incremental expectation), which saves memory. This PR presents a cleaner (and in my opinion, more functional) version of the code. The core idea is that to incrementally compute an expectation, the right thing to do is to modify the kernel, and then pass that kernel to `run_inference_algorithm`, rather than modify `run_inference_algorithm`. As part of this change, we also let `transform` be a general function on `(state, info)`, which allows information from the info to be used (which we have needed). 

 - [ ] The doc is up-to-date;


tests/mcmc/test_sampling.py
```diff
@@ -223,10 +223,11 @@ def test_mala(self):
 
         mala = blackjax.mala(logposterior_fn, 1e-5)
         state = mala.init({"coefs": 1.0, "log_scale": 1.0})
-        _, states, _ = run_inference_algorithm(
+        _, states = run_inference_algorithm(
             rng_key=inference_key,
             initial_state=state,
             inference_algorithm=mala,
+            transform=lambda state, info: state,
```
junpenglao, line 230: Let's simplify all the test in this file to `transform=lambda state, info: state.position,` if it applies.

blackjax/util.py
```diff
@@ -166,104 +163,148 @@ def run_inference_algorithm(
...
     if initial_state is not None and initial_position is not None:
         raise ValueError(
-            "Only one of `initial_state` or `initial_position` must be provided."
+            "Only one of initial_state or initial_position must be provided."
         )
 
-    if initial_state is None:
-        rng_key, init_key = split(rng_key, 2)
+    rng_key, init_key = split(rng_key, 2)
+    if initial_position is not None:
```
junpenglao, line 193: revert these changes.

blackjax/util.py
```diff
@@ -166,104 +163,148 @@ def run_inference_algorithm(
...
+    flat_expectation, unravel_fn = ravel_pytree(expectation)
+    total, average = incremental_val
+    flat_average, _ = ravel_pytree(average)
+    average = (total * flat_average + weight * flat_expectation) / (
+        total + weight + zero_prevention
     )
-    return current_weight, current_average
+    total += weight
+    incremental_val = (total, unravel_fn(average))
+    return incremental_val
```
junpenglao, line 310: could you update this to use `jax.tree.map`

junpenglao (changes requested):
Some nits

blackjax/util.py
```diff
@@ -212,58 +196,117 @@ def run_inference_algorithm(
...
+    average = tree_map(
+        lambda exp, av: (total * av + weight * exp)
+        / (total + weight + zero_prevention),
+        expectation,
+        average,
     )
-    return current_weight, current_average
+    total += weight
+    incremental_val = total, average
+    return incremental_val
```
junpenglao, line 312:
```suggestion
    return total, average
```

blackjax/util.py
reubenharry: Updated

tests/test_util.py
```diff
@@ -41,37 +41,49 @@ def logdensity_fn(x):
             10,
         )
 
-        init_key, run_key = jax.random.split(self.key, 2)
-
+        init_key, state_key, run_key = jax.random.split(jax.random.PRNGKey(0), 3)
```
junpenglao, line 44:
```suggestion
        init_key, state_key, run_key = jax.random.split(self.key, 3)
```

blackjax/util.py
reubenharry: Done

tests/test_util.py
reubenharry: Good catch

junpenglao (approved):

PR #720, opened 2024-08-14
Feat: Implement Full-Rank VI

gil2rok:
Implement variational inference (VI) with full-rank Gaussian approximation

While [mean-field VI](https://github.com/blackjax-devs/blackjax/blob/main/blackjax/vi/meanfield_vi.py) learns a Gaussian approximation $N(\mu, \sigma I)$ with diagonal covariance $\sigma I$, full-rank VI learns a Gaussian approximation $N(\mu, \Sigma)$ with *full-rank* covariance $\Sigma$.

We use the Cholesky decomposition $\Sigma = C C^T$ parameterized by the log standard deviation $\rho$ and lower triangle matrix $L$ such that $C = \exp(\rho) + L$. This (1) ensures $\Sigma$ remains symmetric and positive-definite during optimization [^1] and (2) admits better sampling and log density computation (with improved time complexity, space complexity, and numerical stability)[^2]. Thus the full-rank Gaussian is parameterized by $(\mu, \rho, L)$.

To-Do:
- [ ] Support arbitrary PyTrees instead of just arrays
- [ ] Decide on format for lower triangular matrix $L$ ($n$-length flattened array vs dense $2D$ array for $n = d(d+1)/2$)
- [ ] Confirm initialization strategy from standard normal
- [ ] Write tests (current test fails)

[^1]: [Automatic Differentiation Variational Inference](https://arxiv.org/pdf/1603.00788) Section 2.4
[^2]: [Lecture notes](https://www.cs.helsinki.fi/u/ahonkela/teaching/compstats1/book/multivariate-normal-distributions-and-numerical-linear-algebra.html)

### DIFF
--- a/blackjax/vi/fullrank_vi.py
+++ b/blackjax/vi/fullrank_vi.py
@@ -1,0 +1,165 @@
+# Copyright 2020- The Blackjax Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Callable, NamedTuple
+
+import jax
+import jax.numpy as jnp
+import jax.scipy as jsp
+from optax import GradientTransformation, OptState
+
+from blackjax.base import VIAlgorithm
+from blackjax.types import ArrayLikeTree, ArrayTree, PRNGKey
+
+__all__ = [
+    "FRVIState",
+    "FRVIInfo",
+    "sample",
+    "generate_fullrank_logdensity",
+    "step",
+    "as_top_level_api",
+]
+
+
+class FRVIState(NamedTuple):
+    mu: ArrayTree
+    rho: ArrayTree
+    L: ArrayTree
+    opt_state: OptState
+
+
+class FRVIInfo(NamedTuple):
+    elbo: float
+
+
+def init(
+    position: ArrayLikeTree,
+    optimizer: GradientTransformation,
+    *optimizer_args,
+    **optimizer_kwargs,
+) -> FRVIState:
+    """Initialize the full-rank VI state."""
+    mu = jax.tree.map(jnp.zeros_like, position)
+    rho = jax.tree.map(jnp.zeros_like, position)
+    L = jax.tree.map(lambda x: jnp.zeros((*x.shape, x.shape)), position)
+    opt_state = optimizer.init((mu, rho, L))
+    return FRVIState(mu, rho, L, opt_state)
+
+
+def step(
+    rng_key: PRNGKey,
+    state: FRVIState,
+    logdensity_fn: Callable,
+    optimizer: GradientTransformation,
+    num_samples: int = 5,
+    stl_estimator: bool = True,
+) -> tuple[FRVIState, FRVIInfo]:
+    """Approximate the target density using the full-rank approximation.
+
+    Parameters
+    ----------
+    rng_key
+        Key for JAX's pseudo-random number generator.
+    init_state
+        Initial state of the full-rank approximation.
+    logdensity_fn
+        Function that represents the target log-density to approximate.
+    optimizer
+        Optax `GradientTransformation` to be used for optimization.
+    num_samples
+        The number of samples that are taken from the approximation
+        at each step to compute the Kullback-Leibler divergence between
+        the approximation and the target log-density.
+    stl_estimator
+        Whether to use stick-the-landing (STL) gradient estimator :cite:p:`roeder2017sticking` for gradient estimation.
+        The STL estimator has lower gradient variance by removing the score function term
+        from the gradient. It is suggested by :cite:p:`agrawal2020advances` to always keep it in order for better results.
+
+    """
+
+    parameters = (state.mu, state.rho, state.L)
+
+    def kl_divergence_fn(parameters):
+        mu, rho, L = parameters
+        z = _sample(rng_key, mu, rho, L, num_samples)
+        if stl_estimator:
+            parameters = jax.tree_map(jax.lax.stop_gradient, (mu, rho, L))
+        logq = jax.vmap(generate_fullrank_logdensity(mu, rho, L))(z)
+        logp = jax.vmap(logdensity_fn)(z)
+        return (logq - logp).mean()
+
+    elbo, elbo_grad = jax.value_and_grad(kl_divergence_fn)(parameters)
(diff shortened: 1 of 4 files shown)


gil2rok:
I would love specific feedback on how to support PyTrees instead of just JAX arrays. I am trying to better grok JAX's PyTrees.

For example, consider my sampling implementation vs that in `meanfield_vi.py`:

```python
def _sample(rng_key, mu, rho, L, num_samples):
    cholesky = jnp.tril(L, k=-1) + jnp.diag(jnp.exp(L))
    eps = jax.random.normal(rng_key, (num_samples,) + mu.shape)
    return mu + eps @ cholesky.T
``` 

```python
def _sample(rng_key, mu, rho, num_samples):
    sigma = jax.tree.map(jnp.exp, rho)
    mu_flatten, unravel_fn = jax.flatten_util.ravel_pytree(mu)
    sigma_flat, _ = jax.flatten_util.ravel_pytree(sigma)
    flatten_sample = (
        jax.random.normal(rng_key, (num_samples,) + mu_flatten.shape) * sigma_flat
        + mu_flatten
    )
    return jax.vmap(unravel_fn)(flatten_sample)
```

How + why would I change my code? Any resource recommendations would be greatly appreciated.

blackjax/vi/fullrank_vi.py
```diff
@@ -0,0 +1,165 @@
...
+    cholesky = jnp.tril(L, k=-1) + jnp.diag(jnp.exp(L))
+    log_det = 2 * jnp.sum(rho)
+    const = -0.5 * mu.shape[-1] * jnp.log(2 * jnp.pi)
+
+    def fullrank_logdensity(position):
+        y = jsp.linalg.solve_triangular(cholesky, position - mu, lower=True)
+        mahalanobis_dist = jnp.sum(y ** 2, axis=-1)
+        return const - 0.5 * log_det - 0.5 * mahalanobis_dist
+
+    return fullrank_logdensity
```
junpenglao, line 165: Why not use `multivariate_normal.logpdf` from JAX?

junpenglao (commented):
PyTree with covariate matrix is a headache to deal with. I suggest you take a look at how [pathfinder](https://github.com/blackjax-devs/blackjax/blob/main/blackjax/vi/pathfinder.py#L127) deal with it internally: basically all the state parameter is represented as flatten array, and you only unflatten at the end

blackjax/vi/fullrank_vi.py
gil2rok: I mistakenly thought `jax.random.multivariate_normal.logpdf` only accepts a covariance matrix, not a cholesky factor parameterization. Will change this!

gil2rok:
A couple notes about the Cholesky factor:

1. I believe my implementation is incorrect. I compute the $d \times d$ Cholesky factor as $C = \exp(\rho) + L$ for log standard deviation $\rho$ and lower triangle matrix $L$. However, I store $L$ as a $d \times d$ matrix instead of *just* the lower triangle:
```python
L = jax.tree.map(lambda x: jnp.zeros((*x.shape, x.shape)), position)
```
During optimization, this means we may learn non-zero values for the diagonal and upper triangle in $L$! Thus I instead will store $L$ as list of length $n$ representing the number of elements in the lower triangle of a $d \times d$ matrix *excluding* the diagonal. Thus $n = \sum_{i=1}^{d-1} i = (d - 1)(d) / 2$. We now construct the Cholesky factor as:
```python
def unflatten_lower_triangular(tril_flat):
    n = tril_flat.size  # Number of elements in the lower triangular part
    d = int(jnp.sqrt(1 + 8 * n) - 1) // 2  # Dimension of the original matrix
    
    lower_tri_matrix = jnp.zeros((d, d))
    indices = jnp.tril_indices(d, k=-1)  # Indices for the lower triangle (excluding the diagonal)
    return lower_tri_matrix.at[indices].set(tril_flat)

C = jnp.diag(jnp.exp(rho)) + unflatten_lower_triangular(L)
```

2. Given my full-rank Gaussian parameterization as $(\mu, \rho, L)$, how should I flatten/unflatten to work with arbitrary Pytrees? Following @junpenglao 's advice, I see Pathfinder's `approximate` function:

```python
def approximate(...):
    initial_position_flatten, unravel_fn = ravel_pytree(initial_position)
    ...
    unravel_fn_mapped = jax.vmap(unravel_fn)
    pathfinder_result = PathfinderState(
        elbo,
        unravel_fn_mapped(position),
        unravel_fn_mapped(grad_position),
        alpha,
        beta,
        gamma,
    )
    ...
```

How would I apply this to my `_sample()` function? I'm struggling to understand why flattening helps us handle arbitrary PyTrees, and how this works for the lower triangle $L$ vs mean $\mu$ or log std $\rho$. Is this correct or is more flattening needed?

```python
def _sample(rng_key, mu, rho, L, num_samples):
    cholesky = jnp.diag(jnp.exp(rho)) + unflatten_lower_triangular(L)
    eps = jax.random.normal(rng_key, (num_samples,) + mu.shape)
    return mu + eps @ cholesky.T
```

gil2rok:
1. Largely disregard my previous comments about the Cholesky factor in https://github.com/blackjax-devs/blackjax/pull/720#issuecomment-2292018721. I would love some feedback on my current implementation, which more closely follows the original pull request https://github.com/blackjax-devs/blackjax/pull/479.
2. For simplicity, I changed the code to use `jax.random.multivariate_normal.logpdf` even though it redundantly recommits the Cholesky decomposition internally. Do you think I should keep it or revert back to my more efficient implementation?

blackjax/vi/fullrank_vi.py
```diff
@@ -0,0 +1,188 @@
...
+
+def generate_fullrank_logdensity(mu, chol_params):
+    mu_flatten, _ = jax.flatten_util.ravel_pytree(mu)
+    chol_factor = _unflatten_cholesky(chol_params)
+    cov = chol_factor @ chol_factor.T
+
+    def fullrank_logdensity(position):
+        position_flatten = jax.flatten_util.ravel_pytree(position)[0]
+        # TODO: inefficient because of redundant cholesky decomposition
+        return jsp.stats.multivariate_normal.logpdf(position_flatten, mu_flatten, cov)
```
junpenglao, line 186: You have a good point about computation efficiency, let's keep the cholesky version of the logpdf. Could you rewrite the API to similar to `multivariate_normal.logpdf`, and use function partial when you call it?

gil2rok: Can you specify the API you imagine my implementation would have and how function partial would be applied to it? Feel free to reference my code I just committed.

junpenglao:
I see, you are following the pattern in meanfield VI, I think there are a few places need refactoring there. Let me send out a PR so you see what i meant.

gil2rok:
> I see, you are following the pattern in meanfield VI, I think there are a few places need refactoring there. Let me send out a PR so you see what i meant.

Once you finish refactoring Meanfield VI, I'm happy to adapt the new style. Let me know when you're done!

gil2rok:
Update: need to figure out why the full covariance matrix is not being recovered. May need to come back to this in a few weeks b/c of some deadlines.

ylefay:
Thank you for your work, I am really looking forward to try it ;)

Issue #710, opened 2024-07-29
Implement Waste-free SMC


ciguaran:
https://arxiv.org/abs/2011.02328

AdrienCorenflos:
This is already implemented (and the default I think)
https://github.com/blackjax-devs/blackjax/blob/f8db9aa04d83fed4b3bde0f37ddf6194b229308f/blackjax/smc/base.py#L115
Is this not properly documented?

ciguaran:
The thing is that in the current state is not super usable since you can only use it from base.py. In other words, tempered and adaptive tempered are coupled to the classic SMC case (iterate MCMC and take the last sample).

PR #721, opened 2024-08-15
Waste Free SMC available for adaptive tempered and tempered SMC. 

ciguaran:
Although SMC base.py was ready for waste-free, tempered and adaptive tempered are coupled to using classic SMC (iterate MCMC and keep the last sample). This PR makes Waste-free available at all levels of the hierarchy.
https://github.com/blackjax-devs/blackjax/issues/710


blackjax/smc/tempered.py
```diff
@@ -48,12 +48,42 @@ def init(particles: ArrayLikeTree):
...
+def update_and_take_last(
+    mcmc_init_fn,
+    tempered_logposterior_fn,
+    shared_mcmc_step_fn,
+    num_mcmc_steps,
+    n_particles,
+):
+    """
+    Given N particles, runs num_mcmc_steps of a kernel starting at each particle, and
+    returns the last values, waisting the previous num_mcmc_steps-1
```
junpenglao, line 60:
```suggestion
    returns the last values, discarding the previous num_mcmc_steps-1
```

junpenglao (dismissed):

AdrienCorenflos:
By the way, whilst we are on that, I remember being a bit dubious about waste free for parallel environments: when you have a lot of processors the analysis in the paper does not hold anymore.
@hai-dang-dau have you ever done some back of the envelope calculations on this? I'd like to be able to recommend people to switch off waste free (and maybe increase the number of intermediary distributions in exchange?) if they have a certain number of processors.

hai-dang-dau:
It's true that waste-free SMC shines best when the number of processors is small relative to the number of particles. (Say, a dozen of processors for several thousand of particles.) I think this is the case for most practical applications.

If your number of particles is comparable to that of processors (say, 500 processors and 500 particles), then yes standard SMC is better.

However, if you have that many processors, you probably have a lot of memory as well. When you increase the number of particles to about a million (still with 500 CPUs), waste-free starts to be preferable again.

AdrienCorenflos:
Well, not really: a commercial grade GPU has fairly little memory so it's
not all that black and white. I'll have a think.

hai-dang-dau:
In my previous post, by "processors" I meant the number of MCMC chains that can be run on parallel. Although GPU has a lot of parallel processing capacity, I'm not sure how this translates into the number of MCMC chains. The delicate point is that one MCMC chain itself involves a lot of parallelizable operations (ie matrix multiplications for calculating the likelihood or for generating preconditioned proposals). It is these primitive operations that are usually parallelized automatically by JAX/XLA, not the higher level operations that were taken into account in my reasoning above.

Having said that, it's true that we developed and tested waste-free SMC with CPUs in mind; so it'd definitely be relevant to have recommendations tailored to GPU-accelerated environment.

blackjax/smc/waste_free.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        states, infos = jax.vmap(mcmc_kernel)(rng_key, position, step_parameters)
+
+        # step particles is num_resmapled, num_mcmc_steps, dimension_of_variable
+        # want to transformed into num_resampled * num_mcmc_steps, dimension of variable
+        def reshape_step_particles(x):
+            if len(x.shape) > 2:
+                return x.reshape((x.shape[0] * x.shape[1], -1))
+            else:
+                return x.flatten()
```
junpenglao, line 58:
```suggestion
        def reshape_step_particles(x):
            num_resmapled, num_mcmc_steps, *dimension_of_variable = x.shape
            return x.reshape((num_resmapled*num_mcmc_steps, *dimension_of_variable))
```

junpenglao (approved):
Minor nit

junpenglao (dismissed):

junpenglao:
Feel free to submit after fixing the code style

ciguaran:
@junpenglao the review gets dismissed whenever we commit something new, even if it is just github's commiting a reviewer suggestion

PR #724, opened 2024-08-21
Bug fix

reubenharry:
This PR puts right a few small (but important) bugs in the adaptation algorithm for unadjusted MCLMC. There was a division by 0 error, and a failure to catch NaNs in a certain part of the code.

### DIFF
--- a/blackjax/adaptation/mclmc_adaptation.py
+++ b/blackjax/adaptation/mclmc_adaptation.py
@@ -23,7 +23,6 @@ from blackjax.diagnostics import effective_sample_size
 from blackjax.util import incremental_value_update, pytree_size
 
 
-
 class MCLMCAdaptationState(NamedTuple):
     """Represents the tunable parameters for MCLMC adaptation.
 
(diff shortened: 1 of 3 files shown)


blackjax/adaptation/mclmc_adaptation.py
```diff
@@ -260,6 +259,9 @@ def L_step_size_adaptation(state, params, num_steps, rng_key):
                     xs=(jnp.ones(steps), keys), state=state, params=params
                 )
 
+        jax.debug.print(
```
junpenglao, line 262: remove

blackjax/util.py
```diff
@@ -281,6 +281,10 @@ def transform(state_and_incremental_val, info):
     return SamplingAlgorithm(init_fn, update_fn), transform
 
 
+def safediv(x, y):
```
junpenglao, line 284: is this really make a differences?

reubenharry: Yes, because there was a 0/0 case. It could be avoided other ways, but this was the cleanest.

blackjax/adaptation/mclmc_adaptation.py
reubenharry: Removed

junpenglao (approved):

Issue #726, opened 2024-08-23
SMC: Allow for different path of distributions other than tempering.


ciguaran:
SMC works by traversing a path of distributions, starting with an initial distribution, and progressively getting closer to a target distribution. There are many ways to construct such path.` tempered.py `implements the tempering approach. 

We could for example implement a Partial posteriors path (where the likelihood of each intermediate distribution has more datapoints). Right now, in order to do so, we would need to extend from `base.py`, loosing all the inner kernel parameters work that happens inside` tempered.py.`

My proposal is to split `tempered.py` into two. The first function should do the tempering work, the second function all the wiring for MCMC parameters, and call to `base.py`. This way implementing other distribution paths doesn't need to handle MCMC kernels construction.

AdrienCorenflos:
I don't see how you would do data tempering efficiently within JAX but apart from this I'm okay with the suggestion.

PR #729, opened 2024-08-27
Partial posteriors SMC and refactor to decouple tempering from SMC construction

ciguaran:
- Refactors tempered.py splitting (a) tempering: constructing a tempered posterior and weight function from (b) construcing a SMC kernel from MCMC configuration.
- Implements SMC from partial posteriors path (see Section 2.2 of https://arxiv.org/pdf/2007.11936)
https://github.com/blackjax-devs/blackjax/issues/726


blackjax/smc/partial_posteriors_path.py
```diff
@@ -0,0 +1,127 @@
...
+
+class PartialPosteriorsSMCState(NamedTuple):
+    """Current state for the tempered SMC algorithm.
+
+    particles: PyTree
+        The particles' positions.
+    weights:
+        Weights of the particles, so that they represent a probability distribution
+    selector:
+        {Datapoints used to calculate the posterior the particles represent
```
junpenglao, line 20:
```suggestion
        Datapoints used to calculate the posterior the particles represent, a 1D boolean array to indicante which pariticles to include in the computation of the observed likelihood.
```

Is my understanding correct here? Also instead of `selector` I suggest renaming it to `particle_mask`

junpenglao (commented):
LGTM overall, I have a minor question on `partial_logposterior_factory` and `selector`.

Also, could you go through the doc string and make sure the formatting is consistent? There is some spacing issue (some of which I highlighted.

blackjax/smc/partial_posteriors_path.py
ciguaran: hei! is not. If your likelihood is f(theta|d_1,...d_n) where d_i is a datapoint. The selector is used to choose
which ds get into the likelihood calculation.

ciguaran: Recall that particles are just posterior samples. The datapoints here are from some dataset that defines the likelihood. Of course, if you are working with a potential-like function that has no points, this algorithm makes no sense.

junpenglao: ah I see, would naming it `data_mask` be appropriate here?

ciguaran:
@junpenglao covered the comments. For some reason the benchmark is failing for HMC (code which is untouched by this PR) 🤔

junpenglao (approved):

junpenglao:
Yeah i think it is just some flakyness - merging. Thanks a lot!

Issue #725, opened 2024-08-22
Implement mass matrix in Barker's kernel


AdrienCorenflos:
Barker's proposal should take in a mass matrix for pre-conditioning. The change should follow appendix G here https://arxiv.org/abs/1908.11812 and the interface of HMC https://github.com/blackjax-devs/blackjax/blob/4a11236930a49cf06ebb41e530002a90c6ecec21/blackjax/mcmc/hmc.py#L120

As a unit test, the implementer should check that barker for π(x) with mass matrix M has exactly the same outputs as the scaled version in per appendix G.

PR #731, opened 2024-09-01
Add pre-conditioning matrix to Barker proposal

ismael-mendoza:
# Overview

This PR attempts to add a preconditioning matrix to the barker proposal implementation, as in appendix G of https://arxiv.org/abs/1908.11812

- Relevant discussion #723 
- Relevant issue #725 

# Discussion / Questions
- I changed the `_compute_acceptance_probability` to be flat given how it was necessary to matrix-multiply with the pre-conditioning matrix. Is that OK? 
- Do I need to update the PDF?  (not used anywhere in the codebase)
- It's not very clear in the Barker paper but the step_size (sigma) is a 'global scale' and the inverse mass matrix is separate.  See implementation from authors here (in R) - https://github.com/gzanella/barker/blob/master/functions.R

# Thank you for opening a PR!

 - [x] The doc is up-to-date;
 - [x] *If I add a new sampler** I added/updated related [examples](https://github.com/blackjax-devs/blackjax/tree/main/examples)

### DIFF
--- a/blackjax/mcmc/barker.py
+++ b/blackjax/mcmc/barker.py
@@ -22,7 +22,7 @@ from jax.scipy import stats
 
 from blackjax.base import SamplingAlgorithm
 from blackjax.mcmc.proposal import static_binomial_sampling
-from blackjax.types import ArrayLikeTree, ArrayTree, PRNGKey
+from blackjax.types import Array, ArrayLikeTree, ArrayTree, PRNGKey
 
 __all__ = ["BarkerState", "BarkerInfo", "init", "build_kernel", "as_top_level_api"]
 
@@ -81,7 +81,7 @@ def build_kernel():
     """
 
     def _compute_acceptance_probability(
-        state: BarkerState, proposal: BarkerState, C_t: jnp.Array, C_t_inv: jnp.Array
+        state: BarkerState, proposal: BarkerState, C_t: Array, C_t_inv: Array
     ) -> float:
         """Compute the acceptance probability of the Barker's proposal kernel."""
 
@@ -106,7 +106,7 @@ def build_kernel():
         state: BarkerState,
         logdensity_fn: Callable,
         step_size: float,
-        inverse_mass_matrix: jnp.Array,
+        inverse_mass_matrix: Array,
     ) -> tuple[BarkerState, BarkerInfo]:
         """Generate a new sample with the Barker kernel."""
         grad_fn = jax.value_and_grad(logdensity_fn)
@@ -142,7 +142,7 @@ def build_kernel():
 
 
 def as_top_level_api(
-    logdensity_fn: Callable, step_size: float, inverse_mass_matrix: jnp.Array
+    logdensity_fn: Callable, step_size: float, inverse_mass_matrix: Array
 ) -> SamplingAlgorithm:
     """Implements the (basic) user interface for the Barker's proposal :cite:p:`Livingstone2022Barker` kernel with a
     Gaussian base kernel.
--- a/blackjax/mcmc/barker.py
+++ b/blackjax/mcmc/barker.py
@@ -224,52 +224,6 @@ def as_top_level_api(
     return SamplingAlgorithm(init_fn, step_fn)
 
 
-def _barker_sample_nd(key, mean, a, scale, metric):
-    """
-    Sample from a multivariate Barker's proposal distribution. In 1D, this has the following probability density function:
-
-    .. math::
-        p(x; \\mu, a, \\sigma) = 2 \frac{N(x; \\mu, \\sigma^2)}{1 + \\exp(-a (x - \\mu)}
-
-    where :math:`N(x; \\mu, \\sigma^2)` is the normal distribution with mean :math:`\\mu` and standard deviation :math:`\\sigma`.
-    The multivariate Barker's proposal distribution is the product of one-dimensional Barker's proposal distributions.
-
-
-    Parameters
-    ----------
-    key
-        A PRNG key.
-    mean
-        The mean of the normal distribution, an Array. This corresponds to :math:`\\mu` in the equation above.
-    a
-        The parameter :math:`a` in the equation above, an Array. This is a skewness parameter.
-    scale
-        The global scale, a scalar. This corresponds to :math:`\\sigma` in the equation above.
-        It encodes the step size of the proposal.
-    metric
-        A `metrics.MetricTypes` object encoding the mass matrix information.
-
-    Returns
-    -------
-    A sample from the Barker's multidimensional proposal distribution.
-
-    """
-
-    key1, key2 = jax.random.split(key)
-    z = scale * jax.random.normal(key1, shape=mean.shape)
-    c = metric.scale(mean, a, False, True)
-
-    # Sample b=1 with probability p and 0 with probability 1 - p where
-    # p = 1 / (1 + exp(-a * (z - mean)))
-    log_p = -_log1pexp(-c * z)
-    b = jax.random.bernoulli(key2, p=jnp.exp(log_p), shape=mean.shape)
-
-    # return mean + z if b == 1 else mean - z
-    return jax.tree_util.tree_map(
-        lambda a, b: a + b, mean, metric.scale(mean, b * z - (1 - b) * z, False, False)
-    )
-
-
 def _barker_sample(key, mean, a, scale, metric):
     r"""
     Sample from a multivariate Barker's proposal distribution for PyTrees.
@@ -289,10 +243,20 @@ def _barker_sample(key, mean, a, scale, metric):
         A `metrics.MetricTypes` object encoding the mass matrix information.
     """
 
-    flat_mean, unravel_fn = ravel_pytree(mean)
-    flat_a, _ = ravel_pytree(a)
-    flat_sample = _barker_sample_nd(key, flat_mean, flat_a, scale)
-    return unravel_fn(flat_sample)
+    key1, key2 = jax.random.split(key)
+    flat_mean, _ = ravel_pytree(mean)
+
+    c = metric.scale(mean, a, False, True)
+    z = scale * jax.random.normal(key1, shape=flat_mean.shape)
+
+    # Sample b=1 with probability p and 0 with probability 1 - p where
+    # p = 1 / (1 + exp(-a * (z - mean)))
+    log_p = jax.tree_util.tree_map(lambda x, y: -_log1pexp(-x * y), c, z)
+    b = jax.random.bernoulli(key2, p=jnp.exp(log_p), shape=flat_mean.shape)
+
+    return jax.tree_util.tree_map(
+        lambda a, b: a + b, mean, metric.scale(mean, b * z - (1 - b) * z, False, False)
+    )
 
 
 def _log1pexp(a):
--- a/blackjax/mcmc/barker.py
+++ b/blackjax/mcmc/barker.py
@@ -186,7 +186,7 @@ def as_top_level_api(
     logdensity_fn
         The log-density function we wish to draw samples from.
     step_size
-        The value to use for the step size in the symplectic integrator.
+        The value of the step_size correspnoding to the global scale of the proposal distribution.
     inverse_mass_matrix
         The inverse mass matrix to use for pre-conditioning (see Appendix G of :cite:p:`Livingstone2022Barker`).
 
@@ -231,7 +231,8 @@ def _barker_sample_nd(key, mean, a, scale, C_t):
         The global scale, a scalar. This corresponds to :math:`\\sigma` in the equation above.
         It encodes the step size of the proposal.
     C_t
-        The transpose of the sqrt of the mass matrix, an Array. It is not used in the 1D version of Barker's proposal and thus not present in the equation above.
+        The transpose of the sqrt of the mass matrix, an Array. It is not used in the
+        1D version of Barker's proposal and thus not present in the equation above.
 
     Returns
     -------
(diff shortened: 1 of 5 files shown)


ismael-mendoza:
@AdrienCorenflos I just added a test to specifically check the new implementation with the mass matrix and fixed all the tests that use Barker. I'm open to suggestions on the new test or on any other changes. Thank you!

AdrienCorenflos:
Thanks for this. I think I now know how we can use general `Metrics` here. This requires a small addition to the `Metric` interface, whereby a metric would now also have a `scale` function argument, with signature
```python
# Callable[[ArrayLikeTree, Arr], ArrayLikeTree]
def scale(position: ArrayLikeTree, vector: ArrayLikeTree, inv=False) -> ArrayLikeTree:
    mass_matrix = mass_matrix_fn(position)
    # ravel everything etc
    if inv:
        return triangular_solve(mass_matrix_sqrt, vector)  # plus unravelling of course
    return mass_matrix_sqrt @ vector  # plus unravelling of course
```

@junpenglao would that be ok to add a function like this (essentially scaling a vector by the `inv_sqrt_mass` defined by the metric at position x).

If this is ok, then we can implement Barker's proposal with any sort of metric, even Riemannian, for cheap: 
https://github.com/ismael-mendoza/blackjax/blob/a61af36ab77dc9926ce2b54cdb3d25f6ecb05dd4/blackjax/mcmc/barker.py#L93-L95 would become
```python 
z = metric.scale(y, y, True) - metric.scale(x, x, True)  # Be careful here, it may be metric.scale(x, y, True) - metric.scale(y, x, True), need to check the detailed balanced condition properly and same below.
c_x = metric.scale(x, log_x)
c_y = metric.scale(y, log_y)
```
And something similar here
https://github.com/ismael-mendoza/blackjax/blob/a61af36ab77dc9926ce2b54cdb3d25f6ecb05dd4/blackjax/mcmc/barker.py#L245-L253

There would be some bookkeeping about shapes and all but these seem to be the only change.

@ismael-mendoza @junpenglao what do you think? It;s

AdrienCorenflos:
This is not even covered in the papers by Livingstone and Zanella: it's a good example of why Blackjax composability is cool :D

tests/mcmc/test_barker.py
```diff
@@ -51,5 +61,69 @@ def _test_samples_vs_pdf(samples, pdf):
...
+            rng_key=inference_key,
+            initial_state=state2,
+            inference_algorithm=barker2,
+            transform=lambda state, info: state.position,
+            num_steps=1000,
+        )
+
+        # states should be the exact same with same random key after transforming
+        states2_trans = C_t.dot(states2.T).T
+        np.testing.assert_allclose(states1, states2_trans, atol=1e-2)
```
AdrienCorenflos, line 125: Does this pass with higher precision -> this should really hold up to numerical precision.

blackjax/mcmc/barker.py
```diff
@@ -81,44 +81,57 @@ def build_kernel():
...
-        ratio_proposal = sum(tree_leaves(ratios_proposals))
         return proposal.logdensity - state.logdensity + ratio_proposal
 
     def kernel(
-        rng_key: PRNGKey, state: BarkerState, logdensity_fn: Callable, step_size: float
+        rng_key: PRNGKey,
+        state: BarkerState,
+        logdensity_fn: Callable,
+        step_size: float,
+        inverse_mass_matrix: Array,
```
AdrienCorenflos, line 109: Breaking change, this needs to default to previous behaviour.

blackjax/mcmc/barker.py
```diff
@@ -248,21 +266,47 @@ def _barker_sample(key, mean, a, scale):
...
-    flat_sample = _barker_sample_nd(key, flat_mean, flat_a, scale)
+    flat_sample = _barker_sample_nd(key, flat_mean, flat_a, scale, C_t)
     return unravel_fn(flat_sample)
 
 
 def _log1pexp(a):
     return jnp.log1p(jnp.exp(a))
 
 
+def _get_mass_matrix_sqrt(inverse_mass_matrix):
```
AdrienCorenflos, line 285: Duplicated code, hopefully if we push for the use of the Metric, this will fully live outside this file.

tests/mcmc/test_barker.py
ismael-mendoza: I will check, you are right

ismael-mendoza:
Thanks @AdrienCorenflos that seems like a good idea to me as it avoids the code redundancy and allows us to use Riemannian metrics for free like you pointed out. 

On your comment here 
```
z = metric.scale(y, y, True) - metric.scale(x, x, True)  # Be careful here, it may be metric.scale(x, y, True) - metric.scale(y, x, True), need to check the detailed balanced condition properly and same below.
```

do you have suggestions for how to check the correct detailed balanced condition? I'm not too familiar with this and I assume the Barker paper wouldn't mention it as they only show the algorithm for a fixed mass matrix. Thanks!

ismael-mendoza:
One quick follow up - am I missing something or would you need to implementations of `.scale` ? one for the `gaussian_euclidean` and one for the `gaussian_riemannian` metric?  The former one won't take the position?

AdrienCorenflos:
> One quick follow up - am I missing something or would you need to implementations of `.scale` ? one for the `gaussian_euclidean` and one for the `gaussian_riemannian` metric? The former one won't take the position?


We do need this (or similar) to be added I think, but let's wait for @junpenglao he will likely have a better opinion than me on the right API/design choice for these as he has thought about it a lot more than I have.

junpenglao:
> > One quick follow up - am I missing something or would you need to implementations of `.scale` ? one for the `gaussian_euclidean` and one for the `gaussian_riemannian` metric? The former one won't take the position?
> 
> We do need this (or similar) to be added I think, but let's wait for @junpenglao he will likely have a better opinion than me on the right API/design choice for these as he has thought about it a lot more than I have.

Yes sounds like a great addition to `metric`, the API makes sense as well.

ismael-mendoza:
Hello @AdrienCorenflos and @junpenglao, 

Thanks @AdrienCorenflos for implemeting the metric scaling. Now I can proceed finishing this PR, I just had two remaining questions). 

First, I noticed section 6.1 of the barker [paper](https://arxiv.org/pdf/1908.11812) that the authors use the proposed adaptation of Andrieu and Thoms (2008) for their experiments. I think it would be nice to have the full version of their algorithm blackjax and I wondering if this adaptation has already been implemented somewhere for blackjax? or maybe there is already some clear alternative that should be used to adapt hmc algorithms? It's unclear to me if the `window_adaptation` function that is used commonly in NUTS would be appropriate here for example. 

Second, @AdrienCorenflos I was just wondering if you finished the derivation above and if there is some additional term that must be added? 

Thank you both for your help!

AdrienCorenflos:
Oh right, I had completely forgotten about this bit. I'm sure I've got the derivation on a piece of paper *somewhere* 😬
Let me check tomorrow morning.

AdrienCorenflos:
Regarding the adaptation bit, let's check when that's done

AdrienCorenflos:
Actually, I think the easiest is if you implement Barker for the standard fixed mass matrix using this new scale thing and then I'll adapt it to the manifold case

ismael-mendoza:
Hi @AdrienCorenflos I noticed something strange in the `_format_covariance` fnc that tripped me up when adding the scaling: 

```
if is_inv:
    inv_cov_sqrt = jscipy.linalg.cholesky(cov, lower=True)
    cov_sqrt = jscipy.linalg.solve_triangular(
        inv_cov_sqrt, identity, lower=True, trans=True
    )
else:
    cov_sqrt = jscipy.linalg.cholesky(cov, lower=False).T
    inv_cov_sqrt = jscipy.linalg.solve_triangular(
        cov_sqrt, identity, lower=True, trans=True
    )
```
am I misunderstanding or the `inv_cov_sqrt` will be transposed in the case that `is_inv = False` and not tranposed in the case that `is_inv=True`? Is that the desired implementation? I was just surprised that it would be inconsistent like that but maybe it has to do with other algorithms?

AdrienCorenflos:
Yeah we have been discussing this with @junpenglao 
It's because the standard Euclidean metric passes in the inverse mass matrix, whereas the Riemannian uses the normal mass matrix... It's a bit inconsistent and we were going to look into making this coherent after this specific PR. For the time being, I'd recommend you make it work for the Euclidean in the new metric API and I'll take care of the Riemannian.

ismael-mendoza:
@AdrienCorenflos it required some finagling from what I said above but I think I was able to update my PR with the new metric.scale that at least works with the Euclidean metric. I also needed to add an option to transpose the matrix sqrt when scaling since the Barker algorithm required both (tranposed and untransposed) as far as I could tell. I added it to the Eucliden, let me know if you want me to propagate to the Riemannian. 

Please let me know if you have any other suggestions.

blackjax/mcmc/barker.py
ismael-mendoza: resolved in last commit, removed this duplicated code

blackjax/mcmc/barker.py
ismael-mendoza: made the `inverse_mass_matrix` an optional argument in altest commit

tests/mcmc/test_barker.py
ismael-mendoza: removed the tolerance in the latest commit so now it's default jax.numpy tolerance

AdrienCorenflos:
Yes please, APIs should mirror each other.

ismael-mendoza:
@AdrienCorenflos added it there as well

AdrienCorenflos:
It all seems to work for the standard pre-conditioning, but I'm seeing a lack of invariance for the Riemannian metric:

image

I'll fix, FYI, it comes from the fact that the proposal is not symmetric in the Gaussian anymore, so you have to compute the pdf

AdrienCorenflos (commented):
I've made changes so that the proposal works with Riemannian metrics.
There is one last thing to fix and then we're good and I'll leave it to you @ismael-mendoza  

-> the barker_nd sampling function was meant to take fully flat vectors, and here you pass it the metric which acts on trees. You are not seeing issues because your tests do not use nested trees, but this will fail. Can you modify the logic so that it simply does everything using metric on the non-flat vectors? This may require changing a few tests on the function here and there but should be easy

AdrienCorenflos (changes requested):
I've made changes so that the proposal works with Riemannian metrics.
There is one last thing to fix and then we're good and I'll leave it to you @ismael-mendoza  

-> the barker_nd sampling function was meant to take fully flat vectors, and here you pass it the metric which acts on trees. You are not seeing issues because your tests do not use nested trees, but this will fail. Can you modify the logic so that it simply does everything using metric on the non-flat vectors? This may require changing a few tests on the function here and there but should be easy

ismael-mendoza:
Thanks @AdrienCorenflos 

```
an you modify the logic so that it simply does everything using metric on the non-flat vectors?
 ```

just to make sure I understand, do you want me to remove `metric` from `_barker_sample_nd` entirely and only pass in flat vectors to this function? and the metric logic should live in `_barker_sample` ?

**Comment by AdrienCorenflos** [comment]
Well, I think you can just delete the nd version and do everything inside
the `_barker_sample`. It made sense when there was a pure flat logic, but
it's not the case anymore

**Comment by ismael-mendoza** [comment]
oh I think I understand your suggestion now, will modify and commit soon

**Comment by AdrienCorenflos** [comment]
Thanks, I'll check tomorrow. @junpenglao given I've touched that pr if you could also have a quick glance that would be useful

File: blackjax/mcmc/metrics.py [hunk 4]
```diff
@@ -62,7 +62,11 @@ def __call__(
 
 class Scale(Protocol):
     def __call__(
-        self, position: ArrayLikeTree, element: ArrayLikeTree, inv: ArrayLikeTree
+        self,
+        position: ArrayLikeTree,
+        element: ArrayLikeTree,
+        inv: ArrayLikeTree,
+        trans: ArrayLikeTree,
```
junpenglao, line 69: Let's turn this to required kwarg:
```suggestion
        *,
        inv: bool,
        trans: bool,
```

blackjax/mcmc/metrics.py
```diff
@@ -209,11 +218,23 @@ def scale(
...
+            )
+
+        def _linear_map():
+            return jax.lax.cond(
+                inv,
+                lambda: linear_map(inv_mass_matrix_sqrt, ravelled_element),
+                lambda: linear_map(mass_matrix_sqrt, ravelled_element),
+            )
+
+        scaled = jax.lax.cond(trans, _linear_map_transpose, _linear_map)
```
junpenglao, line 236: I think python if/else here also works as `inv` and `trans` are input as static parameters
```suggestion
        if inv:
            left_hand_side_matrix = inv_mass_matrix_sqrt
        else:
            left_hand_side_matrix = mass_matrix_sqrt
        if trans:
            left_hand_side_matrix = left_hand_side_matrix.T

        scaled = linear_map(left_hand_side_matrix, ravelled_element)
```

junpenglao (commented):
Some minor nit, could you try to see if the suggested changes works?

ismael-mendoza:
@junpenglao thanks for the suggestion, I think works but I see that one of the HMC benchmarks had a regression. Although it's unclear to me how that might have happened after just implementing your suggestion

junpenglao:
> @junpenglao thanks for the suggestion, I think works but I see that one of the HMC benchmarks had a regression. Although it's unclear to me how that might have happened after just implementing your suggestion

Yeah the regression is not related, i will investigate.

junpenglao:
@AdrienCorenflos any more comments? Otherwise LGTM.

AdrienCorenflos:
Nope, I had no additional comments when I asked for your opinion :)

junpenglao (approved):

junpenglao:
Thank you @ismael-mendoza !!

PR #733, opened 2024-09-06
Implement metric scaling

AdrienCorenflos:
# Thank you for opening a PR!

This is a change that relates to the PR https://github.com/blackjax-devs/blackjax/pull/731.
The idea is to allow metrics to scale a vector (possibly momentum, possibly otherwise) by the cholesky of the mass matrix or its inverse.

DRAFT PR so far, taking comments on interface and tentative implementation. Nothing is tested yet.

A comment: for efficiency reasons (in the context of Riemannian metrics), I have allowed the scale function to take a *sequence* of elements to rescale, so as to avoid recomputing the mass matrix all the time. There may be a cleaner way to do this but I can't see it.


AdrienCorenflos:
@ismael-mendoza this is what you need right?

blackjax/mcmc/metrics.py
```diff
@@ -283,4 +287,71 @@ def is_turning(
...
+            for _, element in inv_elements:
+                rav_element, unravel_fn = ravel_pytree(element)
+                rav_element = matmul(inv_sqrt_mass_matrix, rav_element)
+                scaled_elements.append(unravel_fn(rav_element))
+
+        scaled_elements = [scaled_elements[k] for k in argsort]
+
+        return tuple(scaled_elements)
+
     return Metric(momentum_generator, kinetic_energy, is_turning)
```
junpenglao, line 331:
```suggestion
    return Metric(momentum_generator, kinetic_energy, is_turning, scale)
```

junpenglao:
> A comment: for efficiency reasons (in the context of Riemannian metrics), I have allowed the scale function to take a sequence of elements to rescale, so as to avoid recomputing the mass matrix all the time. There may be a cleaner way to do this but I can't see it.

I see... I dont love it but I also dont hate it LOL

ismael-mendoza:
Thanks @AdrienCorenflos I think this will work great, happy to incoporate this into #731 once it's merged!

AdrienCorenflos:
I am a bit unhappy with one design choice: for the function scale to *mean* the same thing for all metrics I had to treat inputs differently: this comes from the fact that euclidean metrics take an inverse mass_matrix, while Riemannian take the mass matrix... I'm not sure what the rationale was behind this, but we will need to rethink it down the line. Maybe Riemannian should also take the inverse mass matrix? Either way, not a problem for this PR.

junpenglao:
A likely over-engineered solution here is to create a new covariance matrix class that has both original and inverse form, kind of like what TFP does for Gaussian Process https://github.com/tensorflow/probability/blob/main/tensorflow_probability/python/math/psd_kernels/positive_semidefinite_kernel.py

AdrienCorenflos:
I changed the api a bit -> you now only give scale(position, vector, inv) rather than something complicated. This is because I remembered we are using JAX and I shouldn't reinvent the wheel: if people want several scalings they should use vmap and that's it.

AdrienCorenflos:
@junpenglao any comments or go?

junpenglao:
Let me take another look tomorrow

blackjax/mcmc/metrics.py
```diff
@@ -30,21 +30,22 @@
 """
 from typing import Callable, NamedTuple, Optional, Protocol, Union
 
+import jax
 import jax.numpy as jnp
 import jax.scipy as jscipy
+from chex import Numeric
```
junpenglao, line 36: Let's add the type into https://github.com/blackjax-devs/blackjax/blob/main/blackjax/types.py instead, copying the 2 lines from chex

https://github.com/google-deepmind/chex/blob/343d03af64d0ead945fdda194b9d5c5111eb31d7/chex/_src/pytypes.py#L53-L54

reason: chex usage should be test only.

blackjax/mcmc/metrics.py
```diff
@@ -64,6 +65,7 @@ class Metric(NamedTuple):
     sample_momentum: Callable[[PRNGKey, ArrayLikeTree], ArrayLikeTree]
     kinetic_energy: KineticEnergy
     check_turning: CheckTurning
+    scale: Callable[[ArrayLikeTree, ArrayLikeTree, bool], ArrayLikeTree]
```
junpenglao, line 68: turn in to a Protocal as well? That way we can see also the names

blackjax/mcmc/metrics.py
```diff
@@ -196,39 +171,60 @@ def is_turning(
...
 
         # rho = m_sum
         rho = m_sum - (m_right + m_left) / 2
         turning_at_left = jnp.dot(velocity_left, rho) <= 0
         turning_at_right = jnp.dot(velocity_right, rho) <= 0
         return turning_at_left | turning_at_right
 
-    return Metric(momentum_generator, kinetic_energy, is_turning)
+    def scale(
+        position: ArrayLikeTree, element: ArrayLikeTree, inv: ArrayLikeTree
```
junpenglao, line 184:
```suggestion
        position: ArrayLikeTree, element: ArrayLikeTree, inv: bool
```

Same below

blackjax/mcmc/metrics.py
```diff
@@ -283,4 +272,69 @@ def is_turning(
...
+        scaled_elements
+            The scaled elements.
+        """
+        mass_matrix = mass_matrix_fn(position)
+        mass_matrix_sqrt, inv_mass_matrix_sqrt, diag = _format_covariance(
+            mass_matrix, is_inv=False
+        )
+        ravelled_element, unravel_fn = ravel_pytree(element)
+        scaled = jax.lax.cond(
+            inv,
```
junpenglao, line 296: Should always be false for Reinmannian kinetic energy for now?

AdrienCorenflos: No it's a different inv -> the is_inv above is about wether the mass matrix was given in inverse form, this one is about wether your want to scale the vector by the inverse of the cholesky or the cholesky.

blackjax/mcmc/metrics.py
AdrienCorenflos: Below where please?

blackjax/mcmc/metrics.py
AdrienCorenflos: Ok

blackjax/mcmc/metrics.py
AdrienCorenflos: Ok

blackjax/mcmc/metrics.py
```diff
@@ -283,4 +272,69 @@ def is_turning(
         # turning_at_right = jnp.dot(velocity_right, rho) <= 0
         # return turning_at_left | turning_at_right
 
-    return Metric(momentum_generator, kinetic_energy, is_turning)
+    def scale(
+        position: ArrayLikeTree, element: ArrayLikeTree, inv: ArrayLikeTree
```
junpenglao, line 276:
```suggestion
        position: ArrayLikeTree, element: ArrayLikeTree, inv: bool
```

AdrienCorenflos:
That should be good now @junpenglao 
Sorry about the hold up @ismael-mendoza but I think this is for a good reason :D

junpenglao (approved):

junpenglao:
Great stuff, thank you!

Issue #734, opened 2024-09-07
Schrodinger Follmer test is too slow.


AdrienCorenflos:
### Describe the issue as clearly as possible:

Schrodinger Follmer test is too slow. This will need to be sped up.

### Steps/code to reproduce the bug:

```python
N/A
```

### Expected result:

```shell
Speed up
```

### Blackjax/JAX/jaxlib/Python version information:

```python
N/A
```

PR #741, opened 2024-09-23
Speed up Schrodinger Follmer test

AdrienCorenflos:
Mostly reduce the data size -> there is no need for so much.

See
https://github.com/blackjax-devs/blackjax/issues/734


junpenglao (approved):

AdrienCorenflos:
Testing in production is the only way to test

Issue #278, opened 2022-09-16
Add nested Rhat diagnostic


rlouf:
https://arxiv.org/abs/2110.13017

twiecki:
I think the more appropriate place for this would be arviz.

junpenglao:
In session 6 of the paper there is some proposal of using nR to develop adaptive warmup length - which to me is a strong argument of having it in BlackJAX (and other sampling related functionary).

PR #752, opened 2024-10-30
ENH: Nested rhat MCMC diagnostic

giladturok:
Implement nested R-hat for Markov chain Monte Carlo (MCMC) diagnostic.

The potential scale reduction factor, also known as R-hat, is a popular MCMC diagnostic from [Gelman and Rubin](https://wwwusers.ts.infn.it/~milotti/Didattica/Bayes/pdfs/Gelman&Rubin_1992.pdf).
R-hat detects convergence of MCMC chains by comparing within chain variance to between chain variance.

Nested r-hat from [Margossian et al.](https://arxiv.org/pdf/2110.13017) better predicts convergence when running thousands of _short_ chains on modern hardware. Nested r-hat uses _superchains_, collections of MCMC chains, and compares within and between chain and superchain variance.

I am seeking feedback on the code style + API design. The code is somewhat complicated by requiring `input_array` to have 4 dimensions -- `num_superchains`, `num_chains`, `num_samples`, and `num_params` -- where most users may expect only 3 (or 2). Tests are also still needed, as well as a brief doc explanation of the math.

Quick nit: why does the existing R-hat function return the potential scale factor after _flattening_ along the sample and chain dimensions? I followed this convention for my implementation.

Addresses issue #278 .

### DIFF
--- a/blackjax/diagnostics.py
+++ b/blackjax/diagnostics.py
@@ -19,7 +19,7 @@ from scipy.fftpack import next_fast_len  # type: ignore
 
 from blackjax.types import Array, ArrayLike
 
-__all__ = ["potential_scale_reduction", "effective_sample_size"]
+__all__ = ["potential_scale_reduction", "nested_rhat", "effective_sample_size"]
 
 
 def potential_scale_reduction(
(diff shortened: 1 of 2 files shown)


gil2rok:
@charlesm93: you may be interested in this.

(For everyone else: Charles is the author of the nested R-hat paper.)

gil2rok:
Code style tests are failing because Flake8 is finding extra spaces around operators on line 43 of smc/resampling:

Screenshot 2024-10-30 at 4 22 19 PM

However I do not see what the problem is. Here's the line: https://github.com/blackjax-devs/blackjax/blob/65ae00e95aa122a581caef2d6eb1949cafbf8905/blackjax/smc/resampling.py#L43

If anyone sees how to fix this issue, please let me know.

blackjax/diagnostics.py
```diff
@@ -75,6 +75,66 @@ def potential_scale_reduction(
...
+        The axis indicating the multiple chains. Default to 1.
+    sample_axis
+        The axis indicating a single chain of MCMC samples. Default to 2.
+    
+    Returns
+    -------
+    NDArray of the resulting statistics (r-hat), with the chain and sample dimensions squeezed.
+
+    """
+    assert input_array.ndim == 4, "The input array must have 4 dimensions."
```
junpenglao, line 103: You should relax the ndim, as our input could have multiple dimensions of event shape (ie the random variable is non-scaler).

junpenglao (commented):
Thanks! Great start. Let me know when you add some test.

blackjax/diagnostics.py
junpenglao: Use `keepdims=True` and it should works.

junpenglao:
> Quick nit: why does the existing R-hat function return the potential scale factor after flattening along the sample and chain dimensions? I followed this convention for my implementation.

The current r-hat function does not flatten the sample, but rather squeeze it, so if you have a random variable with shape=(2, 5), the output result could be
shape=(1, 1, 2, 5) or (1, 2, 5, 1)
doing a squzze makes it return rhat the same shape as the random variable.

gil2rok:
Thanks so much for the quick feedback! I'll continue working on this next week.

PR #765, opened 2025-01-15
SMC Pretuning

ciguaran:
Implements pretuning of SMC from section 3.3 https://arxiv.org/abs/1808.07730 . The approach is complementary to the tuning of inner kernels we already have in Blackjax.

Pretuning works by holding (a set of samples from) a probability distribution of the parameters of the inner chains, and updating such distribution on each SMC step. The update works by taking an MCMC step and measuring chain mixing. Chain mixing scores are then used as weights to resample from the jitted parameter samples. The particles from pretuning are discarded, and then a full mutation step happens, using the new parameters.


blackjax/smc/pretuning.py
```diff
@@ -0,0 +1,343 @@
...
+    """Implements ESJD (expected squared jumping distance). Inner Mahalanobis distance
+    is computed using the Cholesky decomposition of M=LLt, and then inverting L.
+    Whenever M is symmetrical definite positive then it must exist a Cholesky Decomposition.
+    For example, if M is the Covariance Matrix of Metropolis-Hastings or
+    the Inverse Mass Matrix of Hamiltonian Monte Carlo.
+    """
+    L = jnp.linalg.cholesky(m)
+
+    def measure(previous_position, next_position, acceptance_probability):
+        return acceptance_probability * jnp.power(
```
junpenglao, line 38: Please unpack this nested statement for better readability.

blackjax/smc/pretuning.py
```diff
@@ -0,0 +1,343 @@
...
+
+def update_parameter_distribution(
+    key,
+    previous_param_samples: ArrayLikeTree,
+    previous_particles,
+    latest_particles,
+    measure_of_chain_mixing,
+    alpha,
+    sigma_parameters: ArrayLikeTree,
+    acceptance_probability,
```
junpenglao, line 63: if possible please typehint on all inputs.

junpenglao (approved):
Great stuff, thanks!

junpenglao (approved):

PR #776, opened 2025-02-12
SMC: Joint tuning and pretuning

ciguaran:
- Right now we have two mechanisms for parameter update within SMC: tuning and pretuning. This PR adds the possibility of using both at the same time. 
- Renames and modifies mass_matrix_from_particles, since we actually care about the inverse one.
- Extracts default measure factory to add for some user friendly error messages

### DIFF
--- a/blackjax/smc/inner_kernel_tuning.py
+++ b/blackjax/smc/inner_kernel_tuning.py
@@ -1,5 +1,7 @@
 from typing import Callable, Dict, NamedTuple, Tuple
 
+import jax
+
 from blackjax.base import SamplingAlgorithm
 from blackjax.smc.base import SMCInfo, SMCState
 from blackjax.types import ArrayTree, PRNGKey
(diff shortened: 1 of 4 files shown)


blackjax/smc/inner_kernel_tuning.py
```diff
@@ -28,8 +30,11 @@ def build_kernel(
     mcmc_step_fn: Callable,
     mcmc_init_fn: Callable,
     resampling_fn: Callable,
-    mcmc_parameter_update_fn: Callable[[SMCState, SMCInfo], Dict[str, ArrayTree]],
+    mcmc_parameter_update_fn: Callable[
+        [PRNGKey, SMCState, SMCInfo], Dict[str, ArrayTree]
+    ],
     num_mcmc_steps: int = 10,
+    smc_returns_state_with_parameter_override=False,
```
junpenglao, line 37: Is it possible to automatically detect this using the return from `mcmc_parameter_update_fn`

ciguaran: it would be both the input and the return of smc_algorithm().step(), but we need to know them before executing a step (since the input is adapted too). 
I thought about having a decorator based registry of SMC implementations that tune MCMC parameters, so we would check belonging to that registry, but the problem is that any partial application or wrapper wouldn't work out of the box, the user would need to register it.

ciguaran: In other words I don't know how to change while keeping the functional approach and without having the users annotating their functions somehow.

junpenglao: thanks Carlos, I cannot think of a better alternative as well.

PR #784, opened 2025-04-09
Energy error monitoring

reubenharry:


junpenglao (approved):

junpenglao:
Seems we need to fix dependency (JaxOpt is breaking our test)

PR #803, opened 2025-11-17
Improve the robustness of ChEES-HMC.

pipme:
I observed that the existing ChEES-HMC implementation is substantially less robust than the original version in TensorFlow Probability (TFP). This pull request introduces several improvements and bug fixes to the ChEES adaptation algorithm and its integration with the dynamic HMC kernel, as well as enhancements to the corresponding tests. I took [TFP's implementation](https://github.com/tensorflow/probability/blob/main/tensorflow_probability/python/experimental/mcmc/gradient_based_trajectory_length_adaptation.py) for reference.

The main changes focus on improving numerical stability, correctness of step size and trajectory length adaptation, and robustness of the adaptation process. Additionally, the test suite is strengthened to validate the algorithm's behavior better.

**ChEES Adaptation Algorithm Improvements:**
* Improved handling of non-finite values in the harmonic mean of acceptance probabilities to avoid propagation of NaNs or infinities during dual averaging.
* Added a numerically stable weighted empirical mean function and replaced the simple mean with a weighted mean for proposal positions to improve adaptation quality.
* Clipped log-trajectory-length updates and trajectory lengths to prevent excessively large or small values, and ensured minimum/maximum trajectory length bounds based on the step size and a configurable maximum number of leapfrog steps.

**Dynamic HMC Kernel Enhancements:**

* Added input validation to the Halton sequence generator to prevent integer overflow and clarified its documentation.

**Testing Improvements:**

* Updated the ChEES adaptation test: test the harmonic mean instead of the arithmetic mean of the acceptance rate, use the Adam optimizer which is closer to the [original paper](https://proceedings.mlr.press/v130/hoffman21a.html) and TFP's implementation, and added comprehensive checks for empirical mean and standard deviation of samples. 

---

One point I am not sure of is about the default learning rate and trajectory length gradient.  TFP’s default value is 0.025, but in this implementation, such a small value performs poorly; in contrast, learning rates of 0.25–0.5 yield more reliable adaptation in my experiments. TFP applies [trajectory-length updates multiplicatively](https://github.com/tensorflow/probability/blob/732e7d16235a38b639a3b2cd00e76a4add45118c/tensorflow_probability/python/experimental/mcmc/gradient_based_trajectory_length_adaptation.py#L1085-L1090), and the theoretical justification for this choice is not entirely clear to me. In BlackJAX, the gradient is [computed with respect to the **log**-trajectory length and the update is applied additively](https://github.com/pipme/blackjax/blob/25125ccb4d491690c4c089ce9af387a135a8762f/blackjax/adaptation/chees_adaptation.py#L199), which appears more principled. 

I may be able to test this revised algorithm on real-world problems in the future. Nevertheless, given the significant improvements in stability, I believe this pull request is ready to be merged.

### DIFF
--- a/blackjax/adaptation/chees_adaptation.py
+++ b/blackjax/adaptation/chees_adaptation.py
@@ -17,6 +17,10 @@ from blackjax.util import pytree_size
 
 # optimal tuning for HMC, see https://arxiv.org/abs/1001.4460
 OPTIMAL_TARGET_ACCEPTANCE_RATE = 0.651
+# Clip the final log-space update like the original implementation in TFP (~log(2)/2 ≈ 0.35).
+LOG_UPDATE_CLIP = 0.35
+# Small constant to avoid division by zero or log of zero
+EPS_FLOAT = 1e-20
 
 
 class ChEESAdaptationState(NamedTuple):
@@ -59,7 +63,7 @@ def weighted_empirical_mean(x, w):
 
     w_exp = w.reshape((w.shape[0],) + (1,) * (x.ndim - 1))
     num = jnp.sum(w_exp * x_safe, axis=0)
-    den = jnp.sum(w_exp, axis=0) + 1e-20
+    den = jnp.sum(w_exp, axis=0) + EPS_FLOAT
     return jax.lax.stop_gradient(num / den)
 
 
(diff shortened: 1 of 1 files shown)


junpenglao:
Great stuff, thank you!!!

PR #929, opened 2026-05-30
feat(smc): add batch_size for memory-bounded SMC execution via jax.lax.map

thomasckng:
## Description

Large-N SMC runs on GPUs can exhaust memory because `jax.vmap` over all N particles materialises the full particle array in VRAM simultaneously. This PR adds a `batch_size` parameter to every public SMC kernel and update strategy. When set to a positive integer, particle updates and weight computations are sequenced through `jax.lax.map(fn, xs, batch_size=B)` instead of `jax.vmap(fn)`, capping peak memory to `B` particles per operation while producing numerically identical results. The default `batch_size=0` preserves the original `jax.vmap` behaviour for all existing callers.

### Files changed

| Module | Change |
|---|---|
| `smc/base.py` — `update_and_take_last` | Added `batch_size` → switches MCMC kernel to `jax.lax.map` |
| `smc/waste_free.py` — `update_waste_free` | Added `batch_size` → switches waste-free MCMC kernel to `jax.lax.map` |
| `smc/from_mcmc.py` — `build_kernel` | Added `batch_size` → batches both MCMC update and weight computation |
| `smc/tempered.py` — `build_kernel`, `as_top_level_api` | Added `batch_size` |
| `smc/adaptive_tempered.py` — `build_kernel`, `as_top_level_api` | Added `batch_size` → also batches ESS log-likelihood evaluation in `compute_delta` |
| `smc/persistent_sampling.py` — `init`, `step`, `build_kernel`, `as_top_level_api` | Added `batch_size` → batches initial log-likelihood eval, MCMC update, and per-step log-likelihood eval |
| `smc/adaptive_persistent_sampling.py` — `build_kernel`, `as_top_level_api` | Added `batch_size` |

### Tests

- tests/smc/test_smc.py — equivalence tests for `update_and_take_last` and `update_waste_free`: exact match (`np.testing.assert_allclose`) between `batch_size=0` and `batch_size > 0` runs.
- `tests/smc/test_tempered_smc.py` — convergence tests for `tempered` and `adaptive_tempered` with `batch_size > 0`.
- `tests/smc/test_waste_free_smc.py` — convergence test for `waste_free_smc` with `batch_size > 0`.
- `tests/smc/test_persistent_sampling.py` — convergence tests for `persistent_sampling` and `adaptive_persistent_sampling` with `batch_size > 0`, plus `test_fixed_schedule_persistent_sampling_batch_equivalence`: an exact-match equivalence test.

## Checklist

**General**

**Code quality**

### DIFF
--- a/blackjax/smc/adaptive_persistent_sampling.py
+++ b/blackjax/smc/adaptive_persistent_sampling.py
@@ -33,7 +33,7 @@ def build_kernel(
     target_ess: float | Array,
     update_strategy: Callable = update_and_take_last,
     root_solver: Callable = solver.dichotomy,
-    particle_batch_size: int = 0,
+    batch_size: int = 0,
 ) -> Callable:
     """Build an adaptive Persistent Sampling kernel, with signature
     (rng_key,
@@ -78,9 +78,9 @@ def build_kernel(
     root_solver
         The solver used to adaptively compute the temperature given a target number
         of effective samples. By default, blackjax.smc.solver.dichotomy.
-    particle_batch_size: int, optional
+    batch_size: int, optional
         Number of particles processed per sequential batch when
-        ``particle_batch_size > 0``. Passed to the underlying
+        ``batch_size > 0``. Passed to the underlying
         ``persistent_sampling.build_kernel`` call to enable
         ``jax.lax.map``-based batching, reducing peak GPU memory. ``0``
         (default) keeps the original ``jax.vmap`` behaviour.
@@ -137,7 +137,7 @@ def build_kernel(
         mcmc_init_fn=mcmc_init_fn,
         resampling_fn=resampling_fn,
         update_strategy=update_strategy,
-        particle_batch_size=particle_batch_size,
+        batch_size=batch_size,
     )
 
     def kernel(
@@ -170,7 +170,7 @@ def as_top_level_api(
     num_mcmc_steps: int = 10,
     update_strategy: Callable = update_and_take_last,
     root_solver: Callable = solver.dichotomy,
-    particle_batch_size: int = 0,
+    batch_size: int = 0,
 ) -> SamplingAlgorithm:
     """
     Implements the user interface for the adaptive Persistent Sampling
@@ -223,9 +223,9 @@ def as_top_level_api(
     root_solver : Callable, optional
         The solver used to adaptively compute the temperature given a target
         number of effective samples. By default, blackjax.smc.solver.dichotomy.
-    particle_batch_size : int, optional
+    batch_size : int, optional
         Number of particles processed per sequential batch when
-        ``particle_batch_size > 0``. Uses ``jax.lax.map`` for the MCMC update
+        ``batch_size > 0``. Uses ``jax.lax.map`` for the MCMC update
         step, reducing peak GPU memory relative to a full ``jax.vmap``. ``0``
         (default) keeps the original ``jax.vmap`` behaviour.
 
@@ -254,13 +254,11 @@ def as_top_level_api(
         target_ess,
         update_strategy,
         root_solver,
-        particle_batch_size=particle_batch_size,
+        batch_size=batch_size,
     )
 
     def init_fn(position: ArrayLikeTree) -> persistent_sampling.PersistentSMCState:
-        return init(
-            position, loglikelihood_fn, max_iterations, batch_size=particle_batch_size
-        )
+        return init(position, loglikelihood_fn, max_iterations, batch_size=batch_size)
 
     def step_fn(
         rng_key: PRNGKey,
--- a/blackjax/smc/adaptive_persistent_sampling.py
+++ b/blackjax/smc/adaptive_persistent_sampling.py
@@ -33,6 +33,7 @@ def build_kernel(
     target_ess: float | Array,
     update_strategy: Callable = update_and_take_last,
     root_solver: Callable = solver.dichotomy,
+    particle_batch_size: int = 0,
 ) -> Callable:
     """Build an adaptive Persistent Sampling kernel, with signature
     (rng_key,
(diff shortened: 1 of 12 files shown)


junpenglao:
Nice, I took a quick look and it is mostly LGTM, I like that the test has direct comparison between batch=0 and batch>0. I think there is a few API missing coverage, so i will let Claude run it through the PR

junpenglao (changes requested):
_This is BlackJAX-dev AI Tech Lead (Claude-powered) 🤖._

Thanks for this @thomasckng — really clean contribution, and an opt-in `batch_size=0` default that leaves the `vmap` path untouched is exactly the right call for backward compat. I especially like the direct `batch=0` vs `batch>0` equivalence test, and the honest note about routing around resampling FP-divergence with a Gaussian setup. Ran it through on my end — the design is sound; **a few things I'd like fixed before LGTM**, mostly around API coverage:

**1. `update_strategy` receives `batch_size` unconditionally.** `from_mcmc.build_kernel` ([`from_mcmc.py#L97-L103`](https://github.com/blackjax-devs/blackjax/blob/fb7d6d80940935c12d34f6051b6251cccb9414e4/blackjax/smc/from_mcmc.py#L97-L103)) and `persistent_sampling.build_kernel` ([`persistent_sampling.py#L639-L645`](https://github.com/blackjax-devs/blackjax/blob/fb7d6d80940935c12d34f6051b6251cccb9414e4/blackjax/smc/persistent_sampling.py#L639-L645)) now always pass `batch_size=...` into `update_strategy(...)`, even at the default `0`. Since `update_strategy` is a public extension point, any user-defined strategy that doesn't accept the kwarg will start raising `TypeError` without ever opting into the feature. Could we forward it only when set — e.g. `**({"batch_size": batch_size} if batch_size else {})` — so the existing strategy contract is preserved for everyone on the default path?

**2. `partial_posteriors_path` is left out.** Its `build_kernel` calls `smc_from_mcmc(...)` without `batch_size` ([`partial_posteriors_path.py#L73`](https://github.com/blackjax-devs/blackjax/blob/fb7d6d80940935c12d34f6051b6251cccb9414e4/blackjax/smc/partial_posteriors_path.py#L73)), so partial-posteriors users get neither the memory benefit nor the parameter. Since the motivation is large-scale OOM, let's plumb it through there too to keep coverage consistent. (`pretuning` / `tuning.from_particles` also `vmap` over particles, but those are cheap measurement paths — fine to leave.)

**3. The `vmap`-or-`lax.map` ternary repeats ~5×.** It shows up in `adaptive_tempered`, `from_mcmc`, and `persistent_sampling` (×2), plus the tuple-kernel variant in `base` / `waste_free`. Let's centralize it with a couple of small helpers in `smc/base.py` so the semantics live in one place:
```python
def maybe_batched_map(fn, batch_size):       # unary, over one pytree
    return (lambda xs: jax.lax.map(fn, xs, batch_size=batch_size)) if batch_size > 0 else jax.vmap(fn)

def maybe_batched_kernel(kernel, batch_size):  # n-ary, over a tuple pytree
    return (lambda *a: jax.lax.map(lambda t: kernel(*t), a, batch_size=batch_size)) if batch_size > 0 else jax.vmap(kernel)
```
This turns "batching bolted on in 7 places" into one primitive the SMC layer shares, and makes future kernels trivial to opt in.

**Minor nits (non-blocking):**
- `lambda args: mcmc_kernel(args[0], args[1], args[2])` → `lambda args: mcmc_kernel(*args)`.
- The equivalence-test docstring says "bit-identical log-likelihoods" but the assert uses `rtol=1e-5` — worth softening the wording to match.
- Confirmed `jax.lax.map(..., batch_size=N)` handles non-divisor and oversize `N` cleanly (checked `N=30` and `N=200` over 100 particles, both match `vmap` exactly), so no divisibility constraint — nice. A test with a non-divisor `batch_size` (say 7 over 20 particles) would lock that in.

Once 1–3 are addressed this is good to go — the three-layer threading (`build_kernel` + `as_top_level_api` at each level) is consistent and the core design is right. 🙌

thomasckng:
@junpenglao Thanks for the detailed review. I have addressed all comments (including the minor nits) in the new commits. Please let me know if you have any other comments.

junpenglao:
> _This is BlackJAX-dev AI Tech Lead (Claude-powered)._

Thanks @thomasckng — went through the five new commits and all the substantive points are addressed: the `update_strategy` kwarg is now forwarded only when set (clean default path restored), `partial_posteriors_path` carries `batch_size` through, the `map_fn` / `map_kernel` helpers are a nice consolidation, and the softened docstring + non-divisor test are in. Ran the batched SMC tests locally — all green. 🎉

One straggler before I flip to approve: [`waste_free.py#L51-L58`](https://github.com/blackjax-devs/blackjax/blob/82cc01b179c7253e9ba1b327488835401a07d7a8/blackjax/smc/waste_free.py#L51-L58) still has the old inline `if batch_size > 0 / lambda args: mcmc_kernel(args[0], args[1], args[2])` block — the exact tuple-kernel pattern `map_kernel` now exists for. Folding it in closes both the last DRY spot and the `args[0..2]` indexing nit in one edit:

```python
from blackjax.smc.base import map_kernel  # add to imports

    def update(rng_key, position, step_parameters):
        """
        Given the initial particles, runs a chain starting at each.
        The combines the initial particles with all the particles generated
        at each step of each chain.
        """
        states, infos = map_kernel(mcmc_kernel, batch_size)(
            rng_key, position, step_parameters
        )
```

(`import jax.lax` stays — it's still used by the `jax.lax.scan` above.) With that, this is good to go from my side. 🙌

junpenglao (approved):

thomasckng:
Sorry for the oversight. It's fixed now.

junpenglao:
Nice, thanks!

PR #943, opened 2026-06-23
Add slice sampling kernel

dirmeier:
## Description

Adds the Slice Sampler following the API from TFP. Discussed previously [here](https://github.com/blackjax-devs/blackjax/pull/514). 

## Checklist

**General**
- [x] The branch is rebased on the latest `main`
- [x] Commit messages are clear and descriptive
- [x] `pre-commit run --all-files` passes (black, isort, flake8, mypy)
- [x] Tests cover the changes (`mamba run -n blackjax python -m pytest tests/`)

**Code quality**
- [x] Public functions have docstrings following the [NumPy style guide](https://numpydoc.readthedocs.io/en/latest/format.html)
- [x] Naming follows existing conventions (`logdensity`, `jax.tree.map`, `jax.random.key()`, `jnp.clip(min=, max=)`)
- [x] All new code is JIT-compatible

**New sampler / algorithm** *(skip if not applicable)*
- [ ] There is an open issue discussing this algorithm (use the [sampler proposal template](https://github.com/blackjax-devs/blackjax/issues/new?template=sampler_proposal.yml))
- [x] Follows the three-layer pattern: `init` / `build_kernel` / `as_top_level_api`
- [x] Registered in `blackjax/__init__.py`
- [ ] An example notebook has been added or updated

Consider opening a **Draft PR** first if you want early feedback on the design.

PR #911, opened 2026-05-12
Nested Sampling

yallup:
## Description

Nested Sampling (NS) is a workhorse inference method in the physical sciences (cosmology, particle physics, gravitational waves, materials), where it is routinely used for both posterior characterisation and evidence-driven model comparison. Bringing a first-class NS implementation into blackjax lets existing analyses be benchmarked against blackjax's SMC alternatives under a shared, JIT-compatible API, and makes the method available to users who already build pipelines on top of blackjax.

This PR contributes three pieces, layered to match existing blackjax conventions:

1. **Nested Sampling kernel (`blackjax.ns`)** -- a base NS kernel and an adaptive variant, with `from_mcmc` patterns mirroring `blackjax.smc`'s inner-kernel composition. The interface is sampler-agnostic so any constrained inner kernel can drive the live-point update.

2. **Slice Sampling (`blackjax.mcmc.ss`)** -- a standalone hit-and-run slice sampler (HRSS) implementing Neal's stepping-out and rejection with contraction. Usable on its own as an MCMC kernel, and serves as the canonical inner kernel for NS.

3. **Nested Slice Sampling (NSS, `blackjax.ns.nss`)** -- the practical synthesis: a performant constrained sampler wiring HRSS into the NS `from_mcmc` pattern. Matches the algorithm described in the companion paper [arXiv:2601.23252](https://arxiv.org/abs/2601.23252).

Sampling-book examples can follow this PR -- an NSS example extending the existing SMC material, and a standalone slice-sampling example.

This branch has been through extensive iteration; happy to squash changes to avoid git history bloat

## Related issues / discussions

Closes earlier PR #755 , builds on issue #753

## Checklist

**General**
- [x] The branch is rebased on the latest `main`
- [x] Commit messages are clear and descriptive
- [x] `pre-commit run --all-files` passes (black, isort, flake8, mypy)
- [ ] Tests cover the changes (`mamba run -n blackjax python -m pytest tests/`)

**Code quality**
- [x] Public functions have docstrings following the [NumPy style guide](https://numpydoc.readthedocs.io/en/latest/format.html)
- [x] Naming follows existing conventions (`logdensity`, `jax.tree.map`, `jax.random.key()`, `jnp.clip(min=, max=)`)
- [x] All new code is JIT-compatible

**New sampler / algorithm**
- [x] There is an open issue discussing this algorithm (use the [sampler proposal template](https://github.com/blackjax-devs/blackjax/issues/new?template=sampler_proposal.yml))
[...]

PR #946, opened 2026-06-27
Slice integration

yallup:
## Description

Consolidates slice sampling into a single, minimal **univariate spine**
parameterised by a pluggable *proposal generator*, following the design
discussion in #944. Every slice update is one-dimensional; all multivariate
behaviour comes from the proposal — `proposal_generator(rng_key, position,
logdensity_fn) -> slice_fn`, with `slice_fn(t) -> (state, is_valid)` — exactly
as `random_walk` is parameterised by its proposal.

This reconciles the two existing efforts (the merged coordinate-wise doubling
kernel, #943, and the hit-and-run stepping-out kernel from the nested-sampling
work, #911) and settles the questions raised in #944:

- **One spine, two entry points** — `blackjax.slice_sampling` (multivariate /
  hit-and-run) and `blackjax.coordinate_slice` (slice-within-Gibbs) share the
  same univariate core.
- **Seams passed as callables**, NUTS-integrator style (not strings):
  `interval=doubling` (Neal Fig. 4 + Fig. 6 acceptance) / `stepping_out`
  (Fig. 3); `coordinate_order=random_order` / `fixed_order`.
- **Direction `scale`** like `random_walk.normal(sigma)`:
  `direction_proposal(scale)` takes a scalar, vector, or dense `scale`
  (unit by default).
- **Constraints / nested sampling** (the "built-in vs override" question in
  #944) are kept *out* of the core. Because `slice_fn` returns
  `(state, is_valid)`, a downstream override of the proposal can gate a hard
  constraint (`is_valid`) and record extra quantities on the candidate state
  (e.g. a nested-sampling log-likelihood, computed once and threaded out, not
  recomputed); cheap constraint terms can short-circuit the likelihood. The
  coordinate sweep additionally takes an optional `constraint_fn`.
- **NUTS-style diagnostics** — `SliceState` / `SliceInfo` (`num_expansions`,
  `num_shrink`, bracket endpoints as position-shaped offset vectors).

- Design discussion: #944 (with @junpenglao)
- Builds on #943 (slice sampling kernel, @dirmeier)
- Supports #911 (Nested Sampling)

## Checklist

**General**

**Code quality**
- [x] Public functions have docstrings following the NumPy style guide
- [x] Naming follows existing conventions (`logdensity`, `jax.tree.map`, `jax.random.key()`, ...)

**New sampler / algorithm**
- [x] There is an open issue/discussion discussing this algorithm (#944)
- [x] Follows the three-layer pattern: `init` / `build_kernel` / `as_top_level_api`
- [x] Registered in `blackjax/__init__.py`

### DIFF
--- a/blackjax/__init__.py
+++ b/blackjax/__init__.py
@@ -128,6 +128,9 @@ adjusted_mclmc_dynamic = generate_top_level_api_from(_adjusted_mclmc_dynamic)
 adjusted_mclmc = generate_top_level_api_from(_adjusted_mclmc)
 elliptical_slice = generate_top_level_api_from(_elliptical_slice)
 slice_sampling = generate_top_level_api_from(_slice)
+coordinate_slice = GenerateSamplingAPI(
+    _slice.coordinate_slice, _slice.init, _slice.build_coordinate_kernel
+)
 ghmc = generate_top_level_api_from(_ghmc)
 barker = generate_top_level_api_from(_barker)
 barker_proposal = barker  # backwards-compatible alias
@@ -245,6 +248,7 @@ __all__ = [
     "barker",
     "elliptical_slice",
     "slice_sampling",
+    "coordinate_slice",
     "mclmc",
     "adjusted_mclmc",
     "adjusted_mclmc_dynamic",
--- a/blackjax/mcmc/slice.py
+++ b/blackjax/mcmc/slice.py
@@ -18,8 +18,7 @@ point. Multivariate behaviour is determined entirely by the proposal generator
 that produces the line, ``proposal_generator(rng_key, position, logdensity_fn)
 -> slice_fn`` with ``slice_fn(t) -> (state, is_valid)``. The candidate state is
 threaded straight through the kernel, so a proposal can record extra quantities
-on it (for example a nested-sampling log-likelihood) and consume a constraint
-through ``is_valid``.
+on it.
 
 Two samplers are built on this spine:
 
@@ -35,16 +34,10 @@ Two samplers are built on this spine:
 
 The one-dimensional interval is built by the stepping-out or doubling procedure
 of Neal (2003), passed as a callable (``interval=stepping_out`` or
-``interval=doubling``, the way NUTS takes ``integrator=velocity_verlet``), then
-narrowed by the shrinkage procedure. Doubling additionally applies the Fig. 6
-acceptance test. Constraints are not built in: they are added downstream by
+``interval=doubling``). Doubling additionally applies the Fig. 6 acceptance
+test. Additional constraints are not built in but added downstream by
 overriding the proposal, which gates on ``is_valid`` and may record extra
-quantities on the state (the pattern used by nested sampling). The coordinate
-sweep also accepts an optional ``constraint_fn``.
-
-This mirrors the layering of :mod:`blackjax.mcmc.random_walk`: a small core
-kernel parameterized by a proposal generator, with named convenience
-constructors on top.
+quantities on the state.
 
 References
 ----------
@@ -99,9 +92,6 @@ class SliceState(NamedTuple):
 class SliceInfo(NamedTuple):
     """Additional information on a Slice sampling transition.
 
-    This additional information can be used for debugging or computing
-    diagnostics.
-
     is_accepted
         Whether shrinkage found a valid point within ``max_shrinkage`` steps.
         Always ``True`` for an unconstrained target (the slice always contains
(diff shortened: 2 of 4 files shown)


junpenglao:
This is a clean execution of the #944 design — thank you for converging the two efforts so faithfully. I did an independent correctness re-review (focused on the rewrite, since it restructures the merged #943 doubling into the univariate spine) plus a standards/design pass.

**Correctness ✅ — detailed balance preserved across the rewrite.** Verified all six components against Neal (2003): vectorized `doubling` (≡ sequential; `_best_interval` tie-break checked algebraically), `_doubling_accept` (Fig. 6, exact in t-space), `stepping_out` (Fig. 3), the `_shrink` `is_valid`-gating (unconstrained = exact Fig. 5; constrained = a valid restriction to π_C ∝ π·I(x∈C) with reversibility intact, since the same `in_slice` drives both interval-finding and the Fig. 6 acceptance), the `_univariate_slice` spine, the coordinate sweep (scan carry + `widths[order]`/stitch correct), and hit-and-run. No correctness blockers.

A few items before merge:

**1. Modern union syntax (design guide §4).** `from typing import Optional, Union` + `Optional[Callable]` / `Union[float, Array]` (lines 48, 428, 430, 517, 531, 612, 615) → please switch to `Callable | None` and `float | Array` and drop the `Optional`/`Union` imports. Heads-up: `pyupgrade`/the style check doesn't rewrite this form, so it passed CI, but it's our convention.

**2. Test hardening (from the correctness review):**
- Add an **asymmetric/skewed 1D target** (e.g. exponential or gamma) to `MomentRecoveryTest`. Every current recovery target is a symmetric Gaussian — a reversed shrinkage direction would pass those but fail on a skewed target, so this closes a real reversibility-bug gap.
- `IntervalProcedureTest` checks the bracket-ends with `or` where `and` documents the intended invariant (passes trivially at the current `max_expansions`). Minor, worth tightening.

**3. Constraint-API symmetry (suggestion — your call).** Nice resolution overall: `constraint_fn` on `coordinate_slice` + proposal-override for hit-and-run. For API symmetry, would you consider also accepting `constraint_fn` on `slice_sampling` (folded internally into the default direction proposal), so both entry points expose the same clean convenience and proposal-override stays the power-user / nested-sampling path? Not a blocker — just consistency between the two entry points.

Really nice work. Once **1** (and ideally **2**) land, I think this is merge-ready.

yallup:
Thanks @junpenglao — addressed all four.

**1. Union syntax** ✅ — `Optional`/`Union` → `X | None` / `float | Array`, imports dropped.

**2a. Skewed recovery target** ✅ — added a 2D `Exp(1) ⊗ Exp(1)` target to `MomentRecoveryTest` (positive quadrant, marginal skew +2), over both intervals: asserts support, the first two moments, and **per-axis sample skew > 1** — a reversed shrinkage direction mirrors the density and flips that sign. Made it 2D rather than 1D so it also exercises the random-direction projection, not just the spine.

**2b. `IntervalProcedureTest` `or` → `and`** — kept `or`; `and` isn't actually the invariant. `doubling` extends one random side per step and `stepping_out` splits its budget across the two sides, so neither guarantees *both* ends exit within `max_expansions` (over 2000 seeds: both-out 0.99 / 0.39 for doubling / stepping-out; ≥1-out 1.00 / 1.00). `or` is the precise always-true invariant — kept it, and strengthened the test with `num_expansions > 0` plus a comment on why.

**3. Constraint-API symmetry** — resolved by unifying on one mechanism rather than adding `constraint_fn` to `slice_sampling`. `coordinate_slice` now routes each move through an overridable `axis_proposal` (default `coordinate_proposal`, a one-hot `flat[i] += t` step), exactly as `build_kernel` consumes its `proposal_generator`; `constraint_fn` is dropped from both coordinate entry points. One way to constrain anywhere in the family — override the proposal, gate `is_valid` — symmetric across both samplers:

```python
slice_sampling(logp,   proposal_generator=override)   # (rng, pos, logp)    -> slice_fn -> (state, is_valid)
coordinate_slice(logp, axis_proposal=axis_override)   # (rng, pos, logp, i) -> slice_fn -> (state, is_valid)
```

Defaults and reversibility unchanged (the default `coordinate_proposal` reproduces the prior coordinate move; the half-normal `ConstraintTest` exercises the override on both paths as the regression guard).

A few notes / follow-ups:
- **Defaults.** The top-level `slice_sampling` defaults to multivariate hit-and-run with the `doubling` interval. Both are open to change — e.g. defaulting to the coordinate-wise sweep (`coordinate_slice`) instead, and/or to `stepping_out`.
- **Naming.** Should the top-level just be `slice`? Other samplers are `hmc` / `nuts` / `mala`, so the trailing `_sampling` is a bit unusual for a sampler name — but `slice` is a Python built-in, so we kept `slice_sampling` to avoid the collision. Open to views.
- **Docs.** We'll add a sampling-book entry demonstrating this on Neal's funnel downstream of the merge (covers the outstanding example-notebook checklist item).

All green: `pre-commit` (black/isort/flake8/mypy) + the slice and api-protocol suites.

junpenglao:
Beautiful, nice work!!

Currently default is fine (multivariate hit-and-run and `slice_sampling`). And yes please add a sampling-book entry.

junpenglao (approved):

PR #947, opened 2026-06-28
Nested sampling

### LINKED ISSUE #753: Nested Sampling implementation
### Presentation of the new sampler

Nested sampling [[Skilling]](https://projecteuclid.org/journals/bayesian-analysis/volume-1/issue-4/Nested-sampling-for-general-Bayesian-computation/10.1214/06-BA127.full) is a particle monte carlo method typically styled as primarily targeting estimation of the marginal likelihood. It has been justified as a form of SMC [1805.03924](https://arxiv.org/abs/1805.03924), however we think it is in practice a rather singular form and requires some special implementation details hence the PR (although open to debate here). The basic construct is to scan in increasing likelihood levels while sampling from the truncated prior region at each step. It is very well established in the physical sciences [2205.15570](https://arxiv.org/abs/2205.15570), however with limited usage outside of this context. Here it is often noted that as an “athermal” method it sees phase transitions differently to typical methods based on some tempering scheme, and can be very profitable in those sorts of problems.

There are a number of different code implementations of the same generic algorithm, see table 2 [2205.15570](https://arxiv.org/abs/2205.15570) for a relatively up to date list with some specific implementation choices listed. Generally in mixes of fortran/python and C, limited consistent working examples that are compatible with modern python PPLs

The successful implementation of nested sampling is, in our opinion, pretty closely tied to slice sampling, so we have implemented that as well, and this can be extracted in a more abstract way if that is generically useful elsewhere.

### How does it compare to other algorithms in blackjax?

This fits naturally alongside the existing SMC methods in blackjax, and part of our motivation to push this to blackjax is so we can compare here in a like for like way. We have a working implementation already that gives consistent estimation of the marginal likelihood in 10-100 parameters fairly rapidly. In theory the design of SMC in blackjax could be followed, allowing a flexible choice of inner kernel, however generic application of many gradient based MCMC chains in nested sampling is difficult, so we have currently tied quite closely to slice sampling, this design choice could be lifted.

The rather specific requirements of the inner kernel when sampling from truncated regions, means that in our estimation nested sampling cannot be efficiently adapted to directly use the existing SMC abstractions in blackjax (you “see” both the likelihood and prior log prob in nested sampling as separate so the inner kernel construction is rather bespoke). However we tried to follow the SMC design patterns closely and if unification is possible here that would be worthwhile.

### Where does it fit in blackjax

We have already implemented a preliminary version of this and found the blackjax frameworks and abstraction incredibly useful. This would likely have a number of immediate applications in physics. We hope this would benefit blackjax in increased application there, as well as affording uncluttered comparison to other SoTA SMC methods.

### Are you willing to open a PR?

Yes, we have a working implementation, albeit needing some generalising and checking, we would make necessary refactors if this is deemed useful to the core library.

yallup:
## Description

Adds **nested sampling** to BlackJAX. This **supersedes #911** (the original Nested Sampling PR): that branch predated the slice-sampling merge and now conflicts with `main`, so it has been **rebuilt cleanly on top of the new slice primitives** (#943 / #946) — nested-sampling moves are now expressed directly through the merged slice `proposal_generator` abstraction instead of a parallel implementation. The result is a much smaller, in-sync diff (NS-only, no slice divergence).

Nested sampling estimates the Bayesian **evidence** (marginal likelihood) and draws posterior samples in a single run, and stays robust on multimodal, curving, and degenerate posteriors. BlackJAX currently ships MCMC and SMC but no evidence-first sampler; this fills that gap (cosmology, gravitational-wave inference, particle physics).

Two top-level samplers are provided, both built on the slice `proposal_generator` seam so very little new sampling machinery is introduced:

- **`blackjax.nss`** — Nested Slice Sampling: hit-and-run slice moves along directions shaped by the live-point covariance.
- **`blackjax.nsswig`** — Nested Slice-within-Gibbs: an axis-aligned coordinate sweep, scaled by the per-axis live-point spread.

Both share one generic engine:

- `base` — delete the lowest-likelihood live points and replace them via an inner kernel under the likelihood constraint.
- `adaptive` — carries the evidence integrator and re-tunes inner-kernel parameters from the live points each step.
- `from_mcmc` — wrap **any** MCMC kernel as a constrained inner kernel (propose-then-reject); the slice path gates the constraint in the proposal.
- `integrator` — the evidence integral (`logX`, `logZ`, `logZ_live`).
- `utils` — post-processing (`log_weights`, `sample`, `ess`, `finalise`, `uniform_prior`).

Algorithm reference: Yallup, Kroupa & Handley (2026), arXiv:2601.23252; foundational nested sampling: Skilling (2006); plateau handling: Fowlie, Handley & Su (2021).

The diff against `main` is confined to `blackjax/ns/*`, `tests/ns/*`, and the sampler registrations in `blackjax/__init__.py`.

- **Replaces #911** — the original Nested Sampling PR. That branch predated the slice-sampling work and conflicts with current `main`; this is a clean rebuild **rebased onto the merged slice primitives** (#943, #946).
- Resolves #753 — *Nested Sampling implementation* (tracking issue).
- Synced with the new slice-sampling changes: #943 (slice kernel) and #946 (slice integration). 

## Checklist

**General**

**Code quality**

**New sampler / algorithm** *(skip if not applicable)*
- [x] There is an open issue discussing this algorithm (use the [sampler proposal template](https://github.com/blackjax-devs/blackjax/issues/new?template=sampler_proposal.yml))
- [x] Follows the three-layer pattern: `init` / `build_kernel` / `as_top_level_api`
- [x] Registered in `blackjax/__init__.py`


junpenglao:
excellent work. I added minor docstring fix. merging after CI pass.

PR #961, opened 2026-07-06
fix(eca): pass args through shard_map instead of closing over it

### LINKED ISSUE #932: eca.py: ensemble_execute_fn closes over sharded JAX arrays in shard_map — NotImplementedError on N>1 devices
## Summary

`ensemble_execute_fn` in `eca.py` passes `args` to `shard_map(F, ..., in_specs=(p, p))` via **closure** rather than as an explicit `shard_map` argument with a matching `in_specs`. On a single device this is benign (trivial single-shard mapping). On N>1 devices, `args` can be a sharded JAX array (see crash path below) and `shard_map` raises `NotImplementedError` because it encounters a closed-over array with `NamedSharding` that doesn't have explicit sharding declared in `in_specs`.

## Crash path (LAPS burn-in, 4 devices)

```python
# laps_burn_in.py:134
initial_state, equipartition = ensemble_execute_fn(
    sequential_init, key1, num_chains, mesh, ...
)
# equipartition has out_specs=pscalar → NamedSharding(mesh, PartitionSpec())
# It is a *sharded JAX array*.

flat_equi, _ = ravel_pytree(equipartition)
signs = -2.0 * (flat_equi < 1.0) + 1.0
# signs inherits NamedSharding from equipartition — still sharded.

initial_state, _ = ensemble_execute_fn(
    ensemble_init, key2, num_chains, mesh, x=initial_state, args=signs, ...
)
# ^^^^^^^^ signs is now a sharded JAX array passed as args.
```

Inside `ensemble_execute_fn` (eca.py:301–306):
```python
def F(x, keys):
    y, summary_statistics = _F((x, args), (None, keys, None))[0]
    return y, summary_statistics

parallel_execute = shard_map(F, mesh=mesh, in_specs=(p, p), out_specs=(p, pscalar))
#                                         ^^^^^^^^^^^^^^^^^^
#                                         args is NOT in in_specs!
```

On N=4 devices, `shard_map` encounters `signs` (a JAX array with `NamedSharding(mesh, PartitionSpec())`) captured in the closure of `F`. This triggers:
```
NotImplementedError: shard_map does not support closed-over JAX arrays with NamedSharding
```

The traceback would point to `ensemble_init` (which is called inside `F`) because that's where the actual compute using `args`/`signs` happens.

## Secondary gap: degenerate `test_laps` hides the multi-device bug

The only LAPS test uses `jax.devices()[:1]` (always a 1-device mesh):
```python
mesh=jax.sharding.Mesh(jax.devices()[:1], "chains"),
```
On a 1-device mesh, `shard_map` with `NamedSharding` is trivially a no-op — no partitioning occurs and the closed-over sharding issue is never exercised. A real multi-device test would catch both the `keys_sampling.T` bug (fixed in #931) and this closure issue.

## Proposed fix

Pass `args` as an explicit `shard_map` argument with a matching `PartitionSpec`. Since `args` (= `signs`) is a global-broadcast array (not per-chain), it should use `pscalar`:

```python
def F(x, keys, args_local):
    y, summary_statistics = _F((x, args_local), (None, keys, None))[0]
    return y, summary_statistics

# Handle args=None by using a zero-size sentinel
args_for_shard = jnp.array(0.0) if args is None else args
args_spec = pscalar  # signs is broadcast (same for all devices)

parallel_execute = shard_map(
    F, mesh=mesh,
    in_specs=(p, p, args_spec),
    out_specs=(p, pscalar),
)
return parallel_execute(X, keys, args_for_shard)
```

**Complications to address:**
1. `args` type is `Any` — callers may pass `None`, a numpy array, a NamedTuple, or a JAX array. Need type narrowing or explicit `args_partitionspec` parameter.
2. The `None` sentinel requires special handling (cannot be a `shard_map` argument directly).
3. When `args` is per-chain data, `args_spec=p` would be appropriate instead of `pscalar`.

## Multi-device test approach

```python
# Add to test_laps or a new test_laps_multidevice:
import os
os.environ["XLA_FLAGS"] = "--xla_force_host_platform_device_count=4"
import jax  # reimport after setting XLA_FLAGS

mesh = jax.sharding.Mesh(jax.devices(), "chains")  # all 4 simulated devices
# ... rest of test identical to test_laps ...
```

This simulates 4 devices on CPU and would catch both the `keys_sampling.T` shape corruption (Issue #931, now fixed) and this `shard_map` closure issue.

## Related

- Issue fixed in PR #931: `keys_sampling.T` → `jnp.swapaxes(keys_sampling, 0, 1)` (old-style PRNGKey array shape corruption)
- Both issues are latent in the existing `test_laps` because it uses a 1-device mesh that exercises no real multi-device sharding.

Sumu004:
## Description

`ensemble_execute_fn`'s inner `shard_map`-wrapped function `F` passes `args` to `shard_map(F, ..., in_specs=(p, p))` via **closure** rather than as an explicit `shard_map` argument with a matching `in_specs`.

```python
def F(x, keys):
    """This function operates on a single device. key is a random key for this device."""
    y, summary_statistics = _F((x, args), (None, keys, None))[0]
    return y, summary_statistics

parallel_execute = shard_map(F, mesh=mesh, in_specs=(p, p), out_specs=(p, pscalar))
#                                          ^^^^^^^^^^^^^^^^^^
#                                          args is NOT in in_specs!
```

On a single device this is trivial (a single shard is a no-op), so it's never surfaced — which is exactly why every existing `test_laps*` test in `tests/mcmc/test_sampling.py` uses a 1-device mesh (`jax.devices()[:1]`) and never caught this.

On N>1 devices, if `args` is itself a sharded JAX array — which happens in `laps_burn_in.py`, where `signs` is derived from a prior `ensemble_execute_fn` call's `Etheta` output and threaded into the next call as `args` — `shard_map` raises:

```
NotImplementedError: shard_map does not support closed-over JAX arrays with NamedSharding
```

The traceback misleadingly points at the caller's kernel function (`ensemble_init` in the LAPS burn-in path, where `args`/`signs` is actually *used*), not at the closure itself.

## Fix

Pass `args` through `shard_map` explicitly, with an `in_spec` built via `jax.tree.map(lambda _: pscalar, args)`. This one line handles the default `args=None` case for free: `None` is registered as a zero-leaf pytree node in JAX (not a leaf itself), so `jax.tree.map` over `None` returns `None`, which is exactly the spec `shard_map` expects for a `None`-valued argument — no separate branch needed for the common no-`args` path.

```python
def F(x, keys, args):
    y, summary_statistics = _F((x, args), (None, keys, None))[0]
    return y, summary_statistics

args_specs = jax.tree.map(lambda _: pscalar, args)
parallel_execute = shard_map(
    F, mesh=mesh, in_specs=(p, p, args_specs), out_specs=(p, pscalar)
)
...
return parallel_execute(X, keys, args)
```

Fixes #932

## Testing

Added `tests/adaptation/test_eca.py`, forcing a genuine 2-device CPU mesh via `XLA_FLAGS` (set before JAX is first imported, since device count is fixed at first use — following the same pattern already used in `docs/examples/howto_sample_multiple_chains.md`). Two tests:

- `test_args_as_sharded_array_does_not_raise`: mirrors `laps_burn_in.py`'s exact call pattern — a first `ensemble_execute_fn` call produces a sharded `Etheta`, which is used to derive `signs`, which is then passed as `args` to a second call. Fails with `NotImplementedError` on `main` at 2 devices; passes with this fix.
- `test_args_none_still_works`: the default `args=None` path is unaffected.

I don't have real multi-device hardware to hand, but the `XLA_FLAGS` CPU-device-count trick reliably reproduces `shard_map`'s multi-device partitioning behavior (that's exactly why it's the standard technique JAX/Blackjax's own docs use for this).

## Checklist

**General**
- [ ] `pre-commit run --all-files` passes (black, isort, flake8, mypy) — ran `black` locally (clean); could not run `isort`/`flake8`/`mypy` in my environment, would appreciate CI confirming
- [x] Tests cover the changes (new `tests/adaptation/test_eca.py`, exercising the exact multi-device closure scenario from the bug report)

**Code quality**
- [x] Naming follows existing conventions (`jax.tree.map`, existing `p`/`pscalar` `PartitionSpec` naming)
- [x] All new code is JIT-compatible (no Python-level branching on traced values; the `args=None` handling is resolved at trace time via `jax.tree.map`, not inside the traced function)

### DIFF
--- a/blackjax/eca.py
+++ b/blackjax/eca.py
@@ -298,12 +298,25 @@ def ensemble_execute_fn(
         num_chains,
     )
 
-    def F(x, keys):
+    def F(x, keys, args):
         """This function operates on a single device. key is a random key for this device."""
         y, summary_statistics = _F((x, args), (None, keys, None))[0]
         return y, summary_statistics
 
-    parallel_execute = shard_map(F, mesh=mesh, in_specs=(p, p), out_specs=(p, pscalar))
+    # `args` may itself be a sharded JAX array (e.g. when chained from the
+    # output of a prior `ensemble_execute_fn` call, as in LAPS burn-in) rather
+    # than a plain Python value. Closing over it inside `F` instead of passing
+    # it through `shard_map`'s tracked arguments makes `shard_map` raise
+    # NotImplementedError on N>1 devices ("shard_map does not support
+    # closed-over JAX arrays with NamedSharding"); on a single device this is
+    # silently a no-op, which is why it stayed hidden. Passing `args`
+    # explicitly with a spec matching its own pytree structure (built via
+    # `jax.tree.map` so it also handles the `args=None` default, whose empty
+    # pytree structure needs no spec) fixes this for any number of devices.
+    args_specs = jax.tree.map(lambda _: pscalar, args)
+    parallel_execute = shard_map(
+        F, mesh=mesh, in_specs=(p, p, args_specs), out_specs=(p, pscalar)
+    )
 
     if superchain_size == 1:
         _keys = split(rng_key, num_chains)
@@ -318,4 +331,4 @@ def ensemble_execute_fn(
     )  # random keys, distributed across devices
 
     # apply F in parallel
-    return parallel_execute(X, keys)
+    return parallel_execute(X, keys, args)
--- a/tests/adaptation/test_eca.py
+++ b/tests/adaptation/test_eca.py
@@ -1,0 +1,110 @@
+# Copyright 2020- The Blackjax Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for blackjax.eca.ensemble_execute_fn on a multi-device mesh.
+
+This module sets XLA_FLAGS to force multiple host (CPU) devices *before*
+importing jax, since the device count is fixed at JAX's first use. All
+existing ``test_laps*`` tests in ``tests/mcmc/test_sampling.py`` run on a
+single-device mesh (``jax.devices()[:1]``), which is exactly why the
+shard_map closure bug fixed here (blackjax-devs/blackjax#932) went
+undetected: on one device, shard_map's per-device partitioning is a trivial
+no-op, so a closed-over sharded array never actually gets checked against
+shard_map's "no closed-over NamedSharding" rule.
+"""
+
+import os
+
+os.environ["XLA_FLAGS"] = "--xla_force_host_platform_device_count=2"
+
+import jax  # noqa: E402
+import jax.numpy as jnp  # noqa: E402
+import numpy as np  # noqa: E402
+from absl.testing import absltest  # noqa: E402
+from jax.sharding import Mesh  # noqa: E402
+
+from blackjax.eca import ensemble_execute_fn  # noqa: E402
+
+
+class ECAMultiDeviceTest(absltest.TestCase):
+    def setUp(self):
+        super().setUp()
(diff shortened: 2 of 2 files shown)


junpenglao:
Nice, thank you!!

junpenglao (approved):
Note for maintainer: trim the comment and follow up with a dedicated multi-device test suite

PR #976, opened 2026-07-11
Replace flaky single-chain funnel-neck assertions with a multi-chain robust grand-mean gate

### LINKED ISSUE #970: Flaky: test_neal_funnel_neck_marginal (gist_trajectory_length) — 3rd MC-flake occurrence
`tests/mcmc/test_gist_trajectory_length.py::MomentRecoveryTest::test_neal_funnel_neck_marginal` failed on main run 28999014867 and on PR #968's full matrix (all three Python versions — the seed-level signature: one shared bad realization, not a version effect), after three green main runs on the same code.

History: #957 (MC-flake fix) → #959 (self-calibrating ESS-gated tolerance). This is the third tolerance escape. Test is now SKIPPED (see linked PR) pending a sturdier approach — candidates: multi-seed median instead of single realization, or the `assert_decreasing_in_expectation`-style fixture from the noisy-MC-antipattern ledger, or widening the self-calibrating gate's floor.

Unskip criterion: a fix validated by ≥20 consecutive seeds locally.

— 🤖 Blackjax-devs AI TL

junpenglao:
## Problem

Neal's funnel \`test_neal_funnel_neck_marginal\` has escaped its MC tolerance three times now (issues #957, #959, and #970 — the third escape triggered \`@absltest.skip\` in #971). The root cause is structural, not a matter of tuning the existing gate further.

The prior assertion (\`assert_mean_within_ess_gated_tolerance\`) used a single-chain AR-ESS as the calibration signal. On the funnel, AR-ESS **over-estimates** actual ESS by ~7× (trapped chains produce high autocorrelation but also the misleadingly large apparent ESS from brief excursions). The \`ess_min\` floor was therefore a per-realization lottery: 6–17% of *legitimate*, unbiased chains happened to fall below it, causing spurious failures that were indistinguishable from genuine sampler failures. Additionally, \`std(samples) / sqrt(ESS)\` underestimates the true Monte Carlo standard error 2–4× on poorly-mixing targets, so the effective tolerance was tighter than intended.

## Fix

**New helper \`assert_grand_mean_within_robust_tolerance\` in \`tests/fixtures.py\`:**

Run K=24 independent chains (via \`jax.vmap\`) and assert on the *grand mean* of their per-chain means, using a MAD-based robust scale:

- **Location = plain grand mean**: trapped chains' rare deep excursions are physically real target visits that balance the finite-N bias — median/robust location is deliberately *not* used, because it centres on the biased bulk (+0.2 measured skew-bias on the funnel).
- **Scale = 1.4826 × MAD / √K**: the Gaussian consistency constant converts MAD to an asymptotically unbiased std estimate; MAD has breakdown point ~50%, so a single stuck chain cannot inflate the threshold and hide a real bias (std/MCSE loses up to half its detection power to that same trap-inflation mechanism).
- **Floor = \`atol_floor=1.0\`**: same-start chains share a common finite-N bias that cross-chain dispersion is structurally blind to; the floor prevents a collapsible robust scale from triggering false failures. The asserted claim is "no gross systematic bias", not exact recovery.
- **K=24**: large enough to shrink one trapped chain's leverage on the grand mean below the floor (power) and concentrate the null grand mean well inside it (null-escape rate ≲ 1e-3 by bootstrap).

The old \`assert_mean_within_ess_gated_tolerance\` is removed entirely — it had exactly two callers, both in the GIST test files.

**Per-variant chain lengths (halved from the prior single-chain budgets):**
- \`test_gist_trajectory_length\`: N=3000, burn=1500 per chain (×24 = effectively 36k post-burn samples)
- \`test_gist_step_size\`: N=6000, burn=3000 per chain (×24 = 72k post-burn samples)
- A pooled std check across all K chains (rtol=0.4) is kept for the step_size variant as a realization-robust sanity assertion.

The \`@absltest.skip\` decorator is removed from \`test_gist_trajectory_length.py::test_neal_funnel_neck_marginal\`.

## Validation

- The two rewritten funnel tests run in ~4.3s combined including JIT compilation (~1.1s warm) — faster than the single-chain versions they replace.
- Power controls: planted mean-biases ≥1.5 and a shifted-target control kernel at μ≥1.5 are each detected at ~100% — the robust gate still fails on genuine sampler bias.
- 490 tests in \`tests/mcmc\` pass (\`-n 2 --benchmark-disable\`), 0 failures
- \`uv run pre-commit run --all-files\` exits 0 — no new warnings
- Zero remaining references to \`assert_mean_within_ess_gated_tolerance\`; zero \`@absltest.skip\` in the two GIST test files

Closes #970

---

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -20,8 +20,6 @@ import jax.numpy as jnp
 import numpy as np
 from jax.flatten_util import ravel_pytree
 
-import blackjax.diagnostics
-
 
 class BlackJAXTest(chex.TestCase):
     """Base test case with date-based PRNG key management.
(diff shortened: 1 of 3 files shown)


junpenglao:
Extended validation of the shipped gate configuration (K=24, atol_floor=1.0, k_sigma=5, N=3000/6000), run post-review as belt-and-braces:

**220 fresh independent seeds per variant at the exact shipped configuration — zero failures.**
- trajectory_length: 220/220 pass, max |grand_mean| = 0.482 (threshold 1.0)
- step_size: 220/220 pass, max |grand_mean| = 0.670
- Direct 95% upper bound on the null escape rate: ≤1.36%, corroborating the bootstrap estimates (≈0 / ≈9e-4 per run).

**Binding-term decomposition** (which side of `max(atol_floor, k·robust_se)` governs each block): trajectory_length is floor-bound on all 220 blocks, while step_size splits 156 floor-bound / 64 MAD-bound — on 29% of step_size blocks the self-calibrating term rises above the floor and does the work. Both terms of the hybrid are load-bearing; neither a pure absolute tolerance nor a pure self-calibrating gate would suffice for both variants.

**Power-form agreement:** planted additive biases and an end-to-end shifted-target control kernel agree at μ=1.5 (trajectory 1.00, step ≈0.90 detection); the kernel form runs slightly lower at μ=1.0, as expected — it is the stricter ground truth (a few chains still trap on the biased target rather than receiving a clean constant shift).

— 🤖 Blackjax-devs AI TL

PR #982, opened 2026-07-12
Migrate the MEADS low-rank accumulator onto the moment-buffer layer

junpenglao:
## Change

MEADS's bespoke covariance accumulator (`_LowRankAccumulatorState` + inline Chan/Welford body) is replaced by the shared moment-block machinery introduced in #981. The result is a net deletion: one CGL implementation instead of two, with the MEADS consumer wired onto `MomentBlock` / `cgl_update_batch` from `blackjax.adaptation.metric_buffers`.

**Files changed:**
- `blackjax/adaptation/meads_adaptation.py` — remove the bespoke accumulator and two shim wrappers; inline `MomentBlock(count=, mean=, m2=)` and `cgl_update_batch` at the two call sites.
- `blackjax/adaptation/chees_adaptation.py` — update two stale docstring cross-references (docs-only; ChEES's own accumulator is deliberately left intact as a follow-up).

## Behavior-identical

The swap is bit-exact (0.0 difference) before and after through the full warmup path:

- Verified at float32 and float64, including prime chain counts, first-update `MomentBlock` promotion, and exact `2·d` support-gate boundary crossings — cases where float-precision or early-exit conditions could mask a behavioral change.
- Frozen-reference parity test (`EnsembleMeadsParityX64Test`) folds 20 batches through (A) a frozen inline copy of the pre-refactor Chan/Welford math and (B) `ensemble_batch_buffer`, then asserts bit-identical `(count, mean, m2)` under x64. The test is mutation-verified load-bearing: deliberately wrong accumulator math makes it fail, confirmed by two independent mutation controls.
- All existing MEADS tests pass unmodified (`tests/adaptation/test_meads.py`, 26 tests).

This PR also repairs a parity test that had become self-referential after the migration (`EnsembleMeadsParityX64Test` was comparing the shim — which delegated to `cgl_update_batch` — against `ensemble_batch_buffer`, also `cgl_update_batch`; now it compares the frozen inline reference against the production path), and adds an x64 end-to-end sanity test (`TestMEADSLRDX64Sanity`) exercising the full `run()` path at float64 precision.

The remaining duplicate accumulator in `chees_adaptation.py` (`_ChEESCovAccumulatorState`) is deliberately untouched; migrating it is a follow-up.

### DIFF
--- a/tests/adaptation/test_metric_buffers.py
+++ b/tests/adaptation/test_metric_buffers.py
@@ -924,8 +924,8 @@ class EnsembleMeadsParityX64Test(BlackJAXTest):
                 state = update(state, batch)
             block = get_moments(state)
 
-            # Pre-refactor frozen accumulator (reference implementation, not
-            # imported — embedded below in section 16 to prevent silent aliasing)
+            # Pre-refactor frozen accumulator (reference golden, defined in
+            # section 16 below — not imported, to prevent silent aliasing)
             acc = _frozen_lrd_accumulator_init(d)
             for batch in batches:
                 acc = _frozen_lrd_accumulator_update(acc, batch)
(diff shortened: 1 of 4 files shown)


tests/adaptation/test_metric_buffers.py
```diff
@@ -1961,5 +1960,99 @@ def test_correct_batch_shape_passes(self):
...
+            np.array(metric_b.U),
+            err_msg="MEADS parity: U differs (old accumulator vs ensemble_batch_buffer)",
+        )
+        np.testing.assert_array_equal(
+            np.array(metric_a.lam),
+            np.array(metric_b.lam),
+            err_msg="MEADS parity: lam differs (old accumulator vs ensemble_batch_buffer)",
+        )
+
+
```
junpenglao: I think we can remove this test.

junpenglao: Agreed — removed in a3e0e2112. After the earlier repair of `EnsembleMeadsParityX64Test` (which now compares the frozen inline reference against `ensemble_batch_buffer` directly), this end-to-end variant became strictly redundant: bit-equal moments imply bit-equal `sample_covariance_eigh_low_rank` outputs, so the extra eigh stage adds no discriminating power. The frozen reference helpers in that section are retained — the repaired parity test depends on them.

— 🤖 Blackjax-devs AI TL

PR #985, opened 2026-07-13
Add a staged warmup engine with a metric-recipe registry

junpenglao:
## Change

New public entry point `blackjax.staged_adaptation(algorithm, logdensity_fn, metric=..., schedule="stan")`: a staged warmup engine that composes a mass-matrix adaptation recipe with dual-averaging step-size adaptation over the Stan window schedule.

- `metric` accepts a registry name — `"welford_diag"` (the default, matching `window_adaptation`'s behavior), `"welford_dense"`, `"fisher_diag"` — or a directly constructed metric core, which is injectable for testing and extension. The registry sits at the top of `metric_recipes.py` so the available options are immediately visible.
- Recipes carry declarations (what they need, provide, and emit) validated at construction time with plain-Python errors: incompatible combinations fail before any tracing.
- The metric core is an embeddable init/update/final component, designed so other warmups can host it on their own update clocks (groundwork for ensemble-warmup integration).
- `window_adaptation` keeps the exact public surface and behavior of the **last release**, now implemented as a thin shim over the engine. The unreleased `diagonal_estimator` kwarg from #984 is reverted per review — the Fisher estimator is reachable only via `staged_adaptation(metric="fisher_diag")` — and `base()` returns to its released form. `window_adaptation_low_rank` is untouched (planned follow-up).
- Per review, behavioral test coverage begins migrating to `staged_adaptation` (the fisher end-to-end tests now target the engine); the shim carries code-path routing tests instead.
- The `"fisher_diag"` registry entry documents its validated scope: helps concentrated-anisotropy hierarchical geometry; degrades well-conditioned targets with correlated coordinate blocks; its degradation carries no R-hat/divergence signature, so estimator choices should be gated on ESS per gradient.

## Behavior-identical

The shim is verified bit-identical through full `run()`: 57 raw-byte comparisons (step size, inverse mass matrix, final position) covering the shim's diagonal and dense paths and the engine's fisher recipe, float32 and float64, multiple window layouts, non-default acceptance targets, and IMM shrinkage/seeding — zero mismatches, with a live sensitivity control confirming the comparison detects changes. Existing adaptation test suites pass unmodified. The new tests carry executed mutation coverage: estimator-formula inversion, recipe mis-mapping, reset skipping, vacuous contract validation, schedule re-export breakage, and fast-window isolation each fail at least one test.

**Resolved (maintainer decision):** `window_adaptation.base()` is no longer called by `window_adaptation()` itself (it is byte-identical to its released form) and will be deprecated in a follow-up PR.

### DIFF
--- a/blackjax/adaptation/metric_recipes.py
+++ b/blackjax/adaptation/metric_recipes.py
@@ -13,6 +13,30 @@
 # limitations under the License.
 """Metric recipes and the embeddable MetricCore protocol for staged_adaptation.
 
+Available recipes
+-----------------
+Pass any of the following string names as the ``metric=`` argument to
+:func:`~blackjax.adaptation.staged_adaptation.staged_adaptation`:
+
+- ``"welford_diag"`` — Stan-default diagonal Welford estimator; reproduces
+  :func:`~blackjax.adaptation.window_adaptation.window_adaptation` exactly.
+- ``"welford_dense"`` — Dense Welford covariance, same Stan schedule.
+- ``"fisher_diag"`` — Fisher-divergence-minimising diagonal estimator
+  (situational; requires position *and* gradient samples; see registry
+  provenance note for operational guidance).
+
+Usage::
+
+    # String sugar (registry lookup):
+    wu = staged_adaptation(nuts, logdensity_fn, metric="welford_diag")
+
+    # Recipe constructor (testability / custom configuration):
+    from blackjax.adaptation.metric_recipes import REGISTRY
+    core = REGISTRY["welford_diag"].build_core(imm_shrinkage_to_previous=5.0)
+    wu = staged_adaptation(nuts, logdensity_fn, metric=core)
+
+Design
+------
 A :class:`MetricRecipe` declares an (estimator, buffer, representation,
 support_gate) tuple with construction-time validation of the coupling
 contract (``needs ⊆ provides`` and ``emits == representation``): incompatible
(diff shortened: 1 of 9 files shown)


blackjax/adaptation/window_adaptation.py
```diff
@@ -73,7 +91,7 @@ def base(
     Schematically:
 
     +---------+---+------+------------+------------------------+------+
-    |  fast   | s | slow |   slow     |        slow            | fast |
+    |  fast   | s | slow |   slow     |        slow            |  fast |
```
junpenglao: revert

blackjax/adaptation/metric_recipes.py
```diff
@@ -0,0 +1,474 @@
...
+        )
+
+    return MetricCore(init=init, update=update, final=final)
+
+
+# ---------------------------------------------------------------------------
+# Registry
+# ---------------------------------------------------------------------------
+
+REGISTRY: dict[str, MetricRecipe] = {
```
junpenglao, line 378: Move this up so user immediately sees what are the options currently provide

blackjax/adaptation/window_adaptation.py
```diff
@@ -361,6 +388,17 @@ def final(warmup_state: WindowAdaptationState) -> tuple[float, Array]:
     return init, update, final
 
 
+def _pick_recipe_name(*, is_mass_matrix_diagonal: bool, diagonal_estimator: str) -> str:
+    """Map the old (is_mass_matrix_diagonal, diagonal_estimator) params to a
+    metric recipe registry name.  Used by the :func:`window_adaptation` shim."""
+    if diagonal_estimator == "fisher":
```
junpenglao: diagonal_estimator is a new behavior introduced in last PR, which i think we should already revert it so window_adaptation keeps the same API as last release

blackjax/adaptation/window_adaptation.py
junpenglao: Restored to the exact released diagram in 8dc78fdd3.

— 🤖 Blackjax-devs AI TL

blackjax/adaptation/metric_recipes.py
junpenglao: Done in 8dc78fdd3 — the module now leads with an "Available recipes" listing in the docstring and the registry/lookup sit at the top of the module, so the options are the first thing a reader sees.

— 🤖 Blackjax-devs AI TL

blackjax/adaptation/window_adaptation.py
junpenglao: Agreed and done in 8dc78fdd3: `diagonal_estimator` is removed from `window_adaptation` and `base()` is reverted to its released form (verified byte-identical to the pre-#984 file for the whole `base` body) — `window_adaptation` now matches the last release exactly, and the Fisher estimator is reachable only via `staged_adaptation(metric="fisher_diag")`. A pleasant side effect: this deleted the temporarily-duplicated Fisher finalization logic, leaving a single copy in the recipe module. The fisher end-to-end tests were migrated to target `staged_adaptation`, and the shim now carries code-path routing tests per your direction; the broader migration of the welford behavioral tests to `staged_adaptation` will follow in a dedicated test-refactor PR.

— 🤖 Blackjax-devs AI TL

PR #1005, opened 2026-07-17
[Enhancement] Factor the nested slice sampling covariance once per outer step

### LINKED ISSUE #1004: [Enhancement] Factor the nested slice sampling covariance once per outer step
### Checklist

- [x] I have searched existing issues and this is not a duplicate.

### Area

Other - Nested Slice Sampling

### Current behaviour

Nested slice sampling generates each covariance-shaped direction by sampling

```math
y \sim \mathcal N(0,C)
```

and normalising it to Mahalanobis norm 2:

```math
d=\frac{2y}{\sqrt{y^\mathsf{T}C^{-1}y}}.
```

With JAX’s default Cholesky-based multivariate-normal sampler, this requests both a Cholesky factorisation and an explicit inverse of the same live-point covariance from inside the inner slice loop.

The covariance is fixed throughout an outer nested-sampling step, so these operations are loop-invariant. XLA can hoist some of this work on certain backends, but the source does not guarantee that the factorisation is shared, and the explicit inverse is unnecessary.

### Desired behaviour

Compute the live-point covariance and its Cholesky factor once per outer nested-sampling step, then reuse that factor for all replacement chains and inner slice steps.

If

```math
C=LL^\mathsf{T}, \qquad z\sim\mathcal N(0,I),
```

then $y=Lz$ and

```math
y^\mathsf{T}C^{-1}y=z^\mathsf{T}z.
```

The direction can therefore be generated directly as

```math
d=\frac{2Lz}{\lVert z\rVert_2}.
```

### Desired behaviour

The implementation should:

- Store the covariance factor as an unmapped inner-kernel parameter.
- Reuse it across all \(K M\) directions in an outer step.
- Remove the explicit matrix inverse from the default NSS path.
- Preserve the direction distribution and Mahalanobis norm.
- Leave the likelihood constraint, stepping-out, and shrinkage procedures unchanged.
- Preserve the public `live_covariance` and covariance-based custom-proposal extension seams.
- Remain compatible with `jax.jit` and PyTree positions.

No covariance jitter should be introduced: positive definiteness was already required by the original Cholesky-backed multivariate-normal path.

### Motivation

For $N$ live points, deletion batch size $K$, $M$ inner steps, and dimension $d$, the factored formulation performs approximately

```math
O(Nd^2)+O(d^3)+O(KMd^2)
```

work per outer step: compute the covariance, factor it once, then apply the factor to each direction.

This makes the intended placement explicit instead of relying on compiler loop-invariant-code motion, and it avoids the numerical and computational cost of evaluating a quadratic form through $C^{-1}$.

On an Apple M3 CPU with JAX 0.10.0 and `float32`, timing one compiled nested-sampler step with \(N=4d\), \(M=2d\), and \(K=1\) produced:

| Dimension | Existing path | Factored once | Speedup |
|---:|---:|---:|---:|
| 1024 | 199.4 ms | 126.4 ms | 1.58× |
| 1536 | 0.953 s | 0.351 s | 2.71× |
| 2048 | 3.160 s | 1.174 s | 2.69× |
| 3072 | 23.52 s | 9.91 s | 2.37× |
| 4096 | 58.76 s | 22.59 s | 2.60× |

Fixed-seed paths matched through 2048 dimensions within \(4.7\times10^{-6}\). At larger dimensions, changes in floating-point evaluation order altered some slice control-flow decisions; exact seeded trajectories are therefore not expected to remain bitwise identical.

### Acceptance criteria

- The default NSS kernel factors the live covariance once per outer step.
- Inner directions use \(2Lz/\lVert z\rVert_2\).
- The default path contains no explicit covariance inverse.
- Existing covariance-based custom proposal callbacks continue to work.
- Existing nested-sampling tests pass unchanged.
- Sampler semantics and constrained target remain unchanged in exact arithmetic.

### JIT compatibility

- [x] I have considered whether this change is compatible with `jax.jit` and believe it is.

### Are you willing to open a PR?

Yes — I can implement this

Idg772:
## Description

Factor the live-point covariance once per outer nested-sampling step and reuse the resulting Cholesky factor across all inner slice steps.

Previously, each direction was constructed as

```math
y \sim \mathcal N(0,C), \qquad
d = \frac{2y}{\sqrt{y^\mathsf{T}C^{-1}y}}.
```

Writing \(C=LL^\mathsf{T}\) and \(y=Lz\), where \(z\sim\mathcal N(0,I)\), gives

```math
y^\mathsf{T}C^{-1}y=z^\mathsf{T}z,
```

so the same direction distribution can be generated directly with

```math
d=\frac{2Lz}{\lVert z\rVert_2}.
```

This PR therefore:

- computes the Cholesky factor once when the live-point geometry is updated;
- shares it across the replacement chains and inner slice steps;
- removes the explicit covariance inverse from the default NSS path;
- adds `live_covariance_factor` and a factor-based direction sampler;
- retains `live_covariance` and covariance-based calls to `covariance_proposal`;
- keeps custom proposal defaults on the existing covariance-based extension seam.

The likelihood constraint, stepping-out procedure, shrinkage procedure, and constrained target are unchanged. Fixed-seed trajectories need not be bitwise identical because the floating-point evaluation order changes.

No covariance jitter is added: the positive-definiteness requirements are the same as for the previous Cholesky-backed multivariate-normal draw.

### Performance

For \(N\) live points, deletion batch size \(K\), \(M\) inner steps, and dimension \(d\), the new formulation performs approximately

```math
O(Nd^2) + O(d^3) + O(KMd^2)
```

work per outer step.

A benchmark on an Apple M3 CPU using JAX 0.10.0 and `float32`, with \(N=4d\), \(M=2d\), and \(K=1\), measured one compiled sampler step:

| Dimension | Previous | Factored once | Speedup |
|---:|---:|---:|---:|
| 1024 | 199.4 ms | 126.4 ms | 1.58× |
| 1536 | 0.953 s | 0.351 s | 2.71× |
| 2048 | 3.160 s | 1.174 s | 2.69× |
| 3072 | 23.52 s | 9.91 s | 2.37× |
| 4096 | 58.76 s | 22.59 s | 2.60× |

Fixed-seed paths matched through 2048 dimensions within \(4.7\times10^{-6}\). At larger dimensions, floating-point differences changed some slice control-flow decisions.

Closes #1004

## Validation

```console
JAX_PLATFORM_NAME=cpu uv run pytest tests/ns/test_nested_sampling.py -q
26 passed in 29.85s
```

## Checklist

**General**

- [x] The branch is based on the latest `main`
- [ ] `uv run pre-commit run --all-files` passes
- [x] Relevant nested-sampling tests pass

**Code quality**

- [x] Public functions have NumPy-style docstrings
- [x] Naming follows existing conventions

**New sampler / algorithm**

Not applicable; this optimises the existing nested slice sampler.

### DIFF
--- a/blackjax/ns/nss.py
+++ b/blackjax/ns/nss.py
@@ -48,30 +48,83 @@ __all__ = [
     "covariance_proposal",
     "init",
     "live_covariance",
+    "live_covariance_factor",
     "live_widths",
     "slice_constrained_step",
     "swig_as_top_level_api",
 ]
 
 
+def sample_direction_from_covariance_factor(
+    rng_key: PRNGKey, position: ArrayTree, covariance_factor: Array
+) -> ArrayTree:
+    """Sample a covariance-shaped direction with Mahalanobis norm 2.
+
+    Given a lower-triangular factor ``L`` satisfying ``L @ L.T = C``, draws
+    ``z ~ N(0, I)`` and returns ``2 * L @ z / ||z||``. The factor can therefore
+    be computed once per outer nested-sampling step and reused by every inner
+    slice step.
+
+    Parameters
+    ----------
+    rng_key
+        Key used to draw the standard-normal direction.
+    position
+        Position PyTree whose structure the returned direction must match.
+    covariance_factor
+        Lower-triangular Cholesky factor of the live-point covariance.
+
+    Returns
+    -------
+    A direction PyTree with length 2 in the live covariance metric.
+    """
+    _, unravel_fn = jax.flatten_util.ravel_pytree(position)
+    dimension = covariance_factor.shape[-1]
+    standard_direction = jax.random.normal(
+        rng_key,
+        shape=(dimension,),
+        dtype=covariance_factor.dtype,
+    )
+    direction = (
+        2.0
+        * (covariance_factor @ standard_direction)
+        / jnp.linalg.norm(standard_direction)
+    )
+    return unravel_fn(direction)
+
+
 def sample_direction_from_covariance(
     rng_key: PRNGKey, position: ArrayTree, cov: Array
 ) -> ArrayTree:
-    """A random direction shaped by ``cov``, scaled to Mahalanobis norm 2.
+    """Sample a direction from a covariance matrix.
 
-    Samples ``d ~ N(0, cov)`` and normalizes it with ``inv(cov)`` to a length of
-    2 in the covariance metric (~2 std devs, a step size that mixes well). Uses
-    ``inv(cov)`` rather than an explicit Cholesky factor, has encountered differing
-    behavior on GPU.
+    This compatibility helper factors ``cov`` for each call. The default NSS
+    kernel uses :func:`sample_direction_from_covariance_factor` so that the
+    factorization is performed outside the inner sampling loop.
+
+    Parameters
+    ----------
+    rng_key
+        Key used to draw the standard-normal direction.
+    position
+        Position PyTree whose structure the returned direction must match.
+    cov
+        Live-point covariance matrix.
+
+    Returns
+    -------
+    A direction PyTree with length 2 in the live covariance metric.
     """
-    _, unravel_fn = jax.flatten_util.ravel_pytree(position)
-    d = jax.random.multivariate_normal(rng_key, jnp.zeros(cov.shape[0]), cov)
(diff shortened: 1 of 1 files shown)


junpenglao:
nice, thanks!

junpenglao (approved):

PR #1025, opened 2026-08-27
Expose update_strategy on the NS kernel

AdamOrmondroyd:
## Description

`ns/from_mcmc.build_kernel` hardwires `update_with_mcmc_take_last` as the way replacement particles are seeded. This makes it a keyword argument with that function as the default, and threads it through the four `nss` entry points. Nothing changes unless you pass something.

This isn't new; it existed in the handley-lab NS fork, and was lost when nested sampling came upstream in #947. Nothing is impossible without it, and I'm managing to extend stock blackjax≥1.6 today with a multimodal update. However, in its current form, I have to rebuild `blackjax.nss` out of parts, using eight imports which are all internal, and re-declare `num_delete`, `max_steps` and `max_shrinkage` now that `**kwargs` is gone.

The strategy we want draws each replacement from a cluster in proportion to its prior volume, rather than uniformly from the survivors, similar to how PolyChordLite works. This is what keeps separated modes alive when one of them is losing points. It is a decision about seeding rather than about the inner kernel, which is why `proposal` is not the right place for it. None of the clustering code is in this PR; this is just an example I think is important which is much easier with this change.

Cluster bookkeeping which has to persist between steps already has a route. `adaptive.build_kernel` applies the inner kernel as `partial(inner_kernel, **state.inner_kernel_params)`, so whatever the `inner_kernel_params` seam returns reaches the strategy as `step_parameters`.

There is precedent in `smc.from_mcmc.build_kernel`, which already takes an `update_strategy`, and is how `waste_free_smc` plugs in. The contracts are not identical, since SMC consults its strategy every step while NS only does so at construction.

I've already run these changes past @yallup (from #946 and #947) who was happy with my suggestions.

`test_update_strategy_seam` runs against both top-level APIs, and asserts the strategy is consulted exactly once with the caller's `(num_inner_steps, num_delete)`, and that a full step still runs. That first assertion is the important one, since a regression which dropped the keyword would fall back to the default and still pass everything else in the suite. I broke the seam deliberately to check the test notices. `tests/ns/` passes, 28 tests.

Just one question: a parameter threaded through `build_kernel`, `build_swig_kernel` and the two top-level APIs ends up documented four times, and the existing ones are not consistent about how. `coordinate_order` is terse everywhere, `proposal` cross-references its non-SwiG counterpart, and `inner_kernel_params` is written out in full each time. I have given `update_strategy` identical wording at every site, but is there a house rule?

Related to #947 (which superseded #911 and rebuilt on the slice primitives from #946).

## Checklist

**General**

**Code quality**

### DIFF
--- a/blackjax/ns/from_mcmc.py
+++ b/blackjax/ns/from_mcmc.py
@@ -169,6 +169,7 @@ def build_kernel(
     update_inner_kernel_params_fn: Callable,
     num_delete: int = 1,
     delete_fn: Callable = default_delete_fn,
+    update_strategy: Callable = update_with_mcmc_take_last,
 ) -> Callable:
     """Build a Nested Sampling kernel from a constrained inner step.
 
@@ -191,14 +192,15 @@ def build_kernel(
         Number of particles replaced per NS iteration.
     delete_fn
         Selects which particles to delete (default: the lowest-likelihood ones).
+    update_strategy
+        Inner-kernel factory ``(constrained_step_fn, num_inner_steps,
+        num_delete) -> update_fn`` (default: :func:`update_with_mcmc_take_last`)
 
     Returns
     -------
     A Nested Sampling kernel ``kernel(rng_key, state) -> (new_state, info)``.
     """
-    inner_kernel = update_with_mcmc_take_last(
-        constrained_step_fn, num_inner_steps, num_delete
-    )
+    inner_kernel = update_strategy(constrained_step_fn, num_inner_steps, num_delete)
     delete_fn = partial(delete_fn, num_delete=num_delete)
     return build_adaptive_kernel(
         delete_fn,
--- a/blackjax/ns/from_mcmc.py
+++ b/blackjax/ns/from_mcmc.py
@@ -194,7 +194,22 @@ def build_kernel(
         Selects which particles to delete (default: the lowest-likelihood ones).
     update_strategy
         Inner-kernel factory ``(constrained_step_fn, num_inner_steps,
-        num_delete) -> update_fn`` (default: :func:`update_with_mcmc_take_last`)
+        num_delete) -> update_fn`` (default:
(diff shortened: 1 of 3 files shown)


junpenglao:
Thanks — the implementation and default nested-sampling behavior look sound. I found two things to address:

1. **Please make the regression test prove that the strategy's returned updater is installed.** `recording_strategy` currently returns `update_with_mcmc_take_last`, and the assertions only check that the factory was called plus the final particle shape. A broken builder that calls `update_strategy(...)`, discards its return value, and then hardwires the default updater still passes both parametrized cases (I verified this mutation: 2/2 still pass). Please have the custom strategy return a behaviorally distinct updater — for example, one that emits a sentinel in `info.update_info` — and assert that sentinel after the jitted step. This is the one change I would require before merge.

2. **Please document the returned updater's contract in the canonical `from_mcmc.build_kernel` docstring.** The parameter currently documents only the factory signature. The returned callable is actually invoked as
   ```python
   update_fn(rng_key, state, loglikelihood_0, **step_parameters)
       -> (new_particles, info)
   ```
   It should also state that `new_particles` contains exactly `num_delete` fixed-structure/JAX-compatible particle states, and mention the contour/birth-likelihood obligations. For the duplication question in the PR description: I would put the full contract once in `from_mcmc.build_kernel` and keep the four forwarding APIs terse with a cross-reference. I consider this documentation item non-blocking.

I also verified that omitted/default NSS and SwiG calls are bit-identical to `main`, and a distinct custom updater works through all four public constructors under `jax.jit`, including receipt of the adaptive parameters.

— 🤖 Blackjax-devs AI TL

AdamOrmondroyd:
I think I've dealt with both of those suggestions: sentinel added and the mutation now caught, contract moved into from_mcmc.build_kernel with the forwarders cross-referencing it.

junpenglao:
Thank you!!!

junpenglao (approved):

PR #1039, opened 2026-09-17
Use two-sided nonnegative A in the isokinetic kinetic increment

### LINKED ISSUE #1038: Isokinetic momentum update loses up to 19 nats near antiparallel momentum; use the non-negative form
Follow-up from the adversarial review of #1036. That PR fixes the *downstream* consequence (#1035); this is the source.

## The defect

`blackjax/mcmc/integrators.py:459,470-473` computes the isokinetic momentum update's log argument as

```python
A = 1 + momentum_proj + (1 - momentum_proj) * zeta**2
```

`momentum_proj` is a dot product of two unit vectors, so `|momentum_proj| <= 1` mathematically — but in float32 it can round marginally below `-1`, making `A` negative and `jnp.log(A)` a NaN.

Instrumenting all **204,500** momentum updates of a full 1000-iteration, 100-chain LAPS burn-in: each failing run contains **exactly one** call with `A <= 0`, always at

```
momentum_proj = -1.0000001192092896   # exactly -(1 + 2**-23), one float32 ulp below -1
```

With `p = -(1+u)`, `u = 2**-23`: `A = -u + (2+u)*zeta**2`, so `A < 0` iff `zeta < sqrt(u/2) = 2.44e-4`. That condition separates failing from healthy runs exactly — a healthy key had two `p < -1` events whose `zeta` was too large to trigger it. `p < -1` is necessary but not sufficient.

## Do not "fix" this by clamping

The obvious repair, `momentum_proj = jnp.clip(momentum_proj, -1.0, 1.0)`, does remove the NaN — but it converts a loud failure into a quiet one. Scored against a float64 oracle over 200,000 float32 cases with the momentum within `1e-8 .. 1` rad of antiparallel to the gradient:

| form | `A < 0` | `log(A)` non-finite | median abs err in `log A` | p99 | max |
|---|---|---|---|---|---|
| current `1 + p + (1-p)z²` | 623 | 623 | 4.45e-05 | **14.3** | 19.4 |
| clamp `p <- clip(p,-1,1)` | 0 | 0 | 4.42e-05 | **14.3** | 19.4 |
| `A = ½(‖m+ĝ‖² + z²‖m−ĝ‖²)` | 0 | 0 | **2.53e-07** | **0.123** | 2.18 |

The clamp leaves up to **19 nats** of error in the energy accounting, which flows into `EEVPD` and the step-size adaptation with nothing to catch it. That is strictly worse than the status quo, where the NaN is at least visible.

## Proposed fix

Use the algebraically identical form, via `‖m ± ĝ‖² = 2 ± 2p`:

```python
A = 0.5 * (sq_norm(momentum + grad_unit) + zeta**2 * sq_norm(momentum - grad_unit))
```

Non-negative by construction (so no clamp is needed), cancellation-free as `p -> -1`, and it cuts the p99 error from 14.3 nats to 0.12. Cost: two length-`d` squared norms replacing one length-`d` dot product — same order as what is already there.

## Keep the #1036 guards regardless

There is a residual route this does not close: `zeta` underflows to exactly `0.0` on ~1500 of 204,500 calls per run. If `p == -1.0` exactly co-occurs with `zeta == 0`, then `A == 0` and `log(0) = -inf`, not NaN. That did not occur in the four instrumented runs but is reachable — and it is the concrete justification for #1036 choosing `isfinite` over `isnan` in the NaN gate.

## Verification

Any candidate should be scored against a float64 oracle on near-antiparallel cases as above, and checked bit-identical on runs where `p > -1` throughout (the change must be a no-op on the ~204,499 calls that are already well-conditioned).

— 🤖 Blackjax-devs AI TL

shaneraphel:
## Description

The isokinetic (ESH) kinetic increment logs

```
A = 1 + p + (1 - p) * zeta**2
```

with `p = m · ĝ`. In float32, `p` can sit one ulp below `-1`, so `A` goes negative and `log A` is NaN. Clipping `p` to `[-1, 1]` hides the NaN and leaves up to ~19 nats of energy error in the same near-antiparallel cases. Replacing the signed form by a single squared-norm rewrite `½(‖m+ĝ‖² + ζ²‖m-ĝ‖²)` is nonnegative, but it is wrong at a stationary point: `_normalized_flatten_array` returns the raw (zero) vector when the gradient norm is at most `1e-13`, leftover `A` is `2`, and the one-norm form drops to `1` (a `-(d-1) log 2` kinetic error).

This keeps the momentum map unchanged and dests the two-sided identity

```
A = S_plus * (1 - zeta**2) + 2 * zeta**2    if zeta**2 <= 1
A = S_minus * (zeta**2 - 1) + 2             otherwise
```

with `S± = ½ ‖m ± ĝ‖²`. Each branch is exact on the unit sphere, `A = 2` at a zero gradient (matching leftover), and `A` stays nonnegative for Omelyan `ζ > 1`.

Closes #1038

## Checklist

**General**
- [ ] `pre-commit run --all-files` passes (black, isort, flake8, mypy)
- [ ] Tests cover the changes (`mamba run -n blackjax python -m pytest tests/`)

**Code quality**
- [ ] Public functions have docstrings following the NumPy style guide

**New sampler / algorithm** *(skip if not applicable)*
- [ ] There is an open issue discussing this algorithm
- [ ] Follows the three-layer pattern: `init` / `build_kernel` / `as_top_level_api`
- [ ] Registered in `blackjax/__init__.py`


junpenglao (approved):
Thanks!

